PR #2, opened 2021-11-25
Move to scala

TomCranitch:
This PR mainly adds the SBT. If we merge the other branch first this can then be merged into main


TomCranitch:
Some of the notable changes are:
 - The codegen for antlr will now place the generated class into the target director 
 - antlr is now automatically run on compilation (but only if the antlr file changes)

PR #13, opened 2021-12-20
Tom

TomCranitch:


andrewj-brown:
I'm getting a lot of "Not Found Error"s trying to run this after merging with andrew-worklist; most seem to be centered on the astnodes folder(s) - MemLoad in particular. src/main/java and src/main/java are both on the source path.

Is it possible that the packages have gotten messed up through all these changes or is it more likely a problem with my environment?

PR #17, opened 2022-01-25
Scalafmt

TomCranitch:
It would be good to get consistent formatting but this might mess up merging for existing commits  

To run the formatter run `scalafmtAll` in an sbt shell


andrewj-brown:
yeah can the formatting only get applied to a subset of the code? I'm happy to run it separately on my branch but I don't want to have to manually merge every worklist change I've made

TomCranitch:
that is a good question, it probably can (i might look into that). otherwise maybe just worry about this at some point later when its convenient. its definitely not worth the effort to have to manually merge the changes

l-kent:
This is something we'll get to but obviously this pull request is redundant for the time being

PR #26, opened 2022-05-25
WIP: Replace BilParser with BilAdtParser

tomdx:
# To Do
- Parse the annotations (the grammar for this is specified by us so it should not be too difficult)
- Testing (Preferably with a wider range of adt files, or get the testing suite to work)
- General cleanup of code

# Motivation
The current Bil Parser is lacking in some information, e.g. bit vector sizes and explicit control flow graph information. By instead parsing the ADT, we have explicit access to this information, as well as a lot of other information that may be required.

The ADT can be generated by the following command
`bap $BINARY -d adt`

# Changes
- Deprecate the Bil Parser (Bil.g4, StatementLoader.scala)
- Replace with Bil ADT (Abstract data type) parser (BilAdt.g4, AdtListener.scala)
- Configure RunUtils.scala to construct the State using the new parser


l-kent:
If this is a WIP, what still needs to be done?

tomdx:
Hi, the thing still left to do are:- Parse the annotations (the grammar for this is specified by us so it should not be too difficult)- Testing it with more files to minimise bugs- I'd like to go over my changes and clean everything upI've edited the PR accordingly.Thanks,TomOn 30 May 2022 09:29, l-kent ***@***.***> wrote:
If this is a WIP, what still needs to be done?

l-kent:
It would be good to update the readme and perhaps add a shell script to output the ADT given an input C file. I'm still familiarising myself with the project so I can't comment more than that for now

andrewj-brown:
If @tomdx is the only "active" developer of the project at the moment, it might be worth running `scalafmt` (as mentioned here: https://github.com/UQ-PAC/bil-to-boogie-translator/pull/17) and integrating that into this PR, while it's not going to cause merging headaches...

l-kent:
another thing is if it's possible, it would be nice to keep the previous BIL front-end as an option just to make comparisons easy etc.

l-kent:
I've added a shell script for creating an adt file from a given c file, and fixed an error in the parser (though it doesn't seem to have made a difference, it was able to recover before)

tomdx:
Thanks, I'm caught up in assessments right now, so I'll address all the issues on Thursday/friday. In the meanwhile feel free to make changes as you see fit and let me know of any other issues.

l-kent:
Another thing that would be good to have is a short list of benefits you've noticed compared to the old approach, it'd be good to know how much of a difference it has made

tomdx:
Given that the old parser frontend is still accessible in this version, I think this is a safe point to merge. The new parser is invoked if and only if the filename ends with "adt".

Happy for any more feedback.

l-kent:
I had a look at applying Scalafmt to the whole project and it makes some sections quite unreadable so it would probably take some tweaking the settings to get a good outcome there. That's worth doing at some point but I'm going to have to leave it for now

PR #33, opened 2023-05-04
Memory Region Analysis

yousifpatti:
- CFG intra/interproc
- Memory Region Analysis

Fixed conflicts


l-kent:
I tried running the tests, and after fixing up some minor issues so it would compile, it crashes for all examples in the memory region tests at line 269 in Cfg.scala. 
The issue seems to be that if the Cfg generation encounters a block with no statements (only jumps), such as block 000002e6 in the arrays example, then newBlocks(g.target.label) returns an empty ArrayBuffer, so .head fails on that empty ArrayBuffer.

yousifpatti:
Hi Liam,

I have investigated and the issue appears to be coming from changes in the underlying IR.
In the old version this is the procedures collected for basic_array_write example:
oldIRProcedures

Currently, the number of procedures is as follows:
newIRProcedures

The procedures do not seem to be necessary, and they make the graph very complicated. Previously, for most examples, it would only have main, functions, malloc, free, etc. 

Do you require all those procedures for your purposes?

I have fixed the issue and Memory Region Analysis works fine. However, I have had to comment out Constant Propagation Analysis as it causes it to break.

ps. my tests would not pass because the string comparison would fail dude to the many procedures added to the results. The analysis still works and you can check the output graph in _test.dot_ and plot it in _Edotor.net_


Let me know how you wish to proceed, and I can adjust my analysis/CFG accordingly.

l-kent:
A few months ago things were moved around so that now the analysis takes place before all procedures not reachable from main are removed, which is why you're only encountering those now. This is an important change, as previously any procedures that are only accessed via indirect calls are completely removed before the analysis occurs, which means that we can't resolve the indirect calls in the analysis. Obviously we're not quite at the stage of making use of those yet though.

Eventually we'll want to do things in an order like this:
resolving indirect calls -> removing unreachable procedures -> memory region analysis

This might require a more nuanced approach than just running the analyses on every procedure. For example, we probably don't need to do constant propagation on the entire program, just where there are indirect calls that are reachable from main, so we might need to have the constant propagation analysis start with main and resolve its indirect calls before analysing procedures reachable from main. In most cases it's likely that the constant propagation could be further limited to just before the indirect call site if we wanted to optimise things.

For the time being, I can move removing the unreachable procedures back to happening before the analysis if that's what would be easiest for you? Either way, those extra procedures can all be ignored at the moment for your current purposes, as they're mostly initialisation/cleanup code that we're considering out of scope. 

If the constant propagation is breaking on some of those procedures you still should investigate why - it could be edge cases you haven't handled that were always possible (like the blocks with only calls and no statements) in which case they should be fixed.

Aside from all that, this seems fine to merge in now.

yousifpatti:
Yes. Would it be possible to do remove those for now?
It will make visualising the result of the analysis a bit more easier.

I will merge the branch

PR #44, opened 2023-07-19
CFG merging

Thomas-Malcolm:
This merges intra and inter procedural CFGs into one, allowing analysis to be performed on a single CFG instead of creating specific ones for each analysis.

These changes are best not merged until the analyses have been updated to accommodate the new structure. There are a number of issues associated with this PR:

- #45
- #46
- #47 

Of these, #45 and #46 are optimisations / enhancements which could be made, however #47 should be considered a requirement, as inter-procedural analyses will not work with this CFG structure until it is implemented.

Analyses were modified such that the tool would compile, and not tested past this - the exception of this being Constant Propagation, which seems to work for at least the intra-procedural case. Otherwise, as far as I've tested, the CFG creation is successful, fixes some bugs from the previous version (relating to indirect calls), and successfully permits inter and intra walks. A small change-list:

- `InterproceduralCfg` and `IntraproceduralCfg` have both been merged into a single `ProgramCfg`
- New edge types (`RegularEdge`, `InterprocEdge` and `IntraprocEdge`) have been introduced to facilitate distinguishing between intra- and inter- procedural CFG walks
- New node types to reflect behaviour after calls (e.g. whether a call is supposed to return or not). These are treated as NOPs semantically, and exist as a convenience for analyses.
- In-lining to a depth of `n` is now supported (this has a flat approach and does not discriminate depths between procedures, though this can be extended fairly easily to in-line certain functions to specific depths. There is also a question as to whether this is needed, but nonetheless is provided as an option)
- Edges for conditional branches are now reflected in the CFG, with those conditions accessible via the `succ` and `pred` APIs
- The `succ` and `pred` APIs were largely maintained to make integration with analyses easier, however some changes (and additions) were made. These APIs by default still return their respective nodes, though require a flag to be passed (`intra: Boolean`), specifying whether or not the caller wishes to walk the intra/inter CFG. Additional API options were introduced: `(succ|pred)Edges` - returns the associated edges, and `(succ|pred)Conds` - returns the node as for `succ|pred`, but with the condition on moving to that node (if it's an unconditional connection, then this is a `TrueLiteral`). This API can be simplified significantly by only returning edges, which contain all the information provided by in the other accessors. Analyses will have to be modified to accommodate this.
- CFG visualisation (`.dot` file generation) was updated to reflect the new edges. Solid black lines represent regular instruction flow, dashed blue lines indicate an intra-procedural edge. Dashed red lines indicate an inline edge (inter-procedural). Dashed green lines indicate an inter-procedural call (inter-procedural edge).
- CFG stores meta-data about mappings - see `ProgramCfg` object for more information.

During design there was a question on whether it should be the burden of the CFG or the analyses to provide functionality for walking the CFG, affecting 1. the API for doing so, 2. context sensitivity. The decision was that this would be the job of the analyses for these reasons:

1. The API change would be a significant departure from the TIP approach to running analyses, and would require modifying analyses to receive some form of iterator from a CFG. It was decided this is too large a change, and that we wish to stay close to TIP's approach. This would also be dependent on (2).
2. By abstracting it away to analyses, we can support different notions of context sensitivity without affecting other analyses. If this was done in the CFG, different APIs would have to be provided to support different ways of walking the CFG, and would require (1). This would also make the CFG stateful, and would require resetting upon each analysis run - this would then _effectively_ be equivalent to recreating a CFG for each analysis run, which is one of the things joining intra/inter-procedural CFGs hopes to avoid.


l-kent:
The intention is for this to be first merged into yousif-memory-region-analysis isn't it? Is my assistance needed with that?

yousifpatti:
Hi @l-kent 
I have merged @Thomas-Malcolm branch into memory region analysis branch which is up to date with main with minor changes. I have done some modifications to make it work with the analyses but the results coming from the MRA and VSA are a bit different atm because of the evaluate expression method that traverses the graph. Instead, I started works on passing the results of constant prop to MRA.

yousifpatti:
@Thomas-Malcolm made good comments in this PR. I believe we can merge memory region analysis branch into main instead.

l-kent:
Ok, make a pull request when you are ready to merge your branch into `main` and I will review it then. I will leave this open until then.

PR #52, opened 2023-08-22
use trim instead of strip in build file so sbt works under java8

ailrst:
This prevented sbt running at all under java8. It would probably be better to just require java11 but it doesn't look like sbt can do that.


l-kent:
We don't care about maintaining Java 8 support. We'd previously decided upon Java 17 as a minimum after some compatibility issue in the past but I guess that never made it into the ReadMe so I'll fix that now.

PR #54, opened 2023-08-30
moves old examples to separate folder

ailrst:


l-kent:
I don't really want to completely deprecate the examples until we have a good place for things that are worth keeping around but don't really fit into the new tests framework, like the cntlm example. It is worthwhile moving most of them away at least, and cleaning up the scripts.

You should also update the main readme to remove the references to the scripts you've moved.

PR #55, opened 2023-09-05
refactor cli arg parsing

ailrst:
Uses lihaoyi/mainargs to parse CLI args in stead of ad-hoc logic. Arguments are no longer positional and all require flags. This is my preference for clarity and simplicity, but isn't compatible with the previous format. 

```
BASIL
  -a --adt <str>     BAP ADT file name.
  -r --relf <str>    Output of 'readelf -s -r -W'.
  -s --spec <str>    BASIL specification file.
  -o --output <str>  Boogie output destination file.
  -v --verbose       Show extra debugging logs.
  --analyse          Run static analysis pass.
  --interpret        Run BASIL IL interpreter.
  -h --help          Show this help message.
```


l-kent:
This is good to have, but you need to update the tests for this too.

This also seems to make sbt exit entirely if you input incorrect arguments, which does not seem desirable and should be fixed.

ailrst:
Good catch, thanks, those should be fixed now

PR #58, opened 2023-09-08
BASIL-56: Fixed single return call issue

ziggyfish:
https://github.com/UQ-PAC/bil-to-boogie-translator/issues/56

### DIFF
--- a/src/main/scala/analysis/Cfg.scala
+++ b/src/main/scala/analysis/Cfg.scala
@@ -784,7 +784,7 @@ object ProgramCfg:
               }
 
               /* If the first GoTo's condition is not `True`, then we know there must be a second conditional jump.
-               *  We could equivalently check if `jmps` has 2 jumps, though this emphasises why we check for 
+               *  We could equivalently check if `jmps` has 2 jumps, though this emphasises why we check for
                *  a second jump. We can assume we have a maximum of 2 jumps, as that's what the IR (as of writing)
                *  intends.
                */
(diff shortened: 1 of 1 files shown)


ailrst:
Currently in the case of multiple possibilities for indirect calls such as in the  `functionpointer/functionpointer_clang` example all the possible direct calls are just added to the same block, this pr makes the generated cfg handle that correctly. I think instead we should modify the IR to expand the calls into multiple blocks, to ensure 1 call per block, @l-kent would you agree? 

[for reference](https://github.com/UQ-PAC/bil-to-boogie-translator/wiki/Program-Structure#block)

l-kent:
That has been previously discussed and it was decided that the current structure is acceptable. I'm not opposed to replacing conditional jumps with jumps to multiple blocks with assume statements to represent the condition (this is what already happens at the Boogie level), but I don't see any reason to change it now.

The idea that a block can only contain a maximum of two jumps, which is in the comments, is an incorrect assumption and I don't know where that came from. We need to be able to handle the case where an indirect call (such as a function pointer) can resolve to many different possible calls, and this is already accounted for in the IR.

ziggyfish:
> Currently in the case of multiple possibilities for indirect calls such as in the `functionpointer/functionpointer_clang` example all the possible direct calls are just added to the same block, this pr makes the generated cfg handle that correctly. I think instead we should modify the IR to expand the calls into multiple blocks, to ensure 1 call per block, @l-kent would you agree?
> 
> [for reference](https://github.com/UQ-PAC/bil-to-boogie-translator/wiki/Program-Structure#block)

I did it this way because you loose easy access to the prior nodes, and future nodes at the level.

ziggyfish:
> That has been previously discussed and it was decided that the current structure is acceptable. I'm not opposed to replacing conditional jumps with jumps to multiple blocks with assume statements to represent the condition (this is what already happens at the Boogie level), but I don't see any reason to change it now.

Was this decision/discusion ever documented? If not why not?

The ticket was to fix single return calls, if this has been handled at the Boogie level, then not sure the reason for the ticket.

ailrst:
> The idea that a block can only contain a maximum of two jumps, which is in the comments, is an incorrect assumption and I don't know where that came from. We need to be able to handle the case where an indirect call (such as a function pointer) can resolve to many different possible calls, and this is already accounted for in the IR.

Talking to Tom, the comment is left over from when we were intending to implement it the other way, so we should update the comments now.

l-kent:
The ticket was for fixing a bug with the CFG creation, it's still relevant. I'm talking about the structure of the IR in response to Alistair's suggestion - the IR currently supports multiple calls from a single block, which is then translated into Boogie correctly, and Boogie internally translates the resulting structure into multiple blocks. The bug is that the CFG creation does not currently support multiple jumps from a single block correctly.

The current block call structure just mimics the BAP block structure,, with each jump existing in sequence and having a condition attached. Earlier this year, I proposed moving towards a more Boogie-like block structure (with conditions moved to the start of new blocks as assume statements), as part of a broader overhaul of the IR, but after discussion we determined there wasn't much benefit while it would mean Yousif would have to rework a lot of the analysis code. I don't really see a reason to change it now. As you may have noticed, the existing documentation is not the best.

PR #59, opened 2023-09-11
Readme updates

ailrst:
Updates the readme install/setup details and removes old `run_*.sh` scripts, to reflect changed cli args.  

We also got feedback that the tool doesn't work on jdk21


l-kent:
It not working on JDK21 would be due to JDK21 only being supported by newer Scala versions.

ailrst:
Understandable.  It would be really nice if sbt could require a particular version be used but it doesn't seem simple. 

As an aside, my language server keeps warning me that Scala 3.1.0 is deprecated, migrating to 3.3.1 (the current LTS release) doesn't seem trivial but should be doable, should I try get that working at some point?

l-kent:
Yes, it would be sensible to migrate to a LTS release of Scala. Looking at the changes, it doesn't seem like migration would be too difficult - some things have been deprecated but only give warnings, and they would be possible to fix.

Issue #43, opened 2023-07-13
Incorrect Resolution of Indirect Calls


l-kent:
Currently most indirect calls that occur in the main program body (so, not boilerplate initialisation etc. code) are resolved incorrectly or not resolved at all:
in /examples:
- arrays/arrays: call R0 at statement 00000662 is resolved to be either __stck_chk_guard() or __stack_chk_fail(), when it can only be __stack_chk_fail
- arrays_simple/arrays_simple: call R0 at statement 00000607 is resolved to be either __stck_chk_guard() or __stack_chk_fail(), when it can only be __stack_chk_fail()
- function_got/function_got: call R0 at statement 000003ed is resolved to be a "call" to the global variable x, when is actually a call to the function get_two()
- functionpointer/functionpointer: call R0 at statement 000004f3 is unresolved, when it could be to set_seven(), set_six(), or set_two()
- functionpointer/functionpointer_clang: call R8 at statement 000004fd is unresolved, when it could be to set_seven(), set_six, or set_two
- jumptable/jumptable: call R0 at statements:
    - 00000595 is resolved to be a call to main() or the unidentified address 69640/0x11008, when it should be to add_two()
    - 000005a4 is resolved to be a call to main() or the unidentified address 69640/0x11008, when it should be to add_six()
    - 000005b3 is resolved to be a call to main() or the unidentified address 69640/0x11008, when it should be to sub_seven()
    - 00000789 is resolved to be a call to main(), the unidentified address 69640/0x11008, or __stack_chk_fail(), when it can only be __stack_chk_fail()
- jumptable/jumptable_clang: call R8 is unresolved at statements
    - 000004e1 - which should be add_two()
    - 000004f0 - which should be add_six()
    - 000004ff - which should be sub_seven()
- jumptable2/jumptable2_clang: call R8 at statements:
    - 000004d0 is resolved to be a "call" to the global variable x when it should be add_two()
    - 000004e6 is resolved to be a "call" to the global variable x when it should be add_six()
    - 000004fc is resolved to be a "call" to the global variable x when it should be sub_seven()
- jumptable2/jumptable2: call R0 is unresolved at statements:
    - 000004f7 - which should be add_two()
    - 00000512 - which should be add_six()
    - 0000052d - which should be sub_seven()
- indirect_call/indirect_call
    - call R1 at statement 000003ad is resolved to be a call to main() when it should be a call to puts()
[...]

PR #61, opened 2023-09-12
Merge analysis-devel: Unified CFG and Static Analysis

ailrst:
This contains Yousif's masters thesis work and the unified intraprocedural and interprocedural CFG. 

We will work on rebasing this branch on main before merging

- [x] https://github.com/UQ-PAC/bil-to-boogie-translator/issues/65
- [x] merge main / fix conflicts
- [x] run autoformatter
- [x] make println calls logging calls


ailrst:
Merge conflicts have been resolved here https://github.com/UQ-PAC/bil-to-boogie-translator/tree/prepare-analysis-devel-merge

ailrst:
@l-kent I think this should be OK to look at now, the remaining PRs into `analysis-devel` can be merged into main afterwards since they can affect the boogie generation and might need a bit more care before merging.

ailrst:
Test output for boogie v2 on my machine https://gist.github.com/ailrst/324be1f5ec8de2f0ea0256cb1d6984ac

l-kent:
If you run `reload` and `clean` they should work?

Is there a summary of what's included in these changes?

src/main/scala/ir/Expr.scala
```diff
@@ -354,9 +355,13 @@ sealed trait Variable extends Expr {
...
   def size: Int = irType match {
     case b: BitVecType => b.size
-    case _ => throw new Exception("tried to get size of non-bitvector")
+    case _             => throw new Exception("tried to get size of non-bitvector")
   }
 
+  val isRegister: Boolean = name.startsWith("R") || name.startsWith("V") && name != "VF"
+
+  // `rName` : e.g. "R20"
+  def isRegister(rName: String): Boolean = isRegister && rName.equals(name)
```
l-kent, line 364: This is redundant now that we have a separate Register case class, and that should be used instead

src/main/scala/translating/ReadELFLoader.scala
```diff
@@ -70,7 +70,7 @@ object ReadELFLoader {
   }
 
   def visitSymbolTableRow(ctx: SymbolTableRowContext): Option[SpecGlobal] = {
-    if (ctx.entrytype.getText == "OBJECT" && ctx.bind.getText == "GLOBAL" && ctx.vis.getText == "DEFAULT") {
+    if (ctx.entrytype.getText == "OBJECT" && ctx.bind.getText == "GLOBAL") {
```
l-kent, line 73: What's this needed for?

src/main/scala/util/RunUtils.scala
```diff
@@ -120,105 +144,167 @@ object RunUtils {
...
+      IRProgram: Program
+  ): (Program, Boolean) = {
+    var modified: Boolean = false
+    val worklist = ListBuffer[CfgNode]()
+    // find the main function
+    cfg.nodes.foreach {
+      case main: CfgFunctionEntryNode if main.data.name == "main" =>
+        main.succ(true).toSet.union(main.succ(false).toSet).foreach(node => worklist.addOne(node))
+      case _ =>
+    }
```
l-kent, line 210: This should use Program.mainProcedure to find the starting procedure now, instead of just finding the name "main"

src/main/scala/util/RunUtils.scala
```diff
@@ -120,105 +144,167 @@ object RunUtils {
...
     def process(n: CfgNode): Unit = n match {
       case commandNode: CfgCommandNode =>
         commandNode.data match
+          case localAssign: LocalAssign =>
+            localAssign.rhs match
+              case _: MemoryLoad =>
+                if (valueSets(n).contains(localAssign.lhs) && valueSets(n).get(localAssign.lhs).head.size == 1) {
+                  val extractedValue = extractExprFromValue(valueSets(n).get(localAssign.lhs).head.head)
+                  localAssign.rhs = extractedValue
+                  println(s"RESOLVED: Memory load ${localAssign.lhs} resolved to ${extractedValue}")
```
l-kent, line 238: Use logger?

src/main/scala/util/RunUtils.scala
```diff
@@ -120,105 +144,167 @@ object RunUtils {
...
+                } else if (valueSets(n).contains(localAssign.lhs) && valueSets(n).get(localAssign.lhs).head.size > 1) {
+                  println(s"RESOLVED: WARN Memory load ${localAssign.lhs} resolved to multiple values, cannot replace")
```
l-kent, line 240: Use logger?

src/main/scala/util/RunUtils.scala
```diff
@@ -120,105 +144,167 @@ object RunUtils {
...
                   block.jumps = block.jumps.filter(!_.equals(indirectCall))
-                  block.jumps += DirectCall(IRProgram.procedures.filter(_.name.equals(functionNames.head.name)).head, indirectCall.condition, indirectCall.returnTarget)
-                case _ => throw new Exception("Node not found in nodeToBlock map")
-            } else {
+                  block.jumps += DirectCall(
+                    IRProgram.procedures.filter(_.name.equals(functionNames.head.name)).head,
+                    indirectCall.condition,
+                    indirectCall.returnTarget
+                  )
+                case null => throw new Exception("Node not found in nodeToBlock map")
```
l-kent, line 273: Using null is bad practice, we should be using Option types and None instead

src/main/scala/util/RunUtils.scala
```diff
@@ -120,105 +144,167 @@ object RunUtils {
...
                     } else {
-                      block.jumps += DirectCall(IRProgram.procedures.filter(_.name.equals(addressValue.name)).head, Option(BinaryExpr(BVEQ, indirectCall.target, addressValue.expr)), indirectCall.returnTarget)
+                      block.jumps += DirectCall(
+                        IRProgram.procedures.filter(_.name.equals(addressValue.name)).head,
+                        Option(BinaryExpr(BVEQ, indirectCall.target, addressValue.expr)),
+                        indirectCall.returnTarget
+                      )
                     }
-                  case _ => throw new Exception("Node not found in nodeToBlock map")
+                  case null => throw new Exception("Node not found in nodeToBlock map")
```
l-kent, line 299: null

src/main/scala/analysis/solvers/UnionFindSolver.scala
```diff
@@ -61,45 +56,41 @@ class UnionFindSolver[A] {
...
     */
   override def toString: String =
     solution().map(p => s"${p._1} = ${p._2}").mkString("\n")
 }
 
-/**
-  * Exception thrown in case of unification failure.
+/** Exception thrown in case of unification failure.
   */
 class UnificationFailure(message: String = null) extends RuntimeException(message)
```
l-kent, line 96: Probably should define new Exceptions scala-style (there are still some at the bottom of RunUtils that aren't in use if you want an example) rather than just extending RuntimeException and having a null default

src/main/scala/analysis/solvers/FixPointSolver.scala
```diff
@@ -87,19 +89,33 @@ trait ListSetWorklist[N] extends Worklist[N]:
...
-      val n = worklist.head;
-      worklist = worklist.tail
-      process(n)
+      process(n, intra)
+
+  def monotonic_run(first: Set[N], intra: Boolean) =
+    if (!intra) {
+      worklist =
+        new ListSet[N] ++ first.collect {
+          case n: CfgFunctionEntryNode if n.data.name == "main" => n
```
l-kent, line 103: Shouldn't be using "main" to find the initial procedure now that we have Program.mainProcedure

src/main/scala/analysis/Cfg.scala
```diff
@@ -51,382 +196,930 @@ trait CfgNode:
...
+            if (targetCond != TrueLiteral) {
+              val secondGoto: GoTo = jmps.tail.head.asInstanceOf[GoTo]
+              targetBlock = secondGoto.target
+              // IR doesn't store negation of condition, so we must do it manually
+              targetCond = negateConditional(targetCond)
+
+              // Jump to target block
+              if (visitedBlocks.contains(targetBlock)) {
+                val targetBlockEntry: CfgCommandNode = visitedBlocks(targetBlock)
+                cfg.addEdge(precNode, targetBlockEntry, targetCond)
```
l-kent, line 797: Maximum of 2 jumps is not a safe assumption. We should try to handle the jumps in a more robust way than this

src/main/scala/analysis/Cfg.scala
```diff
@@ -51,382 +196,930 @@ trait CfgNode:
...
+            // Branch to this call
+            val calls = jmps.filter(_.isInstanceOf[DirectCall]).map(x => CfgJumpNode(data = x, block = block))
+
+            calls.foreach(node => {
+              cfg.addEdge(precNode, node, node.data.asInstanceOf[DirectCall].condition.getOrElse(TrueLiteral))
+
+              procToCalls(proc) += node
+              procToCallers(targetProc) += node
+              callToNodes(funcEntryNode) += node
+            })
```
l-kent, line 815: probably better to do this in a way without multiple casts?

src/main/scala/analysis/Cfg.scala
```diff
@@ -51,382 +196,930 @@ trait CfgNode:
...
+        * @param prevNode
+        *   Either the previous statement in the block, or the previous block's end node (in the case that this block
+        *   contains no statements)
+        * @param cond
+        *   Jump from `prevNode` to this. `TrueLiteral` if `prevNode` is a statement, and any `Expr` if `prevNode` is a
+        *   jump.
+        * @param solitary
+        *   `True` if this block contains no statements, `False` otherwise
+        */
+      def visitJumps(jmps: List[Jump], prevNode: CfgNode, cond: Expr, solitary: Boolean): Unit = {
```
l-kent, line 733: why use List[Jump] when we are just converting an ArrayBuffer when this is called in visitBlock

src/main/scala/analysis/Cfg.scala
```diff
@@ -51,382 +196,930 @@ trait CfgNode:
...
+    override val predInter: mutable.Set[CfgEdge] = mutable.Set[CfgEdge](),
+    override val succIntra: mutable.Set[CfgEdge] = mutable.Set[CfgEdge](),
+    override val succInter: mutable.Set[CfgEdge] = mutable.Set[CfgEdge](),
+    data: NOP = NOP(),
+    block: Block
+) extends CfgCommandNode:
+  override def toString: String = s"[NOP]"
+
+  /** Copy this node, but give unique ID and reset edges */
+  override def copyNode(): CfgGhostNode = CfgGhostNode(block = this.block)
```
l-kent, line 357: This isn't actually used for anything is it? Do we need this?

src/main/scala/ir/Statement.scala
```diff
@@ -25,6 +27,11 @@ class MemoryAssign(var lhs: Memory, var rhs: MemoryStore) extends Statement {
   override def acceptVisit(visitor: Visitor): Statement = visitor.visitMemoryAssign(this)
 }
 
+class NOP() extends Statement {
+  override def toString: String = "<NOP>"
+  override def acceptVisit(visitor: Visitor): Statement = this
+}
```
l-kent, line 33: Do we need this at all? Regardless, it should be an Object not a Class as there aren't distinct instances

src/main/scala/analysis/VSA.scala
```diff
@@ -105,239 +68,97 @@ trait MemoryRegionValueSetAnalysis:
...
+                    m + (localAssign.lhs -> m(r))
+                  case _ =>
+                    s + (localAssign.lhs -> s(r))
+              case None =>
+                Logger.warn("could not find region for " + localAssign)
+                s
+          case e: Expr => {
+            val evaled = evaluateExpression(e, n, constantProp)
+            evaled match
+              case bv: BitVecLiteral => s + (localAssign.lhs -> Set(getValueType(evaled.asInstanceOf[BitVecLiteral])))
```
l-kent, line 132: Should just use `bv` instead of `evaled.asInstanceOf[BitVecLiteral]`

src/main/scala/analysis/VSA.scala
```diff
@@ -105,239 +68,97 @@ trait MemoryRegionValueSetAnalysis:
...
+              case _ =>
+                println("Warning: could not evaluate expression" + e)
```
l-kent, line 134: Use logger?

src/main/scala/analysis/UtilMethods.scala
```diff
@@ -0,0 +1,52 @@
...
+      val rhs = evaluateExpression(binOp.arg2, n, constantProp)
+      if (!lhs.isInstanceOf[BitVecLiteral] || !rhs.isInstanceOf[BitVecLiteral]) {
+        return exp
+      }
+      binOp.op match
+        case BVADD  => BitVectorEval.smt_bvadd(lhs.asInstanceOf[BitVecLiteral], rhs.asInstanceOf[BitVecLiteral])
+        case BVSUB  => BitVectorEval.smt_bvsub(lhs.asInstanceOf[BitVecLiteral], rhs.asInstanceOf[BitVecLiteral])
+        case BVASHR => BitVectorEval.smt_bvashr(lhs.asInstanceOf[BitVecLiteral], rhs.asInstanceOf[BitVecLiteral])
+        case BVCOMP => BitVectorEval.smt_bvcomp(lhs.asInstanceOf[BitVecLiteral], rhs.asInstanceOf[BitVecLiteral])
+        case _      => throw new RuntimeException("Binary operation support not implemented: " + binOp.op)
```
l-kent, line 31: If casting like this is necessary there should at least be new variables created instead of casting `lhs` and `rhs` on every use.

src/main/scala/analysis/SSA.scala
```diff
@@ -1,426 +1,426 @@
-package analysis
+// TODO: This doesn't seem to be used currently, should it be kept?
```
l-kent, line 1: Good question, is this fine to completely remove?

src/main/scala/analysis/MemoryModelMap.scala
```diff
@@ -21,23 +23,23 @@ class MemoryModelMap {
...
         }
       case d: DataRegion =>
         val dataMap = rangeMap.dataMap
         if (dataMap.isEmpty) {
-          dataMap(RangeKey(offset, MAX_BIGINT)) = regionType
+          dataMap(RangeKey(offset, MAX_BIGINT)) = region.asInstanceOf[DataRegion]
         } else {
           dataMap.keys.maxBy(_.end).end = offset - 1
-          dataMap(RangeKey(offset, MAX_BIGINT)) = regionType
+          dataMap(RangeKey(offset, MAX_BIGINT)) = region.asInstanceOf[DataRegion]
```
l-kent, line 42: You can just use `s` and `d` instead of casting region

src/main/scala/analysis/Cfg.scala
```diff
@@ -51,382 +196,930 @@ trait CfgNode:
...
+          n.data match {
+            case d: DirectCall =>
+              procToCalls(proc) += newNode
+                .asInstanceOf[CfgCommandNode] // This procedure (general) is calling another procedure
+              callToNodes(newEntry) += newNode
+                .asInstanceOf[CfgCommandNode] // This procedure (specfic) is calling another procedure
+              procToCallers(d.target) += newNode.asInstanceOf[CfgCommandNode] // Target of this call has a new caller
+            case i: IndirectCall =>
+              procToCalls(proc) += newNode.asInstanceOf[CfgCommandNode]
+              callToNodes(newEntry) += newNode.asInstanceOf[CfgCommandNode]
```
l-kent, line 1084: Why not type match before copying, instead of copying, type matching the original, and then casting the copy on every use?

src/main/scala/analysis/Cfg.scala
```diff
@@ -51,382 +196,930 @@ trait CfgNode:
...
-              } else {
-                cfg.addEdge(lastAdded, functionExitNode)
-              }
+            // Record call association
+            procToCalls(proc) += jmpNode
+            callToNodes(funcEntryNode) += jmpNode
+
+            // R30 is the link register - this stores the address to return to.
+            //  For now just add a node expressing that we are to return to the previous context.
+            if (iCall.target.isRegister("R30")) {
```
l-kent, line 853: Shouldn't use `isRegister` now that we have a separate Register case class

src/main/scala/analysis/Analysis.scala
```diff
@@ -500,206 +495,149 @@ trait MemoryRegionAnalysisMisc:
...
+            Set(DataRegion(s"Unknown_${bitVecLiteral}", bitVecLiteral, None))
           }
+        } else {
+          //throw new Exception(s"Unknown type for $bitVecLiteral")
+          // unknown region here
+          Set(DataRegion(s"Unknown_${bitVecLiteral}", bitVecLiteral, None))
+        }
+      case variable: Variable =>
+        if (variable.name.contains("#") || variable.equals(stackPointer)) {
+          return env
```
l-kent, line 538: rather than using `variable.name.contains("#")` we should match the `Register` and `LocalVar` types as relevant here

src/main/scala/analysis/Analysis.scala
```diff
@@ -44,30 +45,30 @@ trait ValueAnalysisMisc:
...
-          case BVMUL => bvmul(left, right)
+        bin.op match
+          case BVADD  => bvadd(left, right)
+          case BVSUB  => bvsub(left, right)
+          case BVMUL  => bvmul(left, right)
           case BVUDIV => bvudiv(left, right)
           case BVSDIV => bvsdiv(left, right)
           case BVSREM => bvsrem(left, right)
           case BVUREM => bvurem(left, right)
           case BVSMOD => ??? // Signed modulus. NOTE: this is used.
```
l-kent, line 64: This should be supported?

src/main/scala/analysis/VSA.scala
```diff
@@ -105,239 +68,97 @@ trait MemoryRegionValueSetAnalysis:
...
+                // this is an exception to the rule and only applies to data regions
+                evaluateExpression(memoryLoad.index, n, constantProp) match
+                  case bitVecLiteral: BitVecLiteral =>
+                    val m = s + (r -> Set(getValueType(bitVecLiteral)))
+                    m + (localAssign.lhs -> m(r))
+                  case _ =>
+                    s + (localAssign.lhs -> s(r))
+              case None =>
+                Logger.warn("could not find region for " + localAssign)
+                s
```
l-kent, line 128: What is this method actually doing? There are lots of different layers of evaluation and it's hard to tell how it works, or why only some cases are supported

src/main/scala/analysis/Analysis.scala
```diff
@@ -500,206 +495,149 @@ trait MemoryRegionAnalysisMisc:
...
+  /** Default implementation of eval.
+    */
+  def eval(exp: Expr, env: lattice.sublattice.Element, n: CfgNode): lattice.sublattice.Element = {
+    Logger.debug(s"evaluating $exp")
+    Logger.debug(s"env: $env")
+    Logger.debug(s"n: $n")
+    exp match {
+      case binOp: BinaryExpr =>
+        if (binOp.arg1 == stackPointer) {
+          val rhs: Expr = evaluateExpression(binOp.arg2, n, constantProp)
```
l-kent, line 507: Pointless line?

src/main/scala/analysis/UtilMethods.scala
```diff
@@ -0,0 +1,52 @@
...
+/** Evaluate an expression in a hope of finding a global variable.
+  *
+  * @param exp
+  *   : The expression to evaluate (e.g. R1 + 0x1234)
+  * @param n
+  *   : The node where the expression is evaluated (e.g. mem[R1 + 0x1234] <- ...)
+  * @return:
+  *   The evaluated expression (e.g. 0x69632)
+  */
+def evaluateExpression(exp: Expr, n: CfgNode, constantProp: Map[CfgNode, Map[Variable, Any]]): Expr = {
```
l-kent, line 17: Why not pass in the constant propagation results just for the node we're evaluating? Is there a reason this method only supports such a limited amount of operators? Is it not possible to build a more type-robust evaluator that will only deal with the operators that have bitvector inputs & outputs so there aren't casts everywhere surrounding its usage?

examples/pthread_basic/pthread.bil
l-kent, line 1: should be .bir - .bil is a different BAP output format that we aren't using

.gitignore
```diff
@@ -21,4 +21,5 @@ src/main/antlr4/.antlr
 result.txt
 examplesold/
 src/test/scala/dump/
-src/test/analysis/dump/
\ No newline at end of file
+src/test/analysis/dump/
+-analyse
```
l-kent, line 25: What's this for?

src/main/scala/analysis/MemoryModelMap.scala
```diff
@@ -49,19 +51,17 @@ class MemoryModelMap {
       val node = memoryRegions(exitNode).asInstanceOf[Set[Any]]
       // for each function exit node we get the memory region
       // and add it to the mapping
-      val stackRgns = node.collect{ case r: StackRegion => r }.toList.sortBy(_.start.asInstanceOf[BitVecLiteral].value)
-      val dataRgns = node.collect{ case r: DataRegion => r }.toList
-      val heapRgns = node.collect{ case r: HeapRegion => r }.toList
-      val accessRgns = node.collect{ case r: RegionAccess => r }.toList
+      val stackRgns = node.collect { case r: StackRegion => r }.toList.sortBy(_.start.asInstanceOf[BitVecLiteral].value)
```
l-kent, line 54: Why are we just assuming that StackRegion.start is a BitVecLiteral? If this is safe to do, shouldn't that be the type of it in the first place?

src/main/scala/analysis/VSA.scala
```diff
@@ -105,239 +68,97 @@ trait MemoryRegionValueSetAnalysis:
...
-        val rhs: Expr = evaluateExpression(binOp.arg2, n)
-        if (lhs.equals(stackPointer)) {
-          mmm.findObject(rhs.asInstanceOf[BitVecLiteral].value, "stack") match
-            case Some(obj: MemoryRegion) =>
-              return Some(obj)
-            case _ =>
-              return None
+        if (binOp.arg1 == stackPointer) {
+          val rhs: Expr = evaluateExpression(binOp.arg2, n, constantProp)
+          mmm.findStackObject(rhs.asInstanceOf[BitVecLiteral].value) match
```
l-kent, line 76: more assuming that StackRegion.value is a BitVecLiteral

src/main/scala/util/RunUtils.scala
```diff
@@ -68,38 +82,48 @@ object RunUtils {
       Interpret(IRProgram)
     }
 
+    Logger.info("[!] Removing external function calls")
+    // Remove external function references (e.g. @printf)
     val externalNames = externalFunctions.map(e => e.name)
     val externalRemover = ExternalRemover(externalNames)
+    // Removes BAP naming artefacts (e.g. # preceding variable names)
```
l-kent, line 89: That isn't what it does, it's renaming things that clash with reserved Boogie keywords

ailrst:
Tests are doing better after clean and reload https://gist.github.com/ailrst/94fc10db9b129b849724d988f6209627

src/main/scala/analysis/VSA.scala
```diff
@@ -105,239 +68,97 @@ trait MemoryRegionValueSetAnalysis:
...
-                return None
-            }
-        None
+          val evaluation: Expr = evaluateExpression(binOp, n, constantProp)
+          if (!evaluation.isInstanceOf[BitVecLiteral]) {
+            return None
+          }
+          mmm.findDataObject(evaluation.asInstanceOf[BitVecLiteral].value) match
+            case Some(obj: MemoryRegion) => Some(obj)
+            case _                       => None
```
l-kent, line 86: This is redundant? If findDataObject doesn't return a value of type Some[DataRegion] (and DataRegion is a subclass of MemoryRegion) it will return None so there isn't a need to match it?

l-kent (commented):
I think I've identified the obvious issues, but there may be more from very old code that aren't in this set of changes. I haven't really tested the functionality of the analysis yet - what's the state of it? Which examples are good for demonstrating that functionality?

src/main/scala/analysis/VSA.scala
ailrst: Yeah good point, the notion of "evaluation" as it pertains to the MRA, VSA and CPA should be clearer in the code & documentation. This is really just defining the local transform for assignments. I'm also unclear why only some cases are supported and what it would take to be sure this is a total function.

ailrst:
Thanks for the thorough review I'll go through and fix everything when I get a chance. 

The analysis needs some work mainly for the region detection and framing issues we've talked about before. Like you pointed out, there are a few partial functions over the expressions/statements in the IL and I'm also not 100% sure what the correct approach is with those, we should probably implement and formalize the analyses for the whole language.    

The best examples are probably the indirect call resolution examples you've already looked at and Yousif has addressed https://github.com/UQ-PAC/bil-to-boogie-translator/issues/43. Maybe Yousif has better recommendations.

src/main/scala/analysis/VSA.scala
l-kent: Yeah, I understand that this the local transform for the VSA, but it's a hard to get a good sense of what that's trying to do (the same does go for some of the other eval functions too). It seems very dependent on certain patterns in a way that seems fairly fragile to me?

l-kent:
So out of the examples in #43, the indirect call resolution still doesn't work for:

- jumptable/jumptable_clang - the analysis doesn't terminate
- jumptable/jumptable - fails to resolve an indirect call
- jumptable2/jumptable2 and jumptable2/jumptable2_clang - resolves calls incorrectly to be to the global variable 'jump_table'

l-kent:
I think before merging the main things to address are just the style/clarity issues. We can make note of the functionality that still needs to be addressed, such as the indirect calls that are still not resolved.

ailrst:
Yeah thats reasonable

> jumptable/jumptable_clang - the analysis doesn't terminate

I'd like to look at this one first though, and there's a couple more low hanging functionality issues like the partial eval() implementation I'd like to fix as well.

l-kent:
Yeah, it makes sense to look at the non-termination there. I don't know if dealing with the partial eval() implementations is necessary - that seems like it might be more involved?

ailrst:
> I don't know if dealing with the partial eval() implementations is necessary - that seems like it might be more involved?

Sorry I only meant the missing binops, fully fixing it would be more work yeah

src/main/scala/analysis/VSA.scala
yousifpatti: Fixed [Upcoming merge from yousif-memory-region-analysis]

src/main/scala/analysis/VSA.scala
yousifpatti: Fixed [Upcoming merge from yousif-memory-region-analysis]

src/main/scala/analysis/UtilMethods.scala
yousifpatti: Fixed [Upcoming merge from yousif-memory-region-analysis]

src/main/scala/analysis/SSA.scala
yousifpatti: Functionality defined is deprecated. File has been removed. [Upcoming merge from yousif-memory-region-analysis]

src/main/scala/analysis/MemoryModelMap.scala
yousifpatti: Resolved. [Upcoming merge from yousif-memory-region-analysis]

src/main/scala/analysis/Cfg.scala
yousifpatti: Resolved. [Upcoming merge from yousif-memory-region-analysis]

src/main/scala/analysis/Analysis.scala
yousifpatti: Resolved. [Upcoming merge from yousif-memory-region-analysis]

src/main/scala/analysis/Analysis.scala
yousifpatti: Resolved. [Upcoming merge from yousif-memory-region-analysis]
rhs should be passed instead of binOp.arg2

.gitignore
yousifpatti: Resolved. [Upcoming merge from yousif-memory-region-analysis]
The `-analyse` no longer applies but the `src/test/analysis/dump/` was used when running MRA tests to generate dot output files and compare them with the `dotExpected` folder files

l-kent:
@yousifpatti Which upcoming merge have you resolved those issues in? I don't see it those in #86

yousifpatti:
> @yousifpatti Which upcoming merge have you resolved those issues in? I don't see those in #86

I was waiting to finish looking at all your comments before I push.
I pushed the ones I have commented on now in #86
I will work on the rest

l-kent:
Ah, I was trying to fix up most of the issues here today because that's what I'd said I'd do in the meeting. I want to get this part merged into main before having to look at your memory region analysis changes as well, so it's a bit more difficult if you've made all the fixes in your own branch where you've also changed the memory region analysis.

l-kent:
Can you commit those fixes to this branch instead?

yousifpatti:
Can you jump on zoom for a minute please? https://uqz.zoom.us/j/88543683225

l-kent:
It's probably not a good idea to post zoom links publicly like that, you should send me an email.

yousifpatti:
I will send you an email

src/main/scala/translating/ReadELFLoader.scala
l-kent: @yousifpatti

src/main/scala/analysis/Cfg.scala
l-kent: @yousifpatti

src/main/scala/ir/Statement.scala
l-kent: @yousifpatti

src/main/scala/analysis/VSA.scala
l-kent: @yousifpatti

src/main/scala/analysis/Cfg.scala
yousifpatti: I do not think this is needed as blocks without stmts is being dealt with without adding intermediate nodes

src/main/scala/analysis/MemoryModelMap.scala
l-kent: @yousifpatti

src/main/scala/ir/Statement.scala
yousifpatti: It is just a placeholder (empty stmt). The data field could be made optional but that would require more refactoring. I agree with making it an object instead.

src/main/scala/analysis/MemoryModelMap.scala
yousifpatti: Yes it should be safe to assume so the type could change

l-kent: I think this is one that will make sense to come back to later, I'll leave it for now

src/main/scala/analysis/VSA.scala
yousifpatti: As MRA ignores some regions (regions produced by a LocalVar or specific register assigns like R29) this results in `exprToRegion()` to produce `None` so it is either the pattern matching or doing the same thing in MRA with `ignoredRegions`

l-kent: It would be helpful if you could add comments for more complicated methods like this one to make it clear what they're doing, thanks

l-kent:
This isn't everything but it's most of it. I want to try to clean up some of the messiness around the types.

l-kent:
Handled #95, will merge this soon.

Still some outstanding problems but I'll make tickets for them

l-kent:
The non-termination issue with jumptable/jumptable_clang appears to have been resolved, but it still fails to resolve the indirect call.

Issue #64, opened 2023-09-12
Non-Deterministic Branching in IR


l-kent:
It may be necessary at some point to add a new NonDetGoTo (or similar) class to the IR, a subclass of Jump. It should have multiple targets, and be translated to Boogie as `goto target1, target2, target3;' which is how Boogie represents non-deterministic jumps.

l-kent:
Implemented in #132

PR #74, opened 2023-09-19
Integrating MRA and VSA Results with IR

yousifpatti:
- Replaces stack memory stores/loads with "stack" instead of "mem"
- Replaces memLoads with concrete values


ailrst:
This looks good except for the OR between multiple VSA possibilities, I don't think we can use Exprs to represent nondeterministic choice this way

l-kent:
Yeah, you can't represent non-deterministic choice that way. If an access could be to multiple regions, then you need to merge those regions together in the IR.

l-kent:
To do non-deterministic choice you need to add a block for each choice, and a GoTo that points to both blocks (`goto Block1, Block2;` in Boogie. This is not currently supported in the IR but would be easy to implement - see #64. You would also need to add Assumes (also not in the IR yet but easy to implement) to the start of each of the choices to represent the branching conditions, which would be that the acces index equals each of the possible values corresponding to different regions.

This is probably worth experimenting with in comparison to merging the regions, but I'm not sure how effective it would be. It might prove difficult with say, heap accesses where you don't have a static value; I'm not sure.

We will likely need non-deterministic choice to some extent regardless, in order to maintain stack regions separate from other memory regions.

l-kent:
We also don't want to be replacing the `MemoryLoad`s with concrete values like that, as discussed yesterday. Our aim is to replace the `mem` variable in `MemoryLoad`s and `MemoryStore`s with a new `Memory` variable that represents the region being accessed in each case.

PR #76, opened 2023-09-20
Basil 12 - Thread Identification

ziggyfish:


l-kent:
We can't really merge this into main until the analysis branch is merged into main properly, since this is a fork of one of those branches.

ailrst:
> We can't really merge this into main until the analysis branch is merged into main properly, since this is a fork of one of those branches.

We'll merge into analysis devel first

ailrst:
I'd prefer to use value set analysis result rather than the constant propagation analysis result; this means we make a more conservative estimate of the possibly-concurrent procedures (we get all the possible values in constant-prop plus more). Plus this matches how we resolve regular indirect calls.

ziggyfish:
@ailrst I agree we should use VSA, however register R2 doesn't exist in the VSA but does exist in the constant propegation analysis. So work needs to be done, before we can use constant VSA.

ailrst:
> @ailrst I agree we should use VSA, however register R2 doesn't exist in the VSA but does exist in the constant propegation analysis. So work needs to be done, before we can use constant VSA.

That means the VSA is most likely not working correctly, I'll have to look into it

l-kent:
What is the status of this? Is it working except we want to use VSA instead of constant propagation but there is an issue preventing that? If someone explains the problem with the VSA in further depth and creates an issue for it, I can try to fix that.

ailrst:
VSA should be fixed now in analysis-devel and the main branch

ziggyfish:
Looks like there are a few issues still with the VSA, so I can't do anything with this yet, and how its structured.

So seems that this code also has many conflicts other changes in the code.

So I'll leave it with you guys to make the nessecary changes to get this into a state where it can be merged with main.

Some documentation:

The most of the work is done in analysis.DirectCallAnalysis and analysis.call.ThreadAnalysis.

Essentially it simply looks at the value of R2 and uses this as the entry point into the thread analysis.

Anyway I'll leave it with the rest of the team to handle the required changes.

l-kent:
Can you explain what the issues are with the VSA? In main, R2 is definitely in the VSA result for the pthread example at the pthread_create callsite, so it isn't the issue you were previously talking about.

l-kent:
This is not the approach that will make sense for dealing with threads

PR #78, opened 2023-09-20
Evaluate literal expressions in vsa

ailrst:
Adds local assign expressions that can be evaluated to literals to the VSA 
Functionally this is just merging the constant propagation results into the vsa

Before commit 93d99378cd96185c71f677f72a7bc3dce3cc6a20 the VSA tracked values of registers, but unclear if this was broken in a merge later or something else.

### DIFF
--- a/src/main/scala/analysis/VSA.scala
+++ b/src/main/scala/analysis/VSA.scala
@@ -128,7 +128,14 @@ trait MemoryRegionValueSetAnalysis:
               case None =>
                 println("Warning: could not find region for " + localAssign)
                 s
-          case _ => s
+          case e: Expr => {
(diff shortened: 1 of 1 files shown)


ailrst:
@yousifpatti I think generally you were trying to delegate to the constantprop results rather than duplicate in the VSA is this a correct fix to make? 

We could validly repeat the same analysis in the VSA to get the same results, or just get the result directly from constant prop

yousifpatti:
Good catch. You are right I was trying to not duplicate the results in the VSA but I think doing so will give us the more accurate VSA implementation (should be tracking constants too) which will save us from having to merge the results of constant prop and VSA.

Your fix LGTM 👍

PR #79, opened 2023-09-20
Serialise IL

ailrst:
For debugging and regression detection it is good to have a text output of the intermediate language, this is a basic implementation of that, aiming to make it look close to the internal datastructures with some explanatory syntax.

E.g. for secret_write: 

```
Procedure("1980:main", in(Parameter(Register(R0, bv64)), Parameter(Register(R1, bv64))), out(Parameter(Register(R1, bv64)), Parameter(Register(R0, bv64))), blocks(
  Block("1980:lmain",
    statements(
      LocalAssign(LocalVar(#5, bv64) := Expr(BinaryExpr(Expr(Register(R31, bv64)), Expr(18446744073709551584bv64))))
      MemoryAssign(MemoryStore(Memory(Memory(mem, 64, 8))[Expr(LocalVar(#5, bv64))] := Expr(Register(R29, bv64))))
      MemoryAssign(MemoryStore(Memory(Memory(mem, 64, 8))[Expr(BinaryExpr(Expr(LocalVar(#5, bv64)), Expr(8bv64)))] := Expr(Register(R30, bv64))))
      LocalAssign(Register(R31, bv64) := Expr(LocalVar(#5, bv64)))
      LocalAssign(Register(R29, bv64) := Expr(Register(R31, bv64)))
      LocalAssign(Register(R0, bv64) := Expr(65536bv64))
      LocalAssign(Register(R0, bv64) := Expr(MemoryLoad(Memory(Memory(mem, 64, 8)), [Expr(BinaryExpr(Expr(Register(R0, bv64)), Expr(4056bv64)))])))
      MemoryAssign(MemoryStore(Memory(Memory(mem, 64, 8))[Expr(BinaryExpr(Expr(Register(R31, bv64)), Expr(24bv64)))] := Expr(Register(R0, bv64))))
      LocalAssign(Register(R0, bv64) := Expr(0bv64))
      LocalAssign(Register(R0, bv64) := Expr(BinaryExpr(Expr(Register(R0, bv64)), Expr(2088bv64))))
      LocalAssign(Register(R1, bv64) := Expr(65536bv64))
      LocalAssign(Register(R1, bv64) := Expr(MemoryLoad(Memory(Memory(mem, 64, 8)), [Expr(BinaryExpr(Expr(Register(R1, bv64)), Expr(4072bv64)))])))
      LocalAssign(Register(R30, bv64) := Expr(2020bv64))
    ),
    jumps(
      IndirectCall(Register(R1, bv64), condition())
    )
  )
  Block("2020:l000003af",
    statements(
      LocalAssign(Register(R0, bv64) := Expr(MemoryLoad(Memory(Memory(mem, 64, 8)), [Expr(BinaryExpr(Expr(Register(R31, bv64)), Expr(24bv64)))])))
      LocalAssign(Register(R30, bv64) := Expr(2028bv64))
    ),
    jumps(
      IndirectCall(Register(R0, bv64), condition())
    )
  )
  Block("2028:l000003be",
    statements(
      LocalAssign(Register(R0, bv64) := Expr(0bv64))
      LocalAssign(Register(R29, bv64) := Expr(MemoryLoad(Memory(Memory(mem, 64, 8)), [Expr(Register(R31, bv64))])))
      LocalAssign(Register(R30, bv64) := Expr(MemoryLoad(Memory(Memory(mem, 64, 8)), [Expr(BinaryExpr(Expr(Register(R31, bv64)), Expr(8bv64)))])))
      LocalAssign(Register(R31, bv64) := Expr(BinaryExpr(Expr(Register(R31, bv64)), Expr(32bv64))))
    ),
    jumps(
      IndirectCall(Register(R30, bv64), condition())
    )
  )
)),
```

### DIFF
--- a/src/main/scala/translating/ILtoIL.scala
+++ b/src/main/scala/translating/ILtoIL.scala
@@ -8,7 +8,7 @@ private class ILSerialiser extends Visitor {
   var indentLevel = 0
 
   def getIndent(): String = {
-    "    " * indentLevel
+    "  " * indentLevel
   }
 
   def blockIdentifier(block: Block) : String = {
@@ -93,7 +93,7 @@ private class ILSerialiser extends Visitor {
     program ++= "IndirectCall("
     node.target = visitVariable(node.target)
     program ++= ", "
-    program ++= "Condition("
+    program ++= "condition("
     node.condition = node.condition.map(visitExpr)
     program ++= ")" // Condition
     program ++= ")" // IndirectCall 
@@ -152,7 +152,7 @@ private class ILSerialiser extends Visitor {
     for (i <- node.blocks.indices) {
       node.blocks(i) = visitBlock(node.blocks(i))
     }
-    program ++= "))\n"
+    program ++= ")),\n"
     indentLevel -= 1
     node
   }
@@ -250,7 +250,6 @@ private class ILSerialiser extends Visitor {
   override def visitVariable(node: Variable): Variable = {
     program ++= node.toString()
     node
-    //node.acceptVisit(this)
   }
 
   override def visitRegister(node: Register): Register = {
--- a/src/main/scala/translating/ILtoIL.scala
+++ b/src/main/scala/translating/ILtoIL.scala
@@ -29,10 +29,7 @@ private class ILSerialiser extends ReadOnlyVisitor {
 
 
   override def visitExpr(node: Expr): Expr = {
-    program ++= "Expr("
     node.acceptVisit(this)
-    program ++= ")"
-    node
   }
 
   override def visitStatement(node: Statement): Statement = node.acceptVisit(this)
--- a/src/main/scala/translating/ILtoIL.scala
+++ b/src/main/scala/translating/ILtoIL.scala
@@ -2,7 +2,7 @@ package translating
 import ir._
 
 
-private class ILSerialiser extends Visitor {
+private class ILSerialiser extends ReadOnlyVisitor {
   var program: StringBuilder = StringBuilder()
 
   var indentLevel = 0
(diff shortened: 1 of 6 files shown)


src/main/scala/translating/ILtoIL.scala
```diff
@@ -0,0 +1,279 @@
...
+    node.mem = visitMemory(node.mem)
+    program ++= ", ["
+    node.index = visitExpr(node.index)
+    program ++= "])"
+    node
+  }
+
+  override def visitMemory(node: Memory): Memory = {
+    program ++= "Memory("
+    program ++= node.toString()
```
l-kent, line 245: This results in outputting `Memory(Memory(` since Memory's `toString` also includes `Memory(`

src/main/scala/translating/ILtoIL.scala
```diff
@@ -0,0 +1,279 @@
...
+  }
+
+  override def visitBinaryExpr(node: BinaryExpr): Expr = {
+    program ++= "BinaryExpr("
+    node.arg1 = visitExpr(node.arg1)
+    program ++= ", "
+    node.arg2 = visitExpr(node.arg2)
+    program ++= ")"
+    node
+  }
```
l-kent, line 221: You should include the operators too.

src/main/scala/translating/ILtoIL.scala
```diff
@@ -0,0 +1,279 @@
+package translating
+import ir._
+
+
+private class ILSerialiser extends Visitor {
```
l-kent, line 5: probably better to extend ReadOnlyVisitor since this isn't editing the IR?

src/main/scala/translating/ILtoIL.scala
```diff
@@ -0,0 +1,279 @@
...
+    program ++= ")"
+    node
+  }
+
+  override def visitSignExtend(node: SignExtend): Expr = {
+    program ++= "SignExtend("
+    node.body = visitExpr(node.body)
+    program ++= ")"
+    node
+  }
```
l-kent, line 205: You should include more than just the body for all of these.

src/main/scala/translating/ILtoIL.scala
```diff
@@ -0,0 +1,279 @@
...
+    for (i <- node.procedures.indices) {
+      val updatedProcedure = visitProcedure(node.procedures(i))
+      val targetProcedure = node.procedures(i)
+      if (targetProcedure == node.mainProcedure) {
+        node.mainProcedure = updatedProcedure
+      }
+      node.procedures(i) = updatedProcedure
+    }
+    node
+  }
```
l-kent, line 177: You don't need to copy in methods from the base class that haven't been changed, but using this version which is designed for mutating the IR is incorrect.

l-kent:
This is good to have, but you're missing a bunch of things that weren't in the Visitor (because they're just integers etc. and don't need to be visited themselves), and you should have used ReadOnlyVisitor as a base instead, as you don't need all those `node.body = visitExpr(node.body)` etc. which exist in the Visitor implementation to allow for mutating the IR.

src/main/scala/translating/ILtoIL.scala
ailrst: Thanks :facepalm:

ailrst:
Thanks for the review,

> you're missing a bunch of things that weren't in the Visitor (because they're just integers etc. and don't need to be visited themselves),

I'm not sure what's missing, e.g. Literal just forwards to `toString` which should give a meaningful output for the kind of trivial expression you mention I think?

l-kent:
Oh, I should have been more specific - things such as the Extract etc. parameters and the BinaryExpr operator that I highlighted.

ailrst:
> Looking at this again I think the main issue left is that the `Expr(`s are redundant - we don't need to put that around every subclass of `Expr`.

Yeah I wondered about that, I was thinking it might be better to include everything represented in the IL, e.g. its not automatically obvious that `Register` is an expression, but I decided to omit things like `Literal()` so leaving out Expr is probably better for readability.

Issue #63, opened 2023-09-12
Adding Memory Region Analysis results to IR


l-kent:
We want to add the results from the Memory Region Analysis to the intermediate representation. This means replacing all references to 'mem' in the IR with new variables representing discrete memory regions, derived from the memory region analysis. Each memory load and store should only have a single memory region associated with it - if the analysis says there are multiple regions, those regions need to be universally merged (in terms of the IR, anyway) to only have a single memory variable represent them. However, we want to keep stack regions distinct from other regions, so if there is a case where a memory access could potentially be either a stack access or a non-stack access, we should not merge the regions and instead create new blocks to represent both possibilities. I'm realising now we probably need to add non-deterministic branching to the IR for this case?

We want to keep track of which memory regions are stacks, so we probably need a flag for that in the Memory class.

ailrst:
- Add a switch to choose between #71's implementation and this

l-kent:
Ideally the MRA will be able to replace that entirely, but an option does make sense for the time being

yousifpatti:
Hey I started looking at this and I was able to replace memLoads and memStores with stack identifier using MRA.
I was abe to replace the memLoads with values if only a single one is given from VSA. 

However, I have also used OR to combine values in a case where a memLoad could lead to multiple values (is this correct?). I am not sure what you mean by _"universally merged to only have a single memory variable represent them"_ as I am not sure how you would merge the values without using BVOR for all of them.

Here is the PR for this: #74

l-kent:
What I mean is merging regions - if an access is ambiguous and could possibly be multiple regions according to the MRA & VSA, it's necessary to merge those regions into a single region for the purposes of the IR.

l-kent:
To be clearer, the idea is to replace the `mem` variable for all accesses (`MemoryLoad`s and `MemoryStore`s) with a new `Memory` variable that represents the memory region. If an access could point to multiple regions, then you need to merge all those possible regions together for the purposes of the IR and represent them with a single `Memory` variable wherever any of those regions would be mentioned.

For example, if there is a `MemoryStore` that could be to the regions `region1` or `region2`, then you need to merge `region1` and `region2` and represent them both with a new `Memory` variable `region1_region2` (the names I've chosen are arbitrary) that is used in that `MemoryStore`. If there is a separate `MemoryLoad` that is only to `region1`, then you need to use the `Memory` `region1_region2` for it as well.

[...]

PR #86, opened 2023-09-26
Improved Memory Identification

yousifpatti:
Targets MemoryLoads and Stores

Replaces mem based on type:
data section: the variable name (ie. x, y, arr ..)
stack section: the abstract stack name in that function (stack_1, stack_2, ...)

One step closer to achieving  #63


l-kent:
I'm not sure how complete this is but I will give some feedback on the current state.

We do not want to replace `MemoryLoad`s with constants as this breaks our entire security level analysis. Those constants seem to often be wrong too though. Some of the reasons why include seem to include:
- Values are incorrectly propagated past function call sites, even though the register's value is not maintained by the call
- In some cases loads are replaced by the address being accessed, not the value stored at the memory location accessed

We only want to replace references to `mem` and `stack`, but this means all references to `mem` and `stack`. Currently, the replacement is far from complete, and the left-hand side of the `MemoryAssign`s are not modified, which is incorrect. Many of the regions seem to be incorrect too. An example of one is that if you run the analysis on test/correct/function/clang, we get the following (with the Gammas removed):

```
R29 := memory_load64_le(x, R31)
R30 := memory_load64_le(stack_5, bvadd64(R31, 8bv64))
```
This seems incorrect to me. These two lines should be accessing the same regions (as they are accessing the same addresses) as the following two lines from the start of main():
```
stack := memory_store64_le(stack, #4, R29)
stack := memory_store64_le(stack, bvadd64(#4, 8bv64), R30)
```
It is definitely not accessing the region for the global variable `x`, and there is no other reference to `stack_5` generated.

I think modifying the IR in the middle of the memory region analysis is not going to be the best way to approach this. We really need a whole-program approach to modifying the IR, so it would make more sense to make modifying the IR be a separate step that uses the results of the MRA, after the MRA is complete and all the regions have been identified.

We will also need to modify the references to `mem` etc. created during the `IRToBoogie` translation.

PR #88, opened 2023-09-28
Basil 19 built in non returning functions

ziggyfish:


l-kent:
I don't think it really makes sense to put this at the level of the BAP to IR translation and loading the BAP representation from the parser? I'd prefer to keep the BAP AST to be a fairly direct representation of what we get from BAP, and keep transformations like this as separate passes over the IR. For instance, we may need to apply this sort of pass early on, but then also later again after resolving indirect calls?

This is also a shallow approach that doesn't seem like it will consistently deal with nested non-returning calls (where if f always calls g which always calls h which is non-returning, then f and g should also both also non-returning). We really want something more like the approach described in 'Binary Code is Not Easy' to handle those cases.

The result of the translation also seems like it's dependent on the order the procedure are iterated over, as it uses Procedure.nonReturning of other procedures that are called by the current procedure to set the returnTarget of DirectCalls, but sets Procedure.nonReturning for the current procedure. Parallelising that like you have seems like it would potentially be non-deterministic, which we don't want.

l-kent:
Oh, well you should probably mark this pull request as a draft then until it's ready then.

Where you've done it there isn't the IR stage, that's at the translation step from the BAP representation to the IR representation, and loading the BAP representation from the ADT parser. The IR stage is where the program is represented by the classes in the IR package. The pipeline is parser -> BAP representation -> IR representation -> Boogie representation -> final .bpl file. The CFG is created from the IR for the purposes of the analysis, and then the IR is modified based on the results of the analysis. It would be possible to do this by making use of the CFG or not, whichever works best.

ziggyfish:
I've updated the implementation to work similar the "Binary Code is Hard" article.

I've also added code to remove dead code after while (true) endless loops.

src/main/scala/analysis/NonReturningFunctions.scala
```diff
@@ -0,0 +1,139 @@
...
+                    blocksToRemove.enqueue(goTo.target.label)
+                  case _ =>
+                }
+              }
+            }
+          }
+          if (procedureBlock != null) {
+            procedure.blocks.remove(procedureBlock)
+            blocksDeleted = true
+          }
```
l-kent, line 133: since we have Option types in Scala there isn't any good reason to use null

src/main/scala/analysis/NonReturningFunctions.scala
```diff
@@ -0,0 +1,139 @@
...
+
+        val (_, idx) = mapBlocksToProcedure(goTo.target.label)
+
+        for (loopIndex <- idx.toInt to index.toInt) {
+          val blockAtLoopIndex = proc.blocks(loopIndex)
+          for (loopJump <- blockAtLoopIndex.jumps) {
+            loopJump match {
+              case loopGoTo: GoTo =>
+                if (!mapBlocksToProcedure.contains(loopGoTo.target.label) || mapBlocksToProcedure(loopGoTo.target.label)._2 >= index)
+                  return false
```
l-kent, line 36: mapBlocksToProcedure associates blocks with an index based on the order they are iterated over. Using this is not a safe way to identify the control flow between blocks. The only guarantee we have about the order of blocks is that the first block in a procedure is the entry point.

src/main/scala/ir/Program.scala
```diff
@@ -172,7 +177,8 @@ class Block(
     var label: String,
     var address: Option[Int],
     var statements: ArrayBuffer[Statement],
-    var jumps: ArrayBuffer[Jump]
+    var jumps: ArrayBuffer[Jump],
+    var countOfReturnStatements: Int
```
l-kent, line 181: Wouldn't it make sense for this be in the body of the class instead of the parameters?

src/main/scala/translating/BAPLoader.scala
```diff
@@ -124,13 +128,11 @@ object BAPLoader {
...
-      visitExp(ctx.cond),
-      returnTarget,
-      line,
-      insn
-    )
+    val function = visitQuoteString(ctx.callee.tid.name).stripPrefix("@")
+    if (knownNonReturningFunctions.contains(function))
+      BAPDirectCall(parseAllowed(function), visitExp(ctx.cond), None, line, insn)
+    else
+      BAPDirectCall(parseAllowed(function), visitExp(ctx.cond), returnTarget, line, insn)
```
l-kent, line 135: As I've said, I would really like to keep BAPLoader focused on representing what we get from BAP and directly translating that to the IR. Identifying non-returning calls like this from the function names should happen later on and modify the Program, not happen during the parsing & loading. We want to try to keep the steps separated.

src/main/scala/translating/BAPToIR.scala
```diff
@@ -30,7 +33,7 @@ class BAPToIR(var program: BAPProgram, mainAddress: Int) {
       for (p <- s.out) {
         out.append(p.toIR)
       }
-      val procedure = Procedure(s.name, Some(s.address), blocks, in, out)
+      val procedure = Procedure(s.name, Some(s.address), blocks, in, out, externalFunctionNames.contains(s.name))
```
l-kent, line 36: Since identifying the external functions isn't something we get from BAP, it should happen later on when the program is represented by the IR, not during the process of translating to the IR.

src/main/scala/analysis/NonReturningFunctions.scala
```diff
@@ -0,0 +1,139 @@
...
+            case call: IndirectCall =>
+              if (call.target.name == "R30") {
+                block.countOfReturnStatements += 1
+              }
+            case directCall: DirectCall =>
+              mapJumpsToBlocks.put(directCall.target.name, mapJumpsToBlocks.getOrElse(directCall.target.name, ArrayBuffer()).addOne((directCall, block)))
+            case goTo: GoTo =>
+              mapJumpsToBlocks.put(goTo.target.label, mapJumpsToBlocks.getOrElse(goTo.target.label, ArrayBuffer()).addOne((goTo, block)))
+              if (proc.blocks.length > index && isEndlessLoop(proc, goTo, index)) {
+                blocksToRemove.enqueue(proc.blocks(index + 1).label)
```
l-kent, line 71: This line throws an IndexOutOfBoundsException for the example complex.

l-kent (changes requested):
Do you have any examples used to test it identifying infinite loops? It would be good to commit those, to have as a reference.

Both examples basic_nonreturning_function_exit and basic_function_calling_nonreturning_function_exit have the same issue: in the procedure non_returning_function(), the inaccessible block directly following the non-returning call to exit() is not removed.

The names of those examples are very long too, it might be convenient to name them something a little shorter.

For the test indirect_call/gcc_O2, greet is incorrectly identified as non-returning.

For the test malloc_with_local3/gcc_O2, printCharValue is incorrectly identified as non-returning, and the only block removed is immediately following the call, instead of all inaccessible blocks as you would expect.

src/main/scala/analysis/NonReturningFunctions.scala
ziggyfish: Actually according to code and documentation, (see git repository), you can rely on the order of the blocks being the order in which they apear in the procedure. Meaning that the code may be in different locations, but BAP will reorder these blocks so they are in the same order as the orginal program.

The IR to boogie, also relies on this approache as well.

src/main/scala/translating/BAPLoader.scala
ziggyfish: It was more efficient to do it that way, but have moved it to nonReturningFunction now

src/main/scala/analysis/NonReturningFunctions.scala
l-kent: Where does anything say that you can rely on the order of the blocks in this way? If it says that, it's incorrect.

There is a difference between the order of the blocks and the control flow of the program. If we have a procedure with blocks A, B, C  in that order, the only guarantee is that block A is the first block in the procedure. Control flow between the blocks is determined entirely by jumps. It is not reliant on the order of blocks at all beyond the first block being the entry point.

src/main/scala/analysis/NonReturningFunctions.scala
ziggyfish: Fixed

src/main/scala/translating/BAPToIR.scala
ziggyfish: Removed, although its very inefficient to do this.

src/main/scala/ir/Program.scala
ziggyfish: Moved

src/main/scala/analysis/NonReturningFunctions.scala
ziggyfish: > Where does anything say that you can rely on the order of the blocks in this way? If it says that, it's incorrect.
> 
> There is a difference between the order of the blocks and the control flow of the program. If we have a procedure with blocks A, B, C in that order, the only guarantee is that block A is the first block in the procedure. Control flow between the blocks is determined entirely by GoTos. It is not reliant on the order of blocks at all beyond the first block being the entry point.

See https://github.com/BinaryAnalysisPlatform/bap/blob/master/plugins/print/print_main.ml line 207 and
https://github.com/BinaryAnalysisPlatform/bap/blob/master/lib/bap_types/bap_ir.ml see lines 1888, 1861 and 1853 

BAP itself relies on this fact during other areas of its source code as well (available on request).

> 
> IRToBoogie doesn't rely on the blocks being in any particular order either.

Lines 285, and 379 implicitly rely on the same order the instructions are found in the BAPToIR translation.

Also there isn't really any other way to remove code that can't be jumped to once a endless while loop is found. The fact that I have found an endless loop means there isn't any control flow to the next block, so even using the CFG will have the same issues, the only difference is the CFG will make it much more difficult (both in time and memory costs) to find endless loops.

Hope that helps.

l-kent: I don't see evidence in those examples of the order of the blocks implying anything about the control flow. BAP may order the blocks in a certain way, but we are not relying on such intricacies of BAP where possible, as there are intended to be inputs other than BAP at some point. 

IRToBoogie does not rely on the blocks being in any particular order inside a procedure there, beyond the initial block being first, because that is all that matters to Boogie. Control flow between blocks is entirely dependent on GoTos.

I don't know what you mean by 'using the CFG will have the same issues'? I'm not suggesting that CFG should be used to identify infinite loops. Removing unreachable code should happen in a separate method to identifying non-returning procedures and marking calls as non-returning, because we don't necessarily want to do both at the same time.

ziggyfish: > I don't see evidence in those examples of the order of the blocks implying anything about the control flow. BAP may order the blocks in a certain way, but we are not relying on such intricacies of BAP where possible, as there are intended to be inputs other than BAP at some point.
> 

I can assure you those are the lines that indiccate the order of the control flow (rest of the proof comes from many other files, and don't really have time to prove it entirely). Also its own implementation relies on this fact as well.

> 
> I don't know what you mean by 'using the CFG will have the same issues'? I'm not suggesting that CFG should be used to identify infinite loops. Removing unreachable code should happen in a separate method to identifying non-returning procedures and marking calls as non-returning, because we don't necessarily want to do both at the same time.

We do actually want it to be in the same process. The effect of removing code could actually render a function non-returning.

and example is basic_function_calling_nonreturning_function.c. The function_calling_non_returning_function has a return statement, which is only considered non-returning because the code return statement is dead code. Therefore each time code is removed, an evaluation is needed to determine if the function is still non-returning.

The issue with while loops is there is by definition, no way to identify where the next block is. This is because if there was a jump to the next block then there would be a control path that jumps, and thus making the function a returning function. Therefore this code needs to know the order of the blocks. Also this needs to be done at this level to remove code that could return but is not reachable.

The other way you could do this is to use the CFG, however I'm not sure this would even work either, as I can't simply just traverse the CFG. Also from what I understand while loops is something the CFG has problems with currently.

If your concerned about it, I can remove the detection of endless loops entirely, but this would mean that some functions that are non-returning will be marked as non-returning.

l-kent: It doesn't matter if BAP is implicitly guaranteeing anything about the control flow in relation to the order of blocks. We are not relying on such implicit guarantees from BAP in the IR, because that means we have to maintain said guarantee whenever we modify the IR, and there is also the intention to add support for lifters other than BAP.

In order to remove unreachable blocks, you could simply iterate over the blocks in the procedure and remove the ones that do not have jumps to them (except for the initial block). This could easily be separated from identifying non-returning procedures and called later in the program. 

Dealing with infinite loops is not very important from my perspective, since we do not currently fully support loops in any form for a variety of reasons.

> The issue with while loops is there is by definition, no way to identify where the next block is. This is because if there was a jump to the next block then there would be a control path that jumps, and thus making the function a returning function. Therefore this code needs to know the order of the blocks.

The order blocks are contained within the procedure is irrelevant to reachability except for the first block, which is always reachable. It is possible to identify unreachable blocks without this idea of a 'next block'.

> We do actually want it to be in the same process. The effect of removing code could actually render a function non-returning.
> 
> and example is basic_function_calling_nonreturning_function.c. The function_calling_non_returning_function has a return statement, which is only considered non-returning because the code return statement is dead code. Therefore each time code is removed, an evaluation is needed to determine if the function is still non-returning.

It doesn't seem like this happens though, it seems like it stops trying to detect non-returning calls as soon as blocks have been deleted? There is also a difference between the block being unreachable and being removed. It would be possible to determine if the returns are now unreachable after setting returnTarget to None without removing the unreachable blocks completely.

Marking calls as non-returning is fine to happen when you identify non-returning procedures, but I want to keep removing the unreachable blocks separate and have it happen at the very end of the program, in case later analysis finds that the blocks are in fact reachable in some other way, for instance. I can anticipate needing to consider the order these things happen in and do not want them to be tied so directly together.

ziggyfish: > It doesn't matter if BAP is implicitly guaranteeing anything about the control flow in relation to the order of blocks. We are not relying on such implicit guarantees from BAP in the IR, because that means we have to maintain said guarantee whenever we modify the IR, and there is also the intention to add support for lifters other than BAP.
> 
> In order to remove unreachable blocks, you could simply iterate over the blocks in the procedure and remove the ones that do not have jumps to them (except for the initial block). This could easily be separated from identifying non-returning procedures and called later in the program.
> 
> Dealing with infinite loops is not very important from my perspective, since we do not currently fully support loops in any form for a variety of reasons.
> 

I'll remove the code that looks at infinite loops, athough there are issues here that need to be addressed, if we are truely going to identify all non-returning functions. I'll leave it to someone else to implement.

> 
> The order blocks are contained within the procedure is irrelevant to reachability except for the first block, which is always reachable. It is possible to identify unreachable blocks without this idea of a 'next block'.

Lets do that in another ticket.


> 
> > We do actually want it to be in the same process. The effect of removing code could actually render a function non-returning.
> > and example is basic_function_calling_nonreturning_function.c. The function_calling_non_returning_function has a return statement, which is only considered non-returning because the code return statement is dead code. Therefore each time code is removed, an evaluation is needed to determine if the function is still non-returning.
> 
> It doesn't seem like this happens though, it seems like it stops trying to detect non-returning calls as soon as blocks have been deleted? There is also a difference between the block being unreachable and being removed. It would be possible to determine if the returns are now unreachable after setting returnTarget to None without removing the unreachable blocks completely.


line 88 in NonReturningFunctions.scala loops until there isn't any deleted blocks.

Lets talk about whether we should deleted the blocks unreachable on Tuesday.

> 
> Marking calls as non-returning is fine to happen when you identify non-returning procedures, but I want to keep removing the unreachable blocks separate and have it happen at the very end of the program, in case later analysis finds that the blocks are in fact reachable in some other way, for instance. I can anticipate needing to consider the order these things happen in and do not want them to be tied so directly together.

The way I handle reachability is not to always delete the block but to determine whether the code can infact be jumped too by other blocks. If it can, then it isn't deleted (i.e that is where it stops).

Seperated the deleteing from the marking of deleting.

Any further changes should be in another ticket.

ziggyfish:
> 
> For the test indirect_call/gcc_O2, greet is incorrectly identified as non-returning.
> 
> For the test malloc_with_local3/gcc_O2, printCharValue is incorrectly identified as non-returning, and the only block removed is immediately following the call, instead of all inaccessible blocks as you would expect.

These were not working prior to my additions.

l-kent:
Looking at those cases more closely, this seems to be an issue with tail calls.

ailrst:
Bap also flags some functions with the `noreturn` attribute which isn't currently loaded from the ADT into the IR, it may be worth using that also.

l-kent:
That's a good find and we probably should incorporate that, but I'm not sure how useful it is to rely on, since BAP doesn't seem to add it for __stack_chk_fail which is an obvious case of an external non-returning call.

l-kent:
In Binary Code Is Not Easy, their algorithm for detecting non-returning calls (in Figure 8) says that if a function has a reachable return block or tail calls a returning function, then it is a returning function. The problem with indirect_call/gcc_O2 and malloc_with_local3/gcc_O2 is that tail calls aren't taken into account at all - is this something that will be addressed for #101? 

This still isn't separating the deletion of unreachable blocks from the identification of the non-returning calls. You have just marked which blocks to delete at the same time as identifying non-returning calls, and then created a separate call to remove them. I'm sorry if this was unclear, but the motivation for not deleting the now-unreachable blocks until later was that we might require them later, so it wouldn't be safe to assume they are still unreachable at this later call.

This also still isn't removing all unreachable blocks either, as shown in  your basic_nonreturning_function_exit and basic_function_calling_nonreturning_function_exit examples.

ziggyfish:
> In Binary Code Is Not Easy, their algorithm for detecting non-returning calls (in Figure 8) says that if a function has a reachable return block or tail calls a returning function, then it is a returning function. The problem with indirect_call/gcc_O2 and malloc_with_local3/gcc_O2 is that tail calls aren't taken into account at all - is this something that will be addressed for #101?
> 

This is correct, tail calls, are not taken into account just yet, and will actually need to be a seperate process. This is becaue BAP will mark calls as non-returning when it sees a jump to a function rather than a BL instruction. indirect_call/gcc_O2 and malloc_with_local3/gcc_O2 were already marked as non-returning, because of BAP and how it handles tail calls.

> This still isn't separating the deletion of unreachable blocks from the identification of the non-returning calls. You have just marked which blocks to delete at the same time as identifying non-returning calls, and then created a separate call to remove them. I'm sorry if this was unclear, but the motivation for not deleting the now-unreachable blocks until later was that we might require them later, so it wouldn't be safe to assume they are still unreachable at this later call.
> 

I'm not sure you actually understand what is going on here, nor why blocks need to be marked as deleted. I'm also at a loss as to what exactly you want. Please explain!!

The purpose of deleting the blocks is to eliminate blocks that happen after a return call due to the complier producing stack frame instructions that are never executed. Hence why I mark them as deleted, then delete them as you requested in later parts of the process.

> This also still isn't removing all unreachable blocks either, as shown in your basic_nonreturning_function_exit and basic_function_calling_nonreturning_function_exit examples.

Can you provide examples of this?

ziggyfish:
> Just a note, Bap also flags some calls with the `noreturn` attribute which isn't currently loaded from the ADT into the IR, it may be worth using that also.

This is added into the IR using None as the returnTarget for those calls.

ziggyfish:
> In Binary Code Is Not Easy, their algorithm for detecting non-returning calls (in Figure 8) says that if a function has a reachable return block or tail calls a returning function, then it is a returning function.

This is correct, however there is a diferentiation between a tail call is returning, and a tailing call that is non-returning. Meaning, that tail calls that are non-returning, are not counted as a returning block. You will also notice line 5 of FIgure 8, says continue finding non-returning functions until you have found no more. This is because, the change in the function's non-returning state changes the state of ofther function calling it.

The other point here is the use "reachable", if there is a return block, that is returning after a non-returning block, then return block is not reachable. Therefore, as part of the analysis, we need to check these for these occurances. Hence why soft deleting is required.

We discussed this at the beginning of the task, and it was decided to delete these blocks, I'm not sure why now we want to have the status of these blocks simultainously in two states, deleted and not there. Maybe you haven't been clear as to what you wanted. So please clarify as I'm at a loss of what to do next here.

ziggyfish:
I'm not sure what needs to be changed here, there isn't any real understanding of the reviewers, nor any idea of how it should be done. So I'll leave this with liam.

Issue #87, opened 2023-09-28
basic_arrays_write incorrect array spec


ailrst:
The policy spec no longer allows specifying the policy for single array cells so some of the spec files in the examples need to be updated.

PR #89, opened 2023-09-29
Fix outdated example specs

l-kent:
#87

These specs had been updated in the tests folder but not the examples folder when the implementation changed. A further cleanup there will need to happen at some point, but this will prevent any confusion with this example for now.


ailrst:
Looks good, tests pass on my machine

Issue #81, opened 2023-09-21
Axioms to maintain constant sections of memory


l-kent:
Our current approach to modelling initialised memory is inadequate, and this causes issues for some tests (specifically some of the `gcc_pic` tests). Currently, we add the initial values for the  `.data`, `.rodata` and `.got` sections of the binary (because those have proven relevant so far) to the precondition of the main procedure, which allows pointers stored in the global offset table to be loaded, for instance. This approach works for accesses within main, but these preconditions are not added to other procedures (so obviously do not work for those procedures), and calls to rely() mean that these parts of memory are havoced away if there are rely conditions specified, as we do not adequately preserve all of memory. The issue with rely calls is something the memory region analysis (#63) should help with, by splitting the memory into discrete regions, but may not entirely solve. 

We likely want to perform some analysis to detect initialised memory locations that are read but not written to, and add Boogie axioms so that these memory accesses are constant across all procedures. We could just assume for the time being that the global offset table pointers detailed in the '.rela,dyn' section section are going to be constant (as these are the most important constants in the initialised memory), without actually checking, but a more sophisticated approach could be helpful more broadly in the future.

l-kent:
Out of the sections we use, only `.rodata` is guaranteed to be read-only, but we still want some way to determine if any parts of the others are constant in practice.

PR #92, opened 2023-10-09
Read-only memory sections maintained throughout program

l-kent:
Addresses #81. Since Boogie does not support mutable global variable references in axioms, the solution was to place the relevant predicates in the pre- & post-conditions of every procedure, since we are safely assuming they will be maintained by the program. This allows more tests to pass than before, mostly `/gcc_pic` tests.

The global offset table was handled by just adding in the relevant offset/address pairs listed in the `.rela.dyn` table, which is sufficient for our current purposes.

The following postcondition for the rely procedure in Boogie was removed if the rely is empty, as it is redundant and affects both performance and verification success:
`ensures (forall i: bv64 :: (((mem[i] == old(mem[i])) ==> (Gamma_mem[i] == old(Gamma_mem[i])))))`

#77 was also addressed by making the tests use the `/useArrayAxioms` flag which significantly improves performance.


ailrst:
Can we add a doc comment to explain specifically which elf section is included and why, i.e. the outcome of the email discussion about which sections are modifiable. Correct me if I'm wrong but might we want to do future work to map in more memory from the elf header to include the statically initialized but mutable values?

l-kent:
We do already have some of the statically initialised but mutable sections, but I'll add comments to document it.

ailrst (approved):
Looks good, thanks

Issue #91, opened 2023-09-29
Direct specification parsing errors


ailrst:
For the demo its nice to directly insert relys, guarantees, and modifies into the boogie for procedures. 

There's a bug with the `DIRECT functions: ` parsing where e.g. `bvadd64` is supposed to parse as 3 separate tokens but instead it's treated as a single token. It might be easy to just have a flag to output _all_ the helper functions. 

It would also be good to be able to insert values into the modifies, e.g. for defining summaries for library functions like memset.

PR #93, opened 2023-10-09
Specification Improvements

l-kent:
This covers the specification improvements requested in #91.

The `DIRECT functions:` field now works properly - as an example, the following will parse correctly:
`DIRECT functions: gamma_load32, gamma_load8, bvule64, bvult64, bvuge64, bvsub64, memory_load128_le, zero_extend64_32, sign_extend64_32`

It is not really feasible to create instantiate all possible helper functions due to the very large possible numbers of sizes for some operations, and this approach is ok for now as the DIRECT functions is just a temporary measure until we have a more sophisticated specifications system.

It is now possible to manually specify variables that should be in a procedure's modifies clause in a specification. You do not need to specify `Gamma_` variables separately. For example, the following will add the global variables mem and R0 to the modifies clause of get_secret() and correctly propagate this through to any procedures that call get_secret():
```
Subroutine: get_secret
Modifies: mem, R0
```
You do not need to specify the entire modifies clause, just variables that you want to add to it.

Boogie global variables that were modified by a stubbed out PLT procedure are also now properly added to the modifies clause of those procedures (in practice this is usually just R16 and R17).


ailrst:
This looks good we tested it pretty thoroughly this week

ailrst (approved):

PR #94, opened 2023-10-09
Tests cleanup

l-kent:
- Removes some impossible tests
- Adds specifications required for some tests to ever verify
- Makes unresolved indirect calls cause verification failure (an alternate approach would be to make them havoc all global variables?)
- Fixes bug where global variables referenced in pre/postconditions were not guaranteed to be defined


ailrst:
Some of the removed tests appear to fail since the folders and expected output are still present but not the lifted examples, is this the intention?

```
[info] - iflocal/gcc_O2 *** FAILED ***
[info]   java.nio.file.NoSuchFileException: ./src/test/incorrect/iflocal/gcc_O2/iflocal.adt
```

l-kent:
That's just a result of git not deleting the folders when you switch branches if you have any local untracked files in them. It's awkward but I'm not sure if there's a solution to that. I guess maybe I could make it so that it checks if a folder actually has an example in it but it's more that sometimes, git can't do everything for you.

ailrst:
Yeah okay this is just a git problem

ailrst:
Looks good

PR #99, opened 2023-10-13
add example with quantified specs

ailrst:
This adds an example we tried to put together for the demo, with some notes on what was necessary to get the quantification specs to work. 

This example isn't working well enough to become a test case or anything since the tool currently mis-identifies some stack regions as mem, but I want to keep the notes around.


l-kent:
Please add the .adt, .bir and .relf files for this example. 

I'll merge this, but the specification won't work until you merge in #93 which this relies on. 

Which stack regions are misidentified at present? Please submit an issue for that.

ailrst:
```
  l00000460:
    R8, Gamma_R8 := memory_load64_le(mem, bvadd64(R31, 32bv64)), (gamma_load64(Gamma_mem, bvadd64(R31, 32bv64)) || L(mem, bvadd64(R31, 32bv64)));
    R9, Gamma_R9 := memory_load64_le(mem, bvadd64(R8, 120bv64)), (gamma_load64(Gamma_mem, bvadd64(R8, 120bv64)) || L(mem, bvadd64(R8, 120bv64)));
    R9, Gamma_R9 := bvadd64(R9, 2bv64), Gamma_R9;
```

Should these loads be from stack since theyre offsets from SP?

l-kent:
Minor note but it's .bir not .bil. BIL is a different output from BAP.

That does look like it should be a load from stack, but please make an issue for it.

ailrst:
> Minor note but it's .bir not .bil. BIL is a different output from BAP.

Ah, sorry I remembered your comment on the other pr but couldn't remember which one was correct :(

PR #102, opened 2023-10-13
Stack Identification Bug Fix

### LINKED ISSUE #100: Incorrect stack identification
The tool can output stores to mem at addresses that are explicitly offsets from the stack pointer such as here 
https://github.com/UQ-PAC/bil-to-boogie-translator/pull/99

This may be improved by implementing stack identification based on the memory region analysis, or fixed in the prototype implementation of stack identification first.

l-kent:
Fixes #100

There were two problems with my previous implementation: not traversing to the returnTarget of calls, and counting the rhs of an assignment as a stack value if it contained a stack reference inside a load index or store.

### DIFF
--- a/src/main/scala/ir/Program.scala
+++ b/src/main/scala/ir/Program.scala
@@ -140,7 +140,10 @@ class Procedure(
             }
 
             // update stack references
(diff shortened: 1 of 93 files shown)


ailrst:
Looks good, thanks

Issue #85, opened 2023-09-26
option to output build version


ailrst:
It would be nice (e.g. for godbolt) to have basil be able to output its own version, ideally at this stage this could just be the commit sha it was built from.

l-kent:
Yeah, I don't think we'd need more than just the commit SHA for now

ziggyfish:
This is generally done through git tags, generally when you create a new tag for a commit. Most build systems (including this one), will give the git version or tag during the build process through environment variables. These environment variables can then be picked up by sbt ( `sys.env.get("NAME_OF_ENV_VAR")` ) and compiled into the package, accessable by ( `version = getClass.getPackage.getImplementationVersion` ) .

ailrst:
The plan is to add an sbt action to write the current git tag to a version file, and have that set to the version in scala that is output from the program. It is possibly also useful to have a 'test build' sbt action which derives a new version from the current git commit.

PR #103, opened 2023-10-13
Add version output cli flag derived from git

ailrst:
For  #85

A bit silly that this adds two new sbt plugins but I couldn't figure out a nice way to pass the version string from SBT to Scala, Brendan's suggestion didn't seem to work. 

This git plugin is nice since it appends the git hash to the tagged version and adds a flag if its been modified since checkout. E.g. `wptool-boogie 0.1.1-alpha-65-8914893e-SNAPSHOT`

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,7 +1,7 @@
 import scala.io.Source
 
+import scala.sys.process.*
 ThisBuild / scalaVersion := "3.1.0"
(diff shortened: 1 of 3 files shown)


ailrst:
We don't want this to break builds outside of a git repository and this is low priority so it is on hold for now https://github.com/UQ-PAC/bil-to-boogie-translator/issues/85

Issue #36, opened 2023-06-22
Check that array indices are low?


utting:
Kirsten wrote, re: arrays and array accesses.

For now we do not check array index accesses explicitly as it requires more effort to associate accesses with arrays.

However, since every memory access is effectively an array access of sorts (mem is an array) we should check that all offsets that are used in the address calculations are low to avoid potential leakage.
Mostly these are constants anyway but where they stem from indirect addresses it should be ensured that there is no flow of information.

Is this check already added in the generated VCs? 
It wasn't formalised in the wpif paper.
However, Liam encoded it in his Wemelt tool, based on Nick's earlier formalisation.

Does this make sense, and if so could you please add this in?

l-kent:
This has been implemented in ce5855b525283cccda1bcbd9e3cb74ebc0922bac but is not suitable for merging into main until further work has been done to fix stack identification and issues with initialising the security values of memory have been addressed.

Issue #82, opened 2023-09-22
Security Levels for Initialised Memory Sections


l-kent:
Currently the security level of the initialised memory sections put in the precondition of main() are not specified. We should assume them to be low for now, and either add axioms for them or add this information to the precondition of main(). We should also make care to exclude any global variables in the `.data` section from this assumption, as the security levels of those should be specified by the user.

Whether to use axioms or put this information in the precondition of main relates to #81 - it would only really be safe to add axioms for them if they are never written to by the program. 

The axioms would look like `axiom gamma_load8(mem, _address_);` for each byte of the initialised memory that we want to specify as low.

Future work would include a way for the user to manually specify security levels for other parts of the initial memory, but that isn't relevant for now and should wait until a wider rework of the specifications has taken place.

l-kent:
Since Boogie doesn't support axioms referencing mutable global variables (I forgot), it would make sense to instead insert these as free requires/ensures clauses for every procedure, similar to the approach taken in #92.

The most relevant examples are arrays_simple/gcc and arrays_simple/gcc_pic, but these pose an additional problem - the address of the `stack_chk_guard` symbol isn't currently initialised at all (because it can't be), so we probably want to additionally make the offsets in `.rela.dyn` of type `R_AARCH64_GLOB_DAT` return values that are low, as well as the memory sections we already initialise from #92.

PR #108, opened 2023-10-16
Check indices of memory accesses are low

l-kent:
Implements #36 

Adds assertions to check that addresses in MemoryLoads have a low security value.

This should not be merged until after #92.


l-kent:
Actually we won't want to do this until #82 is implemented.

PR #109, opened 2023-10-16
add .adt pretty-printer in format_adt.py

katrinafyi:
This makes it easier to inspect .adt files manually without the burden of fully parsing them.

The formatter works by exploiting the almost-Python syntax of the ADT. It tweaks it into valid Python, then parses and formats it using Python's ast.literal_eval and pprint.pprint modules. After this, it reverses the transformation to produce something very much like BAP.

Is this something you would like to have?

Example output from the formatted secret_write.adt:
```python
 Program(
  Tid(1_696, "%000006a0"),
  Attrs([]),
  Subs(
   [Sub(
     Tid(1_646, "@__cxa_finalize"),
     Attrs(
      [Attr("c.proto", "signed (*)(void)"),
       Attr("address", "0x5D0"),
       Attr("stub", "()")]),
     "__cxa_finalize",
     Args(
      [Arg(
        Tid(1_697, "%000006a1"),
        Attrs(
         [Attr("c.layout", "[signed : 32]"),
          Attr("c.data", "Top:u32"),
          Attr("c.type", "signed")]),
        Var("__cxa_finalize_result", Imm(32)),
        LOW(32, Var("R0", Imm(64))),
        ("Out"))]),
     Blks(
      [Blk(
        Tid(1_018, "@__cxa_finalize"),
        Attrs([Attr("address", "0x5D0")]),
        Phis([]),
        Defs(
         [Def(
           Tid(1_266, "%000004f2"),
           Attrs(
            [Attr("address", "0x5D0"),
             Attr("insn", "adrp x16, #65536")]),
           Var("R16", Imm(64)),
           Int(65536, 64)),
  ...
```

### DIFF
--- a/scripts/format_adt.py
+++ b/scripts/format_adt.py
@@ -13,9 +13,11 @@ which only supports parsing of a literal Python expression.
 import re
 import ast
 import sys
+import time
 import pprint
-import typing
+import logging
 import argparse
+import contextlib
 
 # grep -E "'([A-Z][a-zA-Z]+)'" src/main/antlr4/BAP_ADT.g4 --only-matching --no-filename | sort | uniq | xargs printf "'%s', " | fold -w80 -s
 heads = [
@@ -29,6 +31,9 @@ heads = [
   'Var', 'XOR',
 ]
 heads_joined = '|'.join(heads)
+
+log = logging.getLogger()
+
 
 def preprocess(data: str) -> str:
   """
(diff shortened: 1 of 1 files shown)


katrinafyi:
There is a big with zero-argument terms, `("Out")` in the example should be `Out()`.

katrinafyi:
With larger files, it is slow and heavy in both the parsing and printing phases. For cntlm.adt, it takes 100 seconds and up to 8GB memory on my computer (pypy reduces time to 40 seconds but the memory is the same).

l-kent:
I don't have an awful lot of personal use for this - I find the .adts are manageable enough with heavy use of CTRL+F, but I'm happy to merge this in, sure.

PR #114, opened 2023-10-17
Adding lifted lattice solver

yousifpatti:
Still under dev


l-kent:
#86 is already a pull request for this branch.

l-kent:
Is this ready to look at merging now?

I still don't really understand what the point of the LiftedLattice is for the analyses we have though. I understand that it is intended to show reachability, but why does this matter for the analyses we have, especially the MRA where we don't even need statement-level results?

yousifpatti:
> Is this ready to look at merging now?
> 
> I still don't really understand what the point of the LiftedLattice is for the analyses we have though. I understand that it is intended to show reachability, but why does this matter for the analyses we have, especially the MRA where we don't even need statement-level results?

MRA will eventually have to be context-sensitive as it will not scale with a function cloning (inlining) approach. As the contexts may be large, we need to infer only regions for contexts that may be feasible (reachable).

l-kent:
The MRA as it currently exists is intraprocedural (so reachability is irrelevant), and in effect just gives a set of memory regions per-procedure. It does not need to provide statement-level results at all with how it currently works. An interprocedural version that actually required context-sensitivity would be such a significant overhaul that it doesn't really make sense to add the LiftedLattice to the current version in preparation for that.

yousifpatti:
That's true. We would need an interprocedural version to support contexts. However, VSA currently is interprocedural so it will require the LiftedLattice as it relies on statement-level results (later could have context-sensitivity as well). Adding LiftedLattice to MRA will not cause any issues with the current development process even if it is not directly beneficial until the interprocedural overhaul.

yousifpatti:
It should be ready for merge now

l-kent:
But there also isn't any reason to actually add it now? All the MRA does in-effect at present is return a set of all the regions in a procedure. It doesn't even really need to be per-procedure, it could be one set for the whole program and nothing would really change. Even if the VSA and constant propagation are eventually properly made interprocedural, that's not really going to be relevant to the MRA as it currently exists.

yousifpatti:
Do you suggest removing the lifted lattice from MRA and merging the changes to VSA, or close the PR and re-open it when proper interprocedural VSA and constant prop have been established?

l-kent:
I will just merge it in for now.

PR #115, opened 2023-10-18
Propagate modifies clause through boogie program

ailrst:
- Fixes https://github.com/UQ-PAC/bil-to-boogie-translator/issues/111 by pushing all the values in the modifies clause up the call tree of the boogie program, right at the end of generation instead of always inserting an extra mem for the rely().
- Also renames "boogie.ProcedureCall" to "boogie.BProcedureCall".

This implementation is fairly inefficient, @l-kent you referred to a broader refactor to do something like this but I'm not sure what you had in mind for that.

### DIFF
--- a/examples/simplified_http_parse_basic/example.spec
+++ b/examples/simplified_http_parse_basic/example.spec
@@ -8,11 +8,11 @@ DIRECT functions: gamma_load64, gamma_load8, memory_load8_le, bvult64, bvule64, 
 
   
 Subroutine: #free
-  Requires DIRECT: "gamma_load8(Gamma_mem, bvadd64(R0, 0bv64)) == true";
-  Requires DIRECT: "gamma_load8(Gamma_mem, bvadd64(R0, 1bv64)) == true";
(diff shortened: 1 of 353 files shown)


l-kent:
The broader refactor I was thinking of would be moving all the insertion of relies, Gammas, etc. as a stage of the IR transformation rather than happening as part of the Boogie translation process at the end. This would allow various analyses to directly relate to those if necessary, so we wouldn't need to do two different fixed point propagation of modifies clauses like is happening here.

This will work for now though, sure.

l-kent (approved):
I've just made it use a Set which should make it marginally more efficient, and made the final modifies output be fully sorted again to ensure the future output is more consistent.

ailrst:
> The broader refactor I was thinking of would be moving all the insertion of relies, Gammas, etc. to be a stage of the IR transformation rather than happening as part of the Boogie translation process at the end. This would allow various analyses to directly relate to those if necessary, so we wouldn't need to do two different fixed point propagation of modifies clauses like is happening here.
> 
> This will work for now though, sure.

Yeah that makers sense that would be a good approach

PR #121, opened 2023-10-24
Performance tests

ailrst:
Change summary:

- Add timing to SystemTests
- Differentiate timeout from verification failure
- Have SystemTests write out two CSV files summarising the test results and times
- Pass CLI configuration through the program in runutils and the boogie translation
    - Add CLI flag `--boogie-use-lambda-stores` to generate the lambda definition for >8bit stores, rather than the nested MemOp definition, this gives reasonable performance without ArrayAxioms. 
- Add inline flag to BFunction

Breaking changes:
- boogie timeout is now always a test failure 

With defaults MemOps and ArrayAxioms:

```
[info] - summary
[info]   + Test summary: 305 verified, 81 failed to verify including 0 timeouts. 
[info]   + Average time to verify: 599 
[info]   + Average time to counterexample: 701
```

With lambdas and array theories:

```
[info] - summary
[info]   + Test summary: 305 verified, 81 failed to verify including 2 timeouts. 
[info]   + Average time to verify: 665 
[info]   + Average time to counterexample: 1851
```


l-kent:
What's the point of storing the expected IL output too? It isn't used for anything here and seems completely redundant since it is just going to be the same information as the final output.

I also think that something like that `--boogie-use-lambda-stores` flag should probably be in its own pull request so we can really assess if it's a better approach, but I guess you've added it as a flag so you can experiment with it for now.

ailrst:
> I also think that something like that --boogie-use-lambda-stores flag should probably be in its own pull request so we can really assess if it's a better approach, but I guess you've added it as a flag so you can experiment with it for now.

Its necessary for some examples to verify and just gives worse performance for others, I think we want to keep both around at this stage. But yes it could have been its own PR.

l-kent:
Can you explain about storing the IL output?

ailrst:
> Can you explain about storing the IL output?

Yeah that's not particularly useful at this point, my thinking was its so when we have more in-place transformation of the IR we have history in case of regressions.

l-kent:
I don't see any point in including it right now, it's just unnecessary bloat and any regressions will be visible in the Boogie output.

src/test/scala/SystemTests.scala
```diff
@@ -1,19 +1,33 @@
...
 
+  case class TestResult(passed: Boolean, verified: Boolean, shouldVerify: Boolean, hasExpected: Boolean, timedOut: Boolean, matchesExpected: Boolean, translateTime: Long, verifyTime: Long) {
+    val toCsv = s"$passed,$verified,$shouldVerify,$hasExpected,$timedOut,$matchesExpected,$translateTime,$verifyTime"
+  }
+
+  object TestResult {
+    val csvHeader = "passed,verified,shouldVerify,hasExpected,timedOut,matchesExpected,translateTime,verifyTime"
+  }
+
+  val testResults: mutable.Map[String, TestResult] = mutable.HashMap()
```
l-kent, line 29: Is there a reason for using a Map here, given that it means the output isn't ordered, which is much less convenient? It doesn't seem like we care about the constant lookup time?

src/test/scala/SystemTests.scala
```diff
@@ -35,37 +49,74 @@ class SystemTests extends AnyFunSuite {
...
     val ADTPath = variationPath + ".adt"
     val RELFPath = variationPath + ".relf"
     Logger.info(outPath)
+    val timer = PerformanceTimer(s"test $name/$variation")
     if (File(specPath).exists) {
-      Main.main(Array("--adt", ADTPath, "--relf", RELFPath, "--spec", specPath, "--output", outPath))
+      Main.main(Array("--adt", ADTPath, "--relf", RELFPath, "--spec", specPath, "--output", outPath, "--dump-il", ilPath))
     } else {
-      Main.main(Array("--adt", ADTPath, "--relf", RELFPath, "--output", outPath))
+      Main.main(Array("--adt", ADTPath, "--relf", RELFPath, "--output", outPath, "--dump-il", ilPath))
```
l-kent, line 87: Not really useful right now, as mentioned

build.sbt
```diff
@@ -45,21 +45,26 @@ updateExpected := {
...
           val verified = result.strip().equals("Boogie program verifier finished with 0 errors")
-          if (verified == shouldVerify && outPath.exists()) {
-            if (!expectedPath.exists() || !compareFiles(outPath, expectedPath)) {
+          if (verified == shouldVerify) {
+            if (outPath.exists() && !(expectedPath.exists() && filesContentEqual(outPath, expectedPath))) {
               IO.copyFile(outPath, expectedPath)
             }
+            if (ILOutPath.exists() && !(ILExpectedPath.exists() && filesContentEqual(ILExpectedPath, ILOutPath))) {
+              IO.copyFile(ILOutPath, ILExpectedPath)
+            }
```
l-kent, line 60: Also unnecessary right now

l-kent (changes requested):

src/test/scala/SystemTests.scala
ailrst: Uses an ArrayBuffer now

l-kent (approved):

src/test/scala/SystemTests.scala
```diff
@@ -53,20 +53,20 @@ class SystemTests extends AnyFunSuite {
...
-    info(s"Test summary: $numVerified verified, $numNotVerified failed to verify including $numTimeout timeouts.")
+    info(s"Test summary: $numSuccess succeeded, $numFail failed (including $numTimeout timeouts).")
     if (verifying.nonEmpty)
       info(s"Average time to verify: ${verifying.sum / verifying.size}")
     if (counterExamples.nonEmpty)
       info(s"Average time to counterexample: ${counterExamples.sum/ counterExamples.size}")
 
     val summaryHeader = "verifiedCount,counterexampleCount,timeoutCount,verifyTotalTime,counterexampleTotalTime"
-    val summaryRow = s"$numVerified,${counterExamples.size},$numTimeout,${verifying.sum},${counterExamples.sum}"
+    val summaryRow = s"$numSuccess,${counterExamples.size},$numTimeout,${verifying.sum},${counterExamples.sum}"
```
ailrst, line 69: I wanted verifyTotalTime/verifyCount =  mean verify time. scalaTest already tells you how many tests passed.

ailrst: I'll just make the summary include both things.

Issue #125, opened 2023-10-25
Allow overriding auto-generated functions and globals


ailrst:
Currently the boogie translation finds all variables named in the function specifications, and all the boogie functions called anywhere in the program, and lazily inserts definitions for these. We want to be able to replace these with different definitions by providing an additional specification to basil, without modifying the original spec or generated boogie directly. The goal is to facilitate testing the scalability and soundness/precision implications of different implementations of these functions, as well as potentially modelling more shadow state as in  https://github.com/UQ-PAC/bil-to-boogie-translator/blob/main/examples/simplified_http_parse_basic/malloc.md. 

This requires two things

1. Glue in additional boogie definitions 
2. Prevent basil from inserting conflicting definitions

This could be done by having a flag which turns off all generated functions and glues in a second file to the generated boogie, or something more precise like having the spec support inline boogie and explicit exclusion for specific definitions.

ailrst:
I just noticed boogie has an `{:extern}` attribute which marks the declaration as possibly-overridden. 

```
    {:extern}
      If two top-level declarations introduce the same name (for example, two
      constants with the same name or two procedures with the same name), then
      Boogie usually produces an error message.  However, if at least one of
      the declarations is declared with :extern, one of the declarations is
      ignored.  If both declarations are :extern, Boogie arbitrarily chooses
      one of them to keep; otherwise, Boogie ignore the :extern declaration
      and keeps the other.
```

l-kent:
Yeah, that's easier than removing them. Good find

PR #128, opened 2023-10-25
Make all generated Boogie declarations extern

l-kent:
Implements #125

This will allow a separate Boogie file to be used to override {:extern} declarations, which will aid in experimentation and prototyping.


ailrst:
I've generalized adding attributes to boogie statements and included this change in https://github.com/UQ-PAC/bil-to-boogie-translator/pull/131 as well

PR #130, opened 2023-10-27
add libclang funcdef to json tool

ailrst:
Adding this so it doesn't get lost. Its more experimental, we could possibly store it in a separate repository.


l-kent:
Yeah I think this should be put in a separate repo, it isn't really directly related at all.

PR #131, opened 2023-10-27
Add extern attribute and captureState attribute to pass binary locations to boogie for better debugging

ailrst:
This PR is to pass binary instruction addresses all the way through so that boogie's `/mv` flag can split the counterexample into states for us for https://github.com/UQ-PAC/bil-to-boogie-translator/issues/127. 

To do this it adds attributes and assumes to the Boogie AST. 

Changes

- Add `BAssume` to the boogie AST. 
- Add attributes to boogie commands and definitions
- Add extern attribute to generated definitions
- Add address to store ops in the BAP AST and load them from the adt file
- Add address to store ops in IL 
- Add `assert {:captureState "addr:0xxxx"} true` to the end of store ops in generated boogie, and to the beginning of basic blocks
- Changes IL statements to case classes; I feel like its likely there was a reason they weren't already case classes, please let me know if its a problem and I'll put them back.


l-kent:
Rather than using the address field for the captureState assumes, I'd suggest using the 'line' field from the BAP instead (this is '%000006af' etc.) - I think it's more useful because it directly be compared to the .bir as well as the .adt files. BAP also doesn't give an address for everything, since the statements that are generated by BAP's basic analysis rather than having directly existed in the binary can't have addresses.

I don't really like the generalised attributes approach - different attributes only meaningfully apply to certain things in Boogie, so I'd prefer to keep things specific and only have the options for relevant attributes?

I don't think the {:extern} attribute on an axiom even does anything either, as axioms don't have an identifier to override? 

The IR doesn't use case classes for most of the classes because they contain mutable data. Case classes are only really designed to hold immutable data.

ailrst:
> Rather than using the address field for the captureState assumes, I'd suggest using the 'line' field from the BAP instead (this is '%000006af' etc.)

Yeah I agree that would be better I'll have a look at how much work it is to incorporate that instead. 

> I don't really like the generalised attributes approach - different attributes only meaningfully apply to certain things in Boogie, so I'd prefer to keep things specific and only have the options for relevant attributes?

Yeah, I guess this implementation doesn't prevent you from setting attributes that aren't meaningful in the context, its just a non-trivial amount of refactoring to add in any new attribute, plus it requires another field for each one. I prefer to have a list like this for that convenience. If we really want to I'd define the attributes an enum and the attribute list key as a union of the allowed values.  

> I don't think the {:extern} attribute on an axiom even does anything either, as axioms don't have an identifier to override?

Yeah that's a mistake, thanks.

> The IR doesn't use case classes for most of the classes because they contain mutable data. Case classes are only really designed to hold immutable data.

Cool makes sense, I'll see if I can define an extractor for pattern matching since that would be convenient for this, and also for writing more complex visitors in the future.

src/main/scala/specification/Specification.scala
```diff
@@ -23,7 +23,7 @@ case class SpecGlobal(name: String, override val size: Int, arraySize: Option[In
   override val toAddrVar: BVar = BVariable("$" + s"${name}_addr", BitVecBType(64), Scope.Const)
   override val toOldVar: BVar = BVariable(s"${name}_old", BitVecBType(size), Scope.Local)
   override val toOldGamma: BVar = BVariable(s"Gamma_${name}_old", BoolBType, Scope.Local)
-  val toAxiom: BAxiom = BAxiom(BinaryBExpr(BoolEQ, toAddrVar, BitVecBLiteral(address, 64)))
+  val toAxiom: BAxiom = BAxiom(BinaryBExpr(BoolEQ, toAddrVar, BitVecBLiteral(address, 64)), List((":extern", "")))
```
l-kent, line 26: Axioms being :extern is pointless as mentioned

src/main/scala/ir/Statement.scala
```diff
@@ -64,7 +69,8 @@ class DirectCall(var target: Procedure, var condition: Option[Expr], var returnT
   override def acceptVisit(visitor: Visitor): Jump = visitor.visitDirectCall(this)
 }
 
-class IndirectCall(var target: Variable, var condition: Option[Expr], var returnTarget: Option[Block]) extends Jump {
+case class IndirectCall(var target: Variable, var condition: Option[Expr], var returnTarget: Option[Block],
+                   val address: Option[Int] = None) extends Jump {
```
l-kent, line 73: Should revert back to being class instead of case class

l-kent (changes requested):
This is fine to merge once you've fixed these minor issues and then updated the expected files again.

src/main/scala/specification/Specification.scala
ailrst: fixed

src/main/scala/ir/Statement.scala
ailrst: fixed

l-kent (approved):

PR #136, opened 2023-11-03
Use LinkedHashSet for worklist solver to improve performance

ailrst:
TIP's worklist solver has worse than O(n^2) computational complexity since it uses a `ListSet` which has `O(n)` insertion and removal because it determines if a value exists through a linear scan. 

The little bit of investigation I did into how the current analysis performs (just throwing cntlm at it and profiling and killing it after 10 seconds) makes it look like the immutable data-structures will also be a problem. 
Before this change it didn't even get to the start of the analysis and spent the whole time initialising the work-list with all the procedures. 

Looking at just the constant prop, now approx 37% of the time is spent in `MapLattice.lub()`, 36% in immutable.Set.foreach, and 5% in the transfer function.

image

This is a really weird-looking flamegraph, the big peaks are the same function (Map.get()), and are probably like that because of the profiler not really understanding Scala's functional control flow; they're probably lambdas constructed from lambdas etc. This makes me wonder if the immutable data-structures are also going to thrash the jit compiler. I don't know if there's a way to coerce scala to std::move() immutable datastructures.

`java -agentpath:.../async-profiler-2.9-linux-x64/build/libasyncProfiler.so=start,event=cpu,file=profile-hashsetworklist.html -Xmx8G -jar target/scala-3.3.1/wptool-boogie-assembly-0.0.1.jar -a examples/cntlm-new/cntlm-new.adt -r examples/cntlm-new/cntlm-new.relf --analyse & sleep 30; pkill -f 'target/scala-3.3.1/wptool-boogie-assembly-0.0.1.jar';`


katrinafyi:
Flamegraph towers are possibly due to https://github.com/UQ-PAC/bil-to-boogie-translator/blame/5f02a105c25e3ee7cecb935c310d572ff7d8d6d1/src/main/scala/analysis/Lattice.scala#L123

l-kent:
Scala does weird things with the JVM to achieve some of its functional features, which is why you get that from the profiler. Nearly all the time spent in immutable.Set.foreach is also just MapLattice.lub.

It looks like the MapLattice.lub that's directly called in PushDownWorklistFixpointSolver.process is completely pointless and should be removed:
https://github.com/UQ-PAC/bil-to-boogie-translator/blob/52f9378bd884936f4ee12fb339e61c3d955f60f1/src/main/scala/analysis/solvers/FixPointSolver.scala#L200C40-L200C40 

You should also look at replacing ListSetWorkList with LinkedHashSetWorklist in SimpleMonotonicSolver too.

ailrst:
I'm also not sure what the impact of immutability is really. Adjacent cfg nodes should share most of their sublattice values, and Scala's persistent trie implementation should de-duplicate those. However profiling shows that the GC is very busy due to the MapLattice. Since the outer lattice (over program states) should really just grow to encompass the whole CFG and then just have some values updated it might make sense for it to be mutable. 

But apparently Scala collections are  just slow in general; even mutable ones. Although this paper claims Graal does a good job at optimising down the generics; https://dl.acm.org/doi/abs/10.1145/3136000.3136002

l-kent:
I think the mutable scala collections are generally supposed to be more performant than the immutable ones? There is definitely some stuff that could be easily replaced with mutable collections, such as the outer lattice mapping from CfgNodes to the inner lattice as you say. The inner lattices being immutable does make conceptual sense but there may easily be a more performant way to implement them, mutable or not.

As it currently exists, the MRA doesn't really need to use a lattice at all, it could easily just use a mutable set that it updates with memory regions as it identifies them - there's nothing that's really local to any particular CfgNode with it, but that's getting a bit away from what you've been looking at so far.

l-kent (approved):

Issue #134, opened 2023-11-02
Make IL control-flow representation match boogie more closely


ailrst:
The IL should represent conditional calls, and branching, closer to boogie. This means removing conditions from DirectCalls and replacing them with an `assume` preceding the call. 

E.g. 

```
                        ┌─────────────┐
                        │ statement1  │
                        │ statement2  │
                        │ ----------  │
                        │ goto a      │
                        │ goto b      │
              ┌─────────┴─────────────┴───────────┐
              │                                   │
 ┌────────────▼───────────┐        ┌──────────────▼─────────┐
 │a:assume R0 == 699900   │        │b:assume R0 == 699908   │
 │ --------------------   │        │ ---------------------- │
 │ DirectCall(ProcedureA) │        │ DirectCall(ProcedureB) │
 └────────────────────────┴────┬───┴────────────────────────┘
                               │
                        ┌──────▼──────┐
                        │ statement3  │
                        │ statement4  │
                        │ ----------  │
                        │ goto c      │
                        └─────────────┘
```

l-kent:
It's GoTos that we would want to remove the conditions from, not DirectCalls (I have already removed conditions from DirectCalls in changes for #132 that are not yet pushed).

Issue #97, opened 2023-10-12
Cfg Construction for Jumps


l-kent:
`visitJumps` in the `ProgramCfgFactory' needs to be rewritten as it contains multiple incorrect assumptions about the IR, such as that there are a maximum of two jumps per block, and GoTos can only follow GoTo. These may hold for the input we get from BAP but are not guaranteed to hold if we transform the program representation further (such as by resolving an indirect call to multiple direct calls) and then reconstruct the IR afterwards. A new approach should create the Cfg from the IR as it is presented, without making assumptions.

PR #140, opened 2023-11-13
Boogie-Style IR Control Flow

l-kent:
Implements #134 and also solves #97

I still need to clean a few things up but hopefully this should help @ailrst


ailrst (approved):
Looks good :+1:

PR #141, opened 2023-11-14
Il cfg iterator

ailrst:
Creating a draft PR for feedback in case you have feedback on the direction at this stage. Still to do are:

- [x] Merge https://github.com/UQ-PAC/bil-to-boogie-translator/pull/140/
- [x] Merge analysis reworked types in main
- [x] More testing for the constprop analysis and the block-level CFG  
- [x] Clean up how it handles block entry/exit notes and procedure entry/exit nodes
- [ ] Finish and test the interprocedural version
- [x] Finish the Dot and text format analysis result output
- [x] Maybe also want to be able to return the position at the beginning or end of a given block or procedure? 

The goal of this interface is to remove the need for a separate CFG structure, and separate 'id's for IL nodes that have to be resolved somehow. Broadly this change adds:

- Parent node to Blocks and Commands
- Adds reverse block-level CFG to Block
- Adds reverse procedure-level CFG to Procedure
- Encapsulates modification to jump lists in methods that maintain the CFG
- Adds methods to return the successor and predecessor of any IL node
- Adds a simple implementation of constant-prop and Direction to using this as the CFG (without changing the underlying solver)
- Cleans up the `intra` flag in the existing CFG 
- Moves blocks' statement list to an IntrusiveList which allows getting the successor or predecessor of any list element in O(1)

This means the CFG part of the analysis domain is just a reference to an IL procedure, statement or command, and we have a pure function that returns the adjacent set of positions in any given direction; e.g. forwards or backwards in control flow. 

See docs/il-cfg.md for more details.


l-kent:
It'll be easier to give feedback for this once you've merged in the changes from #140

src/main/scala/ir/Block.scala
```diff
@@ -0,0 +1,86 @@
...
+    (j, newJ) match {
+       case (g: GoTo, f: GoTo) => {
+         removeJump(j)
+         addJump(newJ)
+       }
+       case (g: Call, f: Call) => {
+         removeJump(g)
+         addJump(f)
+       }
+       case (_, _) => throw Exception("Programmer error: can not replace jump with call or vice versa")
```
l-kent, line 54: Why is this not allowed? It is probably necessary for resolving jump tables, where an indirect call has to be resolved to gotos, not direct calls.

ailrst:
The nondeterministic GoTo allows a having no targets, what is the semantics of that in boogie / what does it get translated to? I assume if it is translated naively boogie would fall through to the next block, but cant find right now what it actually gets translated to.

l-kent:
There isn't a valid translation - it just translates to 'goto ;' or something like that which would cause a syntax error in Boogie. It's not a case that should ever make it to that point - throwing an exception if it's encountered could certainly be done.

ailrst:
This is reasonably ready to review now, the const-prop just hasn't really been validated, and there's some cleanup stuff to do. 
With regards to the procedure exit node we discussed, you can see I've added an explicit return block and entry block but its handling is a bit messy because its only added when the procedure contains any blocks, and they're not stored in order anymore. 
I don't currently have a separate 'call node' and 'call return node' analog; i.e. the landing nodes for the call and call return; they don't really seem neccessary but I could easily be missing something?

l-kent:
Does a block's Jump end up as a node in the analysis? If so, there isn't any real need for the previous call/call return nodes. If not, I would suggest making them exist as nodes in the analysis rather than using the old call/call return nodes.

ailrst:
> Does a block's Jump end up as a node in the analysis? 

Yes they do, but their successor is just the block they jump to; or for calls, the return node if defined. The idea of a 'non terminating / non-halting  procedure exit' is represented by the node having no successors.

src/main/scala/analysis/BasicIRConstProp.scala
```diff
@@ -0,0 +1,91 @@
...
+      case un: UnaryExpr =>
+        val arg = eval(un.arg, env)
+
+        un.op match
+          case BVNOT => bvnot(arg)
+          case BVNEG => bvneg(arg)
+
+      case _ => valuelattice.top
+
+  val calleePreservedRegisters = Set("R0", "R1", "R2", "R3", "R4", "R5", "R6", "R7", "R8", "R9", "R10", "R11")
```
l-kent, line 64: These are not the callee-preserved registers, which are the ones that must be preserved by a subroutine call. Those are R19-R29 and R31.

What this should actually be called is the caller-preserved registers (which means it's the caller's responsibility to preserve them if it wants to) and it should consist of R0 to R18, and R30.

l-kent:
> Yes they do, but their successor is just the block they jump to; or for calls, the return node if defined.

For calls, the return target is only the direct successor if it's an intraprocedural analysis, correct?

> The idea of a 'non terminating / non-halting procedure exit' is represented by the node having no successors.

By 'non terminating / non-halting procedure exit' you mean a non-returning call? That's different to it being non-terminating - a call to exit(), for instance, is a non-returning call but is terminating. The other thing that a non-returning calls can be is a tail call, which in the intraprocedural analysis should just have no successors for now.

src/main/scala/ir/package.scala
```diff
@@ -0,0 +1,3 @@
+package object ir {
+
+}
```
l-kent, line 3: What's the point of this empty package object?

ailrst: Don't know it must be a mistake or generated by intellij

ailrst:
> For calls, the return target is only the direct successor if it's an intraprocedural analysis, correct?

Currently the interprocedural is a strict superset of the intraprocedural, so no. But I haven't really put much thought into interprocedural yet; for direct calls it probably shouldn't include the return edge but the worklist could theoretically just delay traversing that edge until after the call has been analysed. 
It could be made the responsibility of the analyses to choose the correct over-approximation for returning unresolved indirect calls, although it does seem like it would make sense to have some 'unknown call body' node? 

> By 'non terminating / non-halting procedure exit' you mean a non-returning call?

Yes

src/main/scala/util/RunUtils.scala
```diff
@@ -328,21 +449,19 @@ object RunUtils {
...
-              val newCall = DirectCall(targets.head, indirectCall.returnTarget, indirectCall.label)
-              block.jump = newCall
+              block.replaceJump(DirectCall(targets.head, indirectCall.returnTarget, indirectCall.label))
             } else if (targets.size > 1) {
               modified = true
               val procedure = c.parent.data
               val newBlocks = ArrayBuffer[Block]()
               for (t <- targets) {
-                val assume = Assume(BinaryExpr(BVEQ, indirectCall.target, BitVecLiteral(t.address.get, 64)))
+                val assume = Assume(BinaryExpr(BVEQ, indirectCall.target, BitVecLiteral(t.address.get, 64)), null)
```
l-kent, line 458: Why add this just to make the comment null? Why use null at all?

src/main/scala/translating/IRToBoogie.scala
```diff
@@ -415,8 +415,8 @@ class IRToBoogie(var program: Program, var spec: Specification) {
       // collects all targets of the goto with a branch condition that we need to check the security level for
       // and collects the variables for that
       val conditions = g.targets.flatMap(_.statements.headOption).collect { case a: Assume if a.checkSecurity => a }
-      val conditionVariables = conditions.flatMap(_.body.variables).distinct
-      val gammas = conditionVariables.map(_.toGamma).sorted
+      val conditionVariables = conditions.flatMap(_.body.variables)
+      val gammas = conditionVariables.map(_.toGamma).toList.sorted
```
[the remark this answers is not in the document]
ailrst: GoTo.targets is a Set now for arbitrary insertion and removal, so conditionVariables gets inferred to be a set too; the toList here is to make this have an order/be sortable.

l-kent:
> for direct calls it probably shouldn't include the return edge but the worklist could theoretically just delay traversing that edge until after the call has been analysed.

A call node definitely should not have the return target as a direct successor in an interprocedural analysis.

> It could be made the responsibility of the analyses to choose the correct over-approximation for returning indirect calls on the call command itself, although it does seem like it would make sense to have some 'unknown call body' node?

I think it would make sense to just use the call command itself to deal with that. The exact same thing can be done for direct calls for intraprocedural analyses?

src/main/scala/analysis/Dependencies.scala
```diff
@@ -21,11 +22,22 @@ trait Dependencies[N]:
...
+
+trait IntraProcDependencies extends Dependencies[IntraProcIRCursor.Node]:
+  override def outdep(n: IntraProcIRCursor.Node): Set[IntraProcIRCursor.Node] = IntraProcDependencies.outdep(n)
+  override def indep(n: IntraProcIRCursor.Node): Set[IntraProcIRCursor.Node] = IntraProcDependencies.indep(n)
+
+/** Dependency methods for forward analyses.
+  */
+object IntraProcDependencies extends Dependencies[IntraProcIRCursor.Node]:
+  override def outdep(n: IntraProcIRCursor.Node): Set[IntraProcIRCursor.Node] = IntraProcIRCursor.succ(n)
+  override def indep(n: IntraProcIRCursor.Node): Set[IntraProcIRCursor.Node] = IntraProcIRCursor.pred(n)
```
l-kent, line 43: Is there really any point to doing this indirectly with an object?

src/main/scala/analysis/Cfg.scala
```diff
@@ -426,7 +427,9 @@ class ProgramCfgFactory:
       cfg.addEdge(funcEntryNode, funcExitNode)
     } else {
       // Recurse through blocks
-      visitBlock(proc.blocks.head, funcEntryNode)
+      visitBlock(proc.entryBlock.get, funcEntryNode)
+      // If it has no entry-block we still visit the exit block because VSA analysis expects everything to have an Exit
+      visitBlock(proc.returnBlock, funcEntryNode)
```
l-kent, line 432: If the procedure has no entry block, an exception will be thrown by entryBlock.get, so this isn't really doing what you want at all anyway.

src/main/scala/cfg_visualiser/DotTools.scala
```diff
@@ -12,6 +12,7 @@ object IDGenerator {
 }
 
 def wrap(input: String, width: Integer = 20): String =
+  return input
```
l-kent, line 15: This method is now completely redundant and should be removed entirely?

src/main/scala/ir/Statement.scala
```diff
@@ -54,51 +61,92 @@ class Assert(var body: Expr, var comment: Option[String] = None, override val la
...
+  }
+
+
+  override def toString: String = s"${labelStr}NonDetGoTo(${targets.map(_.label).mkString(", ")})"
   override def acceptVisit(visitor: Visitor): Jump = visitor.visitGoTo(this)
 }
 
 object GoTo:
-  def unapply(g: GoTo): Option[(ArrayBuffer[Block], Option[String])] = Some(g.targets, g.label)
+  case class ConstructionPattern(targets: Iterable[Block], label: Option[String] = None)
```
l-kent, line 124: What is this?

src/main/scala/ir/Program.scala
```diff
@@ -179,17 +318,64 @@ class Procedure(
...
+    _jump.asInstanceOf[GoTo].addAllTargets(targets)
+    this
+  }
+
+  def replaceJump(j: Jump): this.type = {
+    _jump.deParent()
+    j.setParent(this)
+    _jump = j
+    this
+  }
```
l-kent, line 364: Why do these return `this`? Is there a need for addGoToTargets to even exist?

src/main/scala/ir/Program.scala
```diff
@@ -111,25 +170,105 @@ class Program(var procedures: ArrayBuffer[Procedure], var mainProcedure: Procedu
...
+   */
+  def blocks: Seq[Block] =
+    (entryBlock match
+      case Some(b) if _blocks.nonEmpty => Seq(b) ++ _blocks.filter(x => x ne b).toSeq
+      case _ => _blocks.toSeq)
+     ++ (if _blocks.nonEmpty then Seq(returnBlock) else Seq())
+
+  def removeCaller(c: Call): Unit = {
+    _callers.add(c)
+  }
```
l-kent, line 211: This isn't doing what it says it does

src/main/scala/ir/Program.scala
```diff
@@ -111,25 +170,105 @@ class Program(var procedures: ArrayBuffer[Procedure], var mainProcedure: Procedu
...
+                  var entryBlock: Option[Block],
+                  private val _blocks: mutable.HashSet[Block],
+                  var in: ArrayBuffer[Parameter],
+                  var out: ArrayBuffer[Parameter],
+                ) {
+  /* First block executed when the procedure begins */
+  /* returnBlock: Single return point for all returns from the procedure to its caller; always defined but only included
+   * in `blocks` if the procedure contains any real blocks.*/
+  val returnBlock: Block = new Block(name + "_return", None, List(), new IndirectCall(Register("R30", BitVecType(64)), None, Some(name)))
+  returnBlock.setParent(this)
```
l-kent, line 185: Why does this need to be defined for every procedure? Not all procedures will have any returns.

src/main/scala/ir/Program.scala
```diff
@@ -111,25 +170,105 @@ class Program(var procedures: ArrayBuffer[Procedure], var mainProcedure: Procedu
...
+
+  /**
+   * Horrible, compensating for not storing the blocks in-order and storing initialBlock and returnBlock separately.
+   * @return
+   */
+  def blocks: Seq[Block] =
+    (entryBlock match
+      case Some(b) if _blocks.nonEmpty => Seq(b) ++ _blocks.filter(x => x ne b).toSeq
+      case _ => _blocks.toSeq)
+     ++ (if _blocks.nonEmpty then Seq(returnBlock) else Seq())
```
l-kent, line 207: Why does returnBlock need to be stored separately? This also would seem to be non-deterministic, since `_blocks` is a Set, which is very undesirable.

Having this be reconstructed every time `Procedure.blocks` is called seems terribly inefficient. It should instead just provide an iterator over `_blocks`?

src/main/scala/ir/Program.scala
```diff
@@ -111,25 +170,105 @@ class Program(var procedures: ArrayBuffer[Procedure], var mainProcedure: Procedu
...
-    var blocks: ArrayBuffer[Block],
-    var in: ArrayBuffer[Parameter],
-    var out: ArrayBuffer[Parameter]
-) {
-  def calls: Set[Procedure] = blocks.flatMap(_.calls).toSet
+class Procedure private (
+                  var name: String,
+                  var address: Option[Int],
+                  var entryBlock: Option[Block],
+                  private val _blocks: mutable.HashSet[Block],
```
l-kent, line 177: What's the point of making this a set? This seems like it will only cause problems for us, since we want to output blocks in a consistent order.

src/main/scala/ir/Program.scala
ailrst: > Is there a need for addGoToTargets to even exist?

It's a transformation we might want but if needed it should be moved to GoTo not block

They return this for convenience of chaining modifications

src/main/scala/analysis/Dependencies.scala
ailrst: Not quite sure what you mean, but Dependencies is just to allow swapping succ and pred when we want to consider backwards dependencies and decouple the implementation from TIP's expected interface

src/main/scala/ir/Program.scala
ailrst: Its a set for O(1) insertion removal if we modify the IL. My understanding is the block order does not have any meaning beyond the first block, Set's order is not defined but is deterministic it seems. I don't think its neccessary for return block to be stored separately, a few things can probably be cleaned up by just inserting it when the transformation is done.

src/main/scala/analysis/Dependencies.scala
l-kent: The IntraProcDependencies trait calls the methods from the IntraProcDependencies object - why does the object need to exist at all?

src/main/scala/ir/Program.scala
ailrst: My understanding is the block order does not have any meaning beyond the first block. The set removes the notion of an index at all and gives removal using the object identity instead, while also being faster. But I guess that only matters under heavy modification.

src/main/scala/ir/Program.scala
l-kent: The block order doesn't have any meaning beyond the first block, but using a Set does mean the output order is fragile which is very undesirable for being able to track changes to the output. It also makes the output harder to follow if we further lose correspondence to the original BAP input.

We don't even really care about arbitrary removal right now - the only time blocks are removed at present is when external methods are stubbed out (removing all blocks at once), and anything else is likely to involve iterating over all blocks, so we could really just use a ListBuffer to get O(1) insertion/removal.

src/main/scala/ir/Statement.scala
ailrst: Forgot to delete oops

src/main/scala/ir/Program.scala
l-kent: LinkedHashSet would be another good option yeah

l-kent:
Broadly, this approach is definitely a big improvement over what existed before by making it so there isn't the difficulty in trying to relate the analysis results back to the IR, as the IR elements are now directly used in the CFG. However, it still needs to calculate the predecessor and successor sets for everything, even simple traversal of the statements within a block which seems to result in things still being rather convoluted in some ways.

I want to try modifying the analysis solvers, which should allow for things to be significantly simplified.

src/main/scala/ir/Program.scala
l-kent: Although that still requires storing the entry block separately which still makes creating the iterator awkward, so I'd really just lean towards ListBuffer. I guess most cases where we iterate over all blocks don't care whether the entry block is first, except the final output, so taking that into account would allow for other possibilities.

src/main/scala/ir/Program.scala
```diff
@@ -116,52 +120,265 @@ class Program(var procedures: ArrayBuffer[Procedure], var mainProcedure: Procedu
     initialMemory = initialMemoryNew
   }
 
+  class ILUnorderedIterator(private val begin: Program) extends Iterator[CFGPosition] {
```
ailrst, line 123: @l-kent Do you want to keep this in? it's just a convenient way to .map() and .filter() etc. the program

ailrst (commented):
I's really like to replace 'CFGPosition' with a sealed trait but that would mean moving procedure, block, and command into one file. 

I think this should be OK to merge now, hopefully it doesn't break anything. 

After merging we need to sort out the interprocedural situation. 

I think it would be useful for have a deep-copy function for the mutable part of the IL for e.g. temporary analyses and inlining.

src/main/scala/ir/Program.scala
l-kent: I think you should make it clear what order it traverses the blocks within a procedure in, for instance (otherwise it could be prone to misuse). Would the idea to just be use .map() with the results of an analysis? I'm not entirely sure how useful that is

src/main/scala/util/RunUtils.scala
```diff
@@ -153,20 +160,58 @@ object RunUtils {
...
+
+      //config.analysisResultsPath.foreach(s => writeToFile(printAnalysisResults(IRProgram, newCPResult), s"${s}_newconstprop$iteration.txt"))
+      config.analysisResultsPath.foreach(s => writeToFile(toDot(IRProgram), s"program.dot"))
+      config.analysisResultsPath.foreach(s => writeToFile(toDot(IRProgram, newCPResult.map((k,v) => (k, v.toString))), s"program-constprop.dot"))
+
+      config.analysisResultsPath.foreach(s => writeToFile(printAnalysisResults(IRProgram, newCPResult), s"${s}_new_cpres$iteration.txt"))
+      config.analysisResultsPath.foreach(s => writeToFile(printAnalysisResults(IRProgram, cfg, constPropResult), s"${s}_old_cpres$iteration.txt"))
+
+    }
+    newSolverTest()
```
l-kent, line 196: This should probably be in a test suite or something instead of being a core part of the analysis? Change that and then we can merge everything in.

src/main/scala/ir/Program.scala
ailrst: I just commented it, its a very special case visitor I guess, probably only useful in very specific cases to map over the whole program, e.g. collect indirect calls and see if they were resolved. E.g. a simpler way to get the analysis domain.

src/main/scala/util/RunUtils.scala
ailrst: I've just removed this and made it run the analysis and output the results like the other constprop analysis. The results won't be the same anyway because this handles calls more carefully.

l-kent: That makes sense. What's the difference with handling calls, does it do some sort of havoc instead of just skipping over them in the intraprocedural analysis?

l-kent:
It should be good to merge now, just make sure that the expected files are updated properly and haven't gotten muddled up in the merges.

PR #144, opened 2023-11-16
Analysis Cleanup and Explicit Types

l-kent:
This is a number of changes to the analysis code.

One of the main changes is making all the types in the analysis explicitly declared instead of relying on a maze of .Elements with convoluted inheritance patterns. I personally find this makes the code easier to follow. I am curious for feedback here from @ailrst and @yousifpatti  - is this an improvement? 

Other changes here:
- Consolidates and cleans up parts of the analysis, to reduce the convoluted inheritance patterns
- Enables procedure inlining in the CFG so that there are actually interprocedural edges in the CFG
- Makes interprocedural analyses actually use the interprocedural edges (before they were in practice not meaningly different from the intraprocedural analyses) but this still doesn't seem to be completely working properly, there might be an issue with the inlining? It might make sense to just make the VSA an intraprocedural analysis for now? It isn't worth trying to fix the inlining since #141 is in the works.
- Removes edges from the CFG (no longer necessary now that jumps don't have conditions)
- Significantly improves the way intra/interprocedural analyses are handled so there isn't a boolean 'intra' flag carried around everywhere
- Makes the SimpleMonotonicSolver obtain its initial set of nodes in a more modular way.

### DIFF
--- a/src/main/scala/analysis/Analysis.scala
+++ b/src/main/scala/analysis/Analysis.scala
@@ -170,6 +170,8 @@ trait MemoryRegionAnalysis(val cfg: ProgramCfg,
 
   val domain: Set[CfgNode] = cfg.nodes.toSet
 
+  val first: Set[CfgNode] = domain.collect { case n: CfgFunctionEntryNode if n.predIntra.isEmpty => n }
+
   private val stackPointer = Register("R31", BitVecType(64))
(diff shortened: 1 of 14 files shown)


yousifpatti:
I am happy with these changes. Are you going to merge them soon I wish to merge the lifted lattice stuff as well?

PR #152, opened 2024-01-05
Stream writer boogie file translation and function trimming for large binaries

ailrst:
Make it so we can again lift cntlm to a state boogie can parse (note you need to give the JVM more memory for the adt to parse). 

Changes

- use a Writer to output the boogie file rather than concatenating to a string
- coalesce the 'constant' data section specifications into 64-bitvectors where allowed rather than always using 8-bitvectors. (the old encoding is preserved but not linked up to a cli flag)
- separate specifications from implementations in the boogie output
- add options to set the main function and function call depth for the reach-ability trimming, omitting the implementation for the leaf functions
- Fixes a bug with the assume() translation with coercing the condition to a bool, since bap seems to output both bools and bitvectors at different times


l-kent:
> Fixes a bug with the assume() translation with coercing the condition to a bool, since bap seems to output both bools and bitvectors at different times

Can you provide an example of this?

src/main/scala/translating/IRToBoogie.scala
```diff
@@ -102,6 +104,18 @@ class IRToBoogie(var program: Program, var spec: Specification) {
...
+   * (Boogie does not like this it if it is too large due to it being a single expression)
+   *
+   * E.g.
+   *  val readOnlyMemoryFunction = readOnlyMemoryPredicate(memoryToCondition(program.readOnlyMemory), mem)
+   *  val readOnlyMemory = List(BFunctionCall(readOnlyMemoryFunction.name, List(mem), BoolBType))
+   */
+  private def readOnlyMemoryPredicate(readonly: List[BExpr], mem: BVar) : BFunction = {
+    BFunction("readonly_memory", List(BParam("mem", mem.bType)), BParam(BoolBType), Some(readonly.reduce((a, b) => BinaryBExpr(BoolAND, a, b))), List(externAttr))
+  }
+
```
l-kent, line 118: This isn't actually used is it? This approach was already tried and does not work

src/test/correct/arrays_simple/clang/arrays_simple.expected
```diff
@@ -40,145 +63,58 @@ procedure {:extern} rely();
...
-
-procedure {:extern} rely_transitive()
+  free ensures (memory_load64_le(mem, 69064bv64) == 1808bv64);
+  free ensures (memory_load64_le(mem, 69072bv64) == 1728bv64);
+  free ensures (memory_load64_le(mem, 69592bv64) == 1812bv64);
+  free ensures (memory_load64_le(mem, 69672bv64) == 69672bv64);
+
+
+
+procedure {:extern} rely_transitive();
```
l-kent, line 73: There are a lot of very extraneous line breaks added

src/test/correct/arrays_simple/clang/arrays_simple.expected
```diff
@@ -1,37 +1,60 @@
...
 var {:extern} stack: [bv64]bv8;
+
 const {:extern} $_IO_stdin_used_addr: bv64;
 axiom ($_IO_stdin_used_addr == 1872bv64);
+
 function {:extern} {:bvbuiltin "bvadd"} bvadd64(bv64, bv64) returns (bv64);
+
 function {:extern} gamma_load32(gammaMap: [bv64]bool, index: bv64) returns (bool) {
-  (gammaMap[bvadd64(index, 3bv64)] && (gammaMap[bvadd64(index, 2bv64)] && (gammaMap[bvadd64(index, 1bv64)] && gammaMap[index])))
+  (gammaMap[bvadd64(index, 3bv64)]&&(gammaMap[bvadd64(index, 2bv64)]&&(gammaMap[bvadd64(index, 1bv64)]&&gammaMap[index])))
```
l-kent, line 27: What's caused this to change? I haven't played around with it yet but I can't obviously see a cause yet because it doesn't seem to happen when && is used for Gamma assignments. It does also happens with ++ in memory_loads though.

l-kent:
> separate specifications from implementations in the boogie output

Is there a particular reason we want to do this now?

src/test/correct/arrays_simple/clang/arrays_simple.expected
l-kent: Oh the && in Gamma assignments is unrelated, so my push fixes this

ailrst:
> > Fixes a bug with the assume() translation with coercing the condition to a bool, since bap seems to output both bools and bitvectors at different times
> 
> Can you provide an example of this?

examples/cntlm-new creates conditionals with `bvslt9(LocalVar(#3331, bv9), 32bv9)` which are type bool.

src/main/scala/translating/IRToBoogie.scala
ailrst: Its not used

l-kent: Probably best to remove it then?

l-kent:
> It does let you override the spec and implementation separately though.

Is this something we want to do or may want to in future?

> examples/cntlm-new creates conditionals with bvslt9(LocalVar(#3331, bv9), 32bv9) which are type bool.

It's great to have an example of BAP actually producing one of those comparison operations, thanks. The tricky thing with the comparator instructions is that technically, directly translating them (as is currently done in BAPtoIR) is incorrect, as the BAP SLT returns a bitvector of size 1, but the SMT/Boogie SLT returns a boolean (BAP doesn't distinguish between these as types but SMT/Boogie do). Knowing these can occur inside the BAP conditionals means we can specifically address them now.

l-kent (approved):
Should be good to merge now

PR #154, opened 2024-01-12
Mill build system

ailrst:
(merged into staging)

Setup the project and tests with mill build system

It seems slightly faster and the command system is much easier to use than SBTs so we could hopefully migrate some of the ad-hoc scripts to the build config eventually

This doesn't remove SBT yet they both seem to coincide ok but we should drop sbt at some point.


utting:
Can you please add a .mill_version file so that we all use the same version?
On MacOS 14.2.1 I get this deprecated warning when using mill 0.11.6 (latest):
[warn] /Users/uqmuttin/git/basil/bil-to-boogie-translator/antlr.sc:12:28: type Sources in package define is deprecated (since Mill after 0.11.0-M8): Use mill.define.Target instead.
[warn]   def antlrGrammarSources: Sources
[warn]                                                     ^

And then it does not build the antlr sources, so fails to compile at: import Parser.*
How does the import $file.antlr in build.sc work?

ailrst:
I'm using the same mill version but the config for the antlr sources was wrong. I added the mill-version file.

utting:
Works nicely now thanks   :-) 
(I renamed .mill_version to .mill-version and committed that).
I can compile, run, and the secret_write output goes through Boogie successfully as usual.

Can you please add a couple of sentences to readme.md to tell people how to use mill?

And maybe we should install and commit the ./mill bootstrap script (I used https://github.com/lefou/millw) so that everyone can easily use mill?  There is a windows millw.bat there too.

I cannot see any mill plugins for IntelliJ, but personally I'm happy to run ./mill from the command line.  What IDE support do you think others might want?

ailrst:
I mainly use intellij and Mill is supposed to have build server protocol support and intellij project generation, https://mill-build.com/mill/Installation_IDE_Support.html#_intellij_idea_support I want to at least check that connecting the debugger in intellij still works. I don't tend to use VSCode, I'm not sure what integration to expect there.

PR #160, opened 2024-01-25
Ir iterator interproc

ailrst:
(merged into staging)

Addresses https://github.com/UQ-PAC/bil-to-boogie-translator/issues/157 and https://github.com/UQ-PAC/bil-to-boogie-translator/issues/156 and https://github.com/UQ-PAC/bil-to-boogie-translator/issues/168

- Interprocedural iterators now return the fallthrough target for indirect calls, and for directcalls, and only returns the procedure when the target has an implementation.
- Adds fallthrough edge to blocks as an optional GoTo. This means it is now included in incomingJumps sets etc. It also means the reverse interprocedural iterator jumps to the called-procedure's return call from the aftercall goto.
- Distinct return blocks are added
- Defines extension methods for checking whether a cfgnode is a procedureEntry, procedureReturn, afterCall node. 

The interface for checking is like this: 
```
        b match {
          case b: GoTo if b.isAfterCall => ...
          case _                => ...
        }
```


ailrst:
Bug; the aftercall flag is not updated when the indirect call is resolved so the pred() / reverse iterator visits an indirect call that is detached from the IL.

ailrst:
This is already part of `staging`

PR #163, opened 2024-02-01
Library procedure RG contradiction proof

ailrst:
Implements the proof by contradiction style of procedure rely/guarantee. 

This should be easier for boogie to reason about.


l-kent:
You probably shouldn't have committed whatever beans and beans2 are?

This way of embedding boogie directly in the specifications seems messy since it's inconsistent with the existing `Ensures DIRECT:` syntax? 

> this isn't great because obviously variables and globals aren't captured for modifies inference etc

I don't think this should be a problem in practice for the external procedure rely/guarantees. It very much will be if you use it for the thread-level rely/guarantees

PR #166, opened 2024-02-07
Irreducible Loop Resolution

Thomas-Malcolm:
In [a separate repo](https://github.com/UQ-PAC/reducible-loop-transform) we implemented an algorithm for identifying irreducible loops in control flow, and subsequently transforming them to be reducible. This PR introduces this functionality to BASIL, operating on the IR.


ailrst (approved):
This is is working and good to merge once https://github.com/UQ-PAC/bil-to-boogie-translator/pull/160 is merged. If we want we could refactor this to work at the block level rather than the statement level which would be slightly more efficient.

PR #174, opened 2024-02-14
Liveness and Parameter Analyses

sadrabt:
Implements an Intraprocedural Liveness Analysis using the worklist solver
Implements bi-directional IDE Solver
Implements an Interprocedural Liveness Analysis using the backward IDE solver
Implements a parameter analysis that uses inter and intra-procedural liveness analyses to identify function parameters

The Interprocedural Liveness Analysis works by inlining the alive registers into and out of function calls. As such, it results in an overapproximation of the parameters. The example below is such a case with R0 being the only parameter to Foo, however, since R1 is alive after the call to Foo in Bar it gets propagated through to Foo and since it does not die (get written to like R1 = X), it will be alive at the function entry point to Foo.

```
Proc: Bar
R0 = 1
R1 = 2
Call Proc Foo
R2 = R1

Proc: Foo
R3 = R0
```

The Parameter analysis uses these results to find the parameters by removing the registers that are propagated but **not** used in the function/passed as parameters to callees.

Additionally, there is a known case of infinite recursion (no base case branch) that breaks the Interprocedural Liveness analysis. Furthermore, the parameter analysis requires the callees of a function being analysed to already have resolved parameters, therefore, it may fail in mutually recursive call patterns. (The analysis produces a result for all the correct and incorrect examples in the repo.)


src/main/scala/analysis/solvers/IDESolver.scala
```diff
@@ -0,0 +1,279 @@
...
+        case ((n, d1, d2), e) =>
+          if isExit(n) then
+            val exit: EE = n.asInstanceOf[EE]
+            val proc = IRWalk.procedure(exit)
+            val m1 = res.getOrElseUpdate(proc, mutable.Map[DL, mutable.Map[DL, EdgeFunction[T]]]().withDefaultValue(mutable.Map[DL, EdgeFunction[T]]()))
+            val m2 = m1.getOrElseUpdate(d1, mutable.Map[DL, EdgeFunction[T]]())
+            m2 += d2 -> e
+
+      }
+      Logger.info(s"Function summaries:\n${
```
ailrst, line 132: This is a bit too much information for the `info` log level, can we make it debug instead?

src/main/scala/analysis/Cfg.scala
```diff
@@ -311,10 +313,12 @@ class ProgramCfg:
...
 
           case (from: CfgCommandNode, to: CfgFunctionEntryNode) =>
-            DotInlineArrow(dotNodes(n), dotNodes(to))
+//            DotInlineArrow(dotNodes(n), dotNodes(to))
+            dotArrows += DotInterArrow(dotNodes(n), dotNodes(to))
 
           case (from: CfgFunctionExitNode, to: CfgNode) =>
-            DotInlineArrow(dotNodes(n), dotNodes(to))
+//            DotInlineArrow(dotNodes(n), dotNodes(to))
+            dotArrows += DotInterArrow(dotNodes(n), dotNodes(to))
```
ailrst, line 321: Does this need to be reverted?

src/main/scala/analysis/InterLiveVarsAnalysis.scala
```diff
@@ -0,0 +1,143 @@
...
+  import edgelattice.{IdEdge, ConstEdge}
+
+  def edgesCallToEntry(call: GoTo, entry: Command)(d: DL): Map[DL, edgelattice.Element] = {
+    // this analysis is implemented in order to Identify parameters and function interface
+    // if parameters are known this should map only parameter registers to IdEdge() or in this case
+    // return (since it is a backward analysis)
+    // all other dataflow facts except lambda should be mapped to bottom (return empty Map())
+    //    d match
+    //      case Left(value) => Map()
+    //      case Right(_) => Map(d -> IdEdge())
```
ailrst, line 29: Can this comment be deleted

src/test/scala/LiveVarsAnalysisTests.scala
```diff
@@ -0,0 +1,350 @@
...
+    val blocks = program.blocks
+    val liveVarAnalysisResults = InterLiveVarsAnalysis(program).analyze()
+
+    assert(liveVarAnalysisResults(blocks("branch1")) == Map(R1 -> Top, r30 -> Top))
+    assert(liveVarAnalysisResults(blocks("branch2")) == Map(R2 -> Top, r30 -> Top))
+    assert(liveVarAnalysisResults(blocks("lmain")) == Map(R1 -> Top, R2 -> Top, r30 -> Top))
+
+  }
+
+  def recursionInfinite(): Unit = { // can't handle this infinite recursion case
```
ailrst, line 247: This just fails because the live variable doesnt propagate back over the infinite recursion right? The test currently doesn't have an assertion so passes trivially, it would be better to have the appropriate assertions and disable the test with the `ignore` flag. Its also ok to leave it here failing so we know its an issue.

ailrst (commented):
Its really good to have tests for everything, I haven't looked too closely at the analyses yet, but there's a few cleanup things to do.

src/main/scala/ir/Program.scala
```diff
@@ -405,7 +405,7 @@ class Block private (
   override def toString: String = {
     // display all statements and jumps
     val statementsString = statements.map(_.toString).mkString("\n")
-    s"Block $label with $statementsString\n$jump"
+    s"Block $label" // with $statementsString\n$jump"
```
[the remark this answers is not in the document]
sadrabt: I wrote this for now
type CFGPosition = Procedure | Block | Command

extension (p: CFGPosition)
  def toShortString: String =
    p match
      case procedure: Procedure => procedure.toString
      case block: Block => s"Block ${block.label}"
      case command: Command => command.toString

src/main/scala/analysis/AnalysisResult.scala
```diff
@@ -0,0 +1,12 @@
...
+
+/**
+ * Analysis Results String utility trait
+ * @tparam T type of analysis results
+ */
+trait AnalysisResult[T] {
+
+  def encodeAnalysisResults(result: T) : String
+
+  def parseAnalysisResults(input: String): String
```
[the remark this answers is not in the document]
sadrabt: These are used in tests that use loadandtranslate function in run utils. Only, made this trait to keep function signatures consistent across my analyses. I have added comments clarifying this but not sure if there is a better way to test currently before run utils refactoring. I can do something like assign analysis results to a publicly available member variable of run utils but

src/main/scala/util/IntrusiveList.scala
```diff
@@ -90,62 +78,34 @@ final class IntrusiveList[T <: IntrusiveListElement[T]] private (
...
-  /**
-   * Unsafely return the first element of the list.
-   */
   override def head(): T = firstElem.get
 
   override def headOption(): Option[T] = firstElem
 
-  /**
-   * Unsafely return the first element of the list.
-   */
```
ailrst, line 142: I'm not sure why these have disappeared, you may have to merge `staging` again

PR #177, opened 2024-02-16
Runutils cleanup

ailrst:
Refactors runutils to more cleanly separate the loading, analysis, and translation phases. 

We now have 3 objects, `Loading` `IRTransform` `StaticAnalysis` which contain only methods and no state, and 3 case classes, `IRContext`, `StaticAnalysisContext` and `BasilResult` which hold the results from each phase. 

We might still want to break up the `StaticAnalysis.analyze` further in the future.


katrinafyi (commented):
I can't really review this for how appropriate the new structure is but more separation is usually good. It is not super clear to me what order things happen in (is it top to bottom? how are tasks divided between loading/transform/analysis?) . Some comments might help a person like me.

katrinafyi (approved):

PR #179, opened 2024-02-19
Merge staging into main

ailrst:
Contains the following changes

- [x] Steensgaard https://github.com/UQ-PAC/bil-to-boogie-translator/pull/147
- [ ] The test/analysis/*/dump folders should be deleted from the repo and added to .gitignore
- [ ] Remove --analyse from tests

- Liveness analysis and function parameter identification https://github.com/UQ-PAC/bil-to-boogie-translator/pull/174
- Irreducible loop transformation https://github.com/UQ-PAC/bil-to-boogie-translator/pull/166
- Procedure rely/guarantee proof by contradiction https://github.com/UQ-PAC/bil-to-boogie-translator/pull/163
- Mill build system https://github.com/UQ-PAC/bil-to-boogie-translator/pull/172
- Interprocedural IR iterator fixes https://github.com/UQ-PAC/bil-to-boogie-translator/pull/171
- Refactor RunUtils https://github.com/UQ-PAC/bil-to-boogie-translator/pull/177
- Fix conditional goto https://github.com/UQ-PAC/bil-to-boogie-translator/pull/169


l-kent:
The readme should be updated in relation to the switch from sbt to mill.

The test/analysis/*/dump folders should be deleted from the repo and added to .gitignore

Rather than dumping text files and matching the results for those analysis tests, it would be much better if the tests can more cleanly get the results of the analysis internally, using data structures to represent the important parts of the analysis results rather than strings. For the parameter analysis, you just need a mapping of procedure name to registers that are parameters, which is simple to set up. 

It would be good to organise the analysis tests by which analyses they are relevant to. 

Having both `Ensures DIRECT: " "` and `Ensures: DIRECT :: " "` as syntax in the specifications is very messy and confusing. I know a more substantial overhaul of the specifications is planned soon, but this does not seem like the cleanest way to do this for the time being. Can the functionality be combined somehow?

src/main/scala/analysis/InterLiveVarsAnalysis.scala
```diff
@@ -0,0 +1,115 @@
...
+ * The result of what variables are alive at each point in the program should still be correct
+ * However, the functions that are callees of other functions will have an over approximation of their parameters
+ * alive at the top of the function
+ * Tip SPA IDE Slides include a short and clear explanation of microfunctions
+ * https://cs.au.dk/~amoeller/spa/8-distributive.pdf
+ */
+trait LiveVarsAnalysisFunctions extends BackwardIDEAnalysis[Variable, FlatElement[Nothing] ,TwoElementLattice] {
+
+  val valuelattice: TwoElementLattice = TwoElementLattice()
+  val edgelattice: EdgeFunctionLattice[FlatElement[Nothing], valuelattice.type] = new EdgeFunctionLattice[FlatElement[Nothing], valuelattice.type](valuelattice)
```
l-kent, line 19: There are a lot of these `FlatElement[Nothing]` and `TwoElementLattice` (which extends `FlatLattice[Nothing] floating around - are they to represent a lattice that only has two values, Top and Bottom? There should be new types created for that instead of awkwardly trying to repurpose `FlatElement` and `FlatLattice`

src/main/scala/analysis/IDEAnalysis.scala
```diff
@@ -0,0 +1,55 @@
...
+ * Base trait for IDE analyses.
+ * @tparam D the type of items
+ * @tparam L the type of the value lattice
+ * Adapted from Tip
+ * https://github.com/cs-au-dk/TIP/blob/master/src/tip/solvers/IDEAnalysis.scala
+ * The special item representing the empty element in IDE.
+ */
+final case class Lambda()
+
+trait IDEAnalysis[E, EE, C, R, D, T, L <: Lattice[T]] {
```
l-kent, line 15: really need comments for all these type parameters

src/main/scala/analysis/solvers/FixPointSolver.scala
```diff
@@ -259,3 +260,30 @@ trait SimplePushDownWorklistFixpointSolver[N, T, L <: Lattice[T]] extends PushDo
     x = lattice.bottom
     run(domain)
     x
+
+trait WorklistFixPointFunctions[N, T, L <: Lattice[T]]  extends  LinkedHashSetWorklist[N]:
```
l-kent, line 264: What is this worklist for & what differentiates it?

src/test/scala/SystemTests.scala
```diff
@@ -83,7 +83,7 @@ class SystemTests extends AnyFunSuite {
     Logger.info(outPath)
     val timer = PerformanceTimer(s"test $name/$variation")
 
-    val args = mutable.ArrayBuffer("--adt", ADTPath, "--relf", RELFPath, "--output", outPath)
+    val args = mutable.ArrayBuffer("--adt", ADTPath, "--relf", RELFPath, "--output", outPath, "--analyse")
```
l-kent, line 86: Do not make this the default for running tests. Make a separate test suite for that.

src/main/scala/analysis/IDEAnalysis.scala
```diff
@@ -0,0 +1,55 @@
...
+
+  /**
+   * Edges for call-to-entry.
+   */
+  def edgesCallToEntry(call: C, entry: E)(d: DL): Map[DL, edgelattice.Element]
+
+  /**
+   * Edges for exit-to-aftercall.
+   */
+  def edgesExitToAfterCall(exit: EE, aftercall: R)(d: DL): Map[DL, edgelattice.Element]
```
l-kent, line 38: This should be named edgesReturnToAfterCall given that we don't use TIP's concept of a 'function exit'?

l-kent:
> This has been done and can be seen in the diff, we are maintaining sbt for now so it mentions both build systems.

Oh, I missed that. Thanks! When will we want to deprecate sbt? 

> cleaning up the tests is not a huge priority, text matching works fine. 

They aren't checking correctness, just changes to the output. Tests should ideally do more than just identify when the output has changed.

> The idea is the direct expr version replaces all the direct clauses in the spec, but both of them are hacks IMO and I don't see the need to refactor the specification to unify them and fix all the tests just for aesthetic reasons.

Then just remove all the existing direct clauses? There are not many tests at all that use those so it shouldn't be hard to change them over?

src/main/scala/analysis/IDEAnalysis.scala
sadrabt: I think edgesExitToAfterCall makes more sense since the exit can be a return (indirect call) or a procedure depending on whether the analysis is forward or not. Exit just refers to the CFGPosition where the analysis exits the function call. 
Similar to edgesCallToEntry where entry can also be a return (indirect call) or a procedure.

src/main/scala/analysis/IntraLiveVarsAnalysis.scala
```diff
@@ -0,0 +1,36 @@
...
+          case Assume(expr, maybeString, maybeString1, bool) => s ++ expr.variables
+          case Assert(expr, maybeString, maybeString1) => s ++ expr.variables
+          case nop: NOP => ???
+          case _ => ???
+      case j: Jump =>
+        j match
+          case call: Call =>
+            call match
+              case IndirectCall(variable, maybeBlock, maybeString) => s + variable
+              case call: DirectCall => s
```
l-kent, line 26: you don't need to match Call before matching IndirectCall and DirectCall

l-kent:
As I said in #189, for the analysis tests, it would be preferable to avoid using the examples folder - if you want to use an existing input then it'd be ideal to use something from the test folder and not the examples folder. If you want to use specific examples that do not exist in the test folder, they should be added to a new test suite.

It would also probably be good to have at least one less trivial test case for these analyses, something with a few procedures with many more variables? This isn't necessary before the branch is merged though.

src/main/scala/analysis/IrreducibleLoops.scala
```diff
@@ -0,0 +1,389 @@
...
+     */
+    def irreducible_loops(): Set[Loop] = {
+        val irreducibleLoops: Set[Loop] = loops.values.filter(l => !l.reducible).toSet;
+
+        if (irreducibleLoops.isEmpty) {
+            return Set[Loop]();
+        }
+
+        val wantedLoops: mutable.Set[Loop] = mutable.Set[Loop]();
+        val irrHeaders: Set[CFGPosition] = irreducibleLoops.map(l => l.header);
```
l-kent, line 86: What are these lines for?

src/main/scala/analysis/IrreducibleLoops.scala
```diff
@@ -0,0 +1,389 @@
...
+
+/* A connection between to IL nodes, purely for the representation of loops in
+ *
+ */
+class LoopEdge(val from: CFGPosition, val to: CFGPosition):
+    override def toString: String = s"(${label(from)}, ${label(to)})"
+    override def equals(n: Any): Boolean = n match {
+        case edge: LoopEdge => edge.from == from && edge.to == to;
+        case _ => false
+    }
```
l-kent, line 33: should be a case class with the equals method removed

l-kent:
The irreducible loop analysis isn't currently called anywhere in the program itself right, just its own test cases? Has it been tested on cases where there are irreducible loops between procedures, not just internally within a procedure?

ailrst:
> Has the irreducible loop analysis been tested on cases where there are irreducible loops between procedures, not just internally within a procedure?

No, and its not expected to work for those cases

l-kent:
Why is it based around CFGPosition and not Block then?

ailrst:
> Why is it based around CFGPosition and not Block then?

Because it was refactored from the old CFGNode and CFGPosition has closest semantics, it can be refactored to be based around block in the future.

l-kent:
The mill scripts should probably be in root, not in /scripts, and the readme should explain to use ./mill (or ./mill.bat for windows) if mill isn't installed? A proper explanation for setting up the IntelliJ integration with mill would also help - simultaneously supporting sbt seems to make it trickier? The .bsp folder shouldn't have been committed either?

src/main/scala/translating/IRToBoogie.scala
```diff
@@ -81,13 +81,20 @@ class IRToBoogie(var program: Program, var spec: Specification) {
...
+      case Some(ProcRelyVersion.Function) =>
+        // if rely/guarantee lib exist, create genRelyInv, and genInv for every procedure where rely/guarantee lib exist
+        if (libRelies.values.flatten.nonEmpty && libGuarantees.values.flatten.nonEmpty) {
+          List(genRelyInv) ++ libGuarantees.flatMap((k, v) => if v.nonEmpty then genInv(k) :+ genLibGuarantee(k) else Nil)
+        } else {
+          List()
+        }
+      case Some(ProcRelyVersion.IfCommandContradiction) => libRGFunsContradictionProof.values.flatten
+      case None => None
     }
```
l-kent, line 95: This should consistently return a List[BProcedure] even if it is an empty list

src/test/scala/ir/IRTest.scala
```diff
@@ -0,0 +1,383 @@
...
+    assert(prev.size == 1 && prev.collect {
+      case c : GoTo => (c.parent == p.blocks("l_main")) && c.isAfterCall
+    }.contains(true))
+
+    assert(next.toSet == Set(p.procs("p1"), p.blocks("l_main").fallthrough.get))
+
+    val prevB : Block = (p.blocks("l_main").jump match
+      case c : IndirectCall  => c.returnTarget
+      case c : DirectCall => c.returnTarget
+    ).get
```
l-kent, line 371: shouldn't use a partially defined match to get the return target

src/test/scala/ir/IRTest.scala
```diff
@@ -0,0 +1,383 @@
...
+
+    assert(prevB.isAfterCall)
+    assert(InterProcIRCursor.pred(prevB).size == 1)
+    assert(InterProcIRCursor.pred(prevB).toSet == Set(p.blocks("l_main").fallthrough.get))
+    assert(InterProcBlockIRCursor.pred(prevB).toSet == Set(p.blocks("l_main"), p.procs("p1").returnBlock.get))
```
l-kent, line 376: These .toSets (and some others) are redundant

src/test/scala/test_util/TestUtil.scala
```diff
@@ -0,0 +1,45 @@
...
+  def getSubdirectories(directoryName: String): Array[String] = {
+    File(directoryName).listFiles.filter(_.isDirectory).map(_.getName)
+  }
+
+
+  def runExample(name: String, path: String = correctPath, variation: String = "gcc/"): BasilResult = {
+    RunUtils.loadAndTranslate(
+      BASILConfig(
+        loading = ILLoadingConfig(
+          adtFile = path + s"/$name/$variation$name.adt",
```
l-kent, line 32: needs to be inputFile now

src/test/scala/ir/IRTest.scala
ailrst: They're defensive in case the IR gets refactored to not be a set anymore

l-kent:
The irreducible control flow removal does not currently produce semantically equivalent results - it currently just merges all edges that lead to loop entrances, leaving out the part of the algorithm where all the new control flow is conditional to ensure semantic equivalence. This is not difficult to fix though.

src/test/scala/ir/IRTest.scala
l-kent: wouldn't it make more sense to use .head (and not convert the rhs to a set) given that the previous line is an assertion that there is only 1 predecessor

src/main/scala/analysis/IntraLiveVarsAnalysis.scala
```diff
@@ -7,29 +7,19 @@ abstract class LivenessAnalysis(program: Program) extends Analysis[Any]:
...
+  def transfer(n: CFGPosition, s: Set[Variable]): Set[Variable] = {
+    n match {
+      case LocalAssign(variable, expr, _) => (s - variable) ++ expr.variables
+      case MemoryAssign(_, store, _) => s ++ store.index.variables ++ store.value.variables
+      case Assume(expr, _, _, _) => s ++ expr.variables
+      case Assert(expr, _, _) => s ++ expr.variables
+      case IndirectCall(variable, _, _) => s + variable
+      case c: DirectCall => s
+      case g: GoTo => s
+      case _ => ???
```
sadrabt, line 19: ~~case c: DirectCall => s~~
~~case g: GoTo => s~~
~~case _ => ???~~
`case _ => s`

sadrabt:
should `build.sbt` include 
libraryDependencies += "io.spray" %%  "spray-json" % "1.3.6",
it's included in the mill build file and you need it if you build with sbt

l-kent:
That is in build.sbt?

sadrabt:
> That is in build.sbt?

yes

readme.md
```diff
@@ -8,127 +8,92 @@ binaries that have been lifted to intermediate formats. Supported input formats
...
 
-##### Compiler Explorer Container
+Installing [mill](https://mill-build.com/mill/Installation_IDE_Support.html) or [sbt](https://www.scala-sbt.org/download.html) and [JDK >= 17](https://openjdk.org/install/) is required.
 
-```sh
-podman pull ghcr.io/uq-pac/basil-compiler-explorer:latest
-```
+This can be done via [coursier](https://get-coursier.io/docs/cli-overview).
 
+A mill script is checked in at `./scripts/mill` so it does not need to be installed.
```
l-kent, line 49: You've changed this now and you should explain to use ./mill or ./mill.bat on Windows

src/main/scala/util/RunUtils.scala
```diff
@@ -393,134 +623,76 @@ object RunUtils {
...
+  }
+
+  /** Use static analysis to resolve indirect calls and replace them in the IR until fixed point.
+    */
+  def staticAnalysis(config: StaticAnalysisConfig, ctx: IRContext): StaticAnalysisContext = {
+    var iteration = 1
+    var modified: Boolean = true
+    var analysisResult = mutable.ArrayBuffer[StaticAnalysisContext]()
+    while (modified) {
+      var newIR = null
```
l-kent, line 670: This is never used? We also really should not use null for anything.

src/main/scala/util/RunUtils.scala
```diff
@@ -393,134 +623,76 @@ object RunUtils {
...
+
+    BasilResult(ctx, analysis, boogieProgram)
+  }
+
+  /** Use static analysis to resolve indirect calls and replace them in the IR until fixed point.
+    */
+  def staticAnalysis(config: StaticAnalysisConfig, ctx: IRContext): StaticAnalysisContext = {
+    var iteration = 1
+    var modified: Boolean = true
+    var analysisResult = mutable.ArrayBuffer[StaticAnalysisContext]()
```
l-kent, line 668: This doesn't need to be a var

src/main/scala/util/RunUtils.scala
```diff
@@ -393,134 +623,76 @@ object RunUtils {
...
+    while (modified) {
+      var newIR = null
+      Logger.info("[!] Running Static Analysis")
+      val result = StaticAnalysis.analyse(ctx, config, iteration)
+      analysisResult.append(result)
+      Logger.info("[!] Replacing Indirect Calls")
+      val (ir, mod) = IRTransform.resolveIndirectCalls(result.cfg, result.vsaResult, ctx.program)
```
l-kent, line 675: ir is never used for anything? Which means further iterations won't actually build upon the resolved indirect calls?

src/main/scala/util/RunUtils.scala
```diff
@@ -34,15 +34,65 @@ import analysis.CfgCommandNode
...
+  }
 
-object RunUtils {
-  var memoryRegionAnalysisResults: Map[CfgNode, LiftedElement[Set[MemoryRegion]]] = Map()
+  /** Load a program from files using the provided configuration.
+    */
+  def load(q: ILLoadingConfig): IRContext = {
+    val (externalFunctions, globals, globalOffsets, mainAddress) = IRLoading.loadReadELF(q.relfFile, q)
+
+    var program: Program = if (q.inputFile.endsWith(".adt")) {
```
l-kent, line 82: This doesn't need to be a var?

src/main/scala/util/RunUtils.scala
```diff
@@ -34,15 +34,65 @@ import analysis.CfgCommandNode
...
+    IRconstPropResult: Map[CFGPosition, Map[Variable, FlatElement[BitVecLiteral]]],
+    memoryRegionResult: Map[CfgNode, LiftedElement[Set[MemoryRegion]]],
+    vsaResult: Map[CfgNode, LiftedElement[Map[Variable | MemoryRegion, Set[Value]]]],
+    interLiveVarsResults: Map[CFGPosition, Map[Variable, TwoElementLatticeEl]],
+    paramResults: Map[Procedure, Set[Variable]]
+) {}
+
+/** Results of the main program execution.
+  */
+case class BasilResult(val ir: IRContext, analysis: Option[StaticAnalysisContext], boogie: BProgram) {}
```
l-kent, line 65: don't need to specify that a case class parameter is a val

examples/paper_example_librg/paper_example_librg.bpl
l-kent, line 1: don't commit .bpl files unless they're necessary like extraspec.bpl

src/main/scala/ir/Program.scala
```diff
@@ -158,17 +158,17 @@ class Program(var procedures: ArrayBuffer[Procedure], var mainProcedure: Procedu
...
-                  var returnBlock: Option[Block],
+                  private var _entryBlock: Option[Block],
+                  private var _returnBlock: Option[Block],
                   private val _blocks: mutable.LinkedHashSet[Block],
                   var in: ArrayBuffer[Parameter],
                   var out: ArrayBuffer[Parameter],
                 ) {
-  private val _callers = mutable.HashSet[DirectCall]()
-  _blocks.foreach(_.setParent(this))
+  private var _callers = mutable.HashSet[DirectCall]()
```
l-kent, line 167: This doesn't need to be a var?

src/main/scala/ir/Program.scala
```diff
@@ -269,51 +344,61 @@ class Parameter(var name: String, var size: Int, var value: Register) {
 }
 
 
-class Block private (var label: String,
+
+class Block private (
+  var label: String,
```
l-kent, line 349: This doesn't need to be a var?

src/test/scala/ParamAnalaysisTests.scala
l-kent, line 1: rename to ParamAnalysisTests.scala

readme.md
```diff
@@ -8,127 +8,95 @@ binaries that have been lifted to intermediate formats. Supported input formats
...
-   - To recompile with the currently checked-out repo run `podman-compose build basil-build`
-   - To enter a shell inside the container
-      - `podman compose run basil-build`
-- `basil` precompiled jar file and tools
-   - To run the jar inside the docker image `podman-compose run basil-dev $arguments...`
-- `compiler-explorer`
-    - instance of [godbolt.org](godbolt.org) with the tools installed. 
 
-Or enter the dev container manually, mounting the current directory (the same as podman-compose basil-dev).
+The tool is a Scala project, and is therefore OS-independent, but is only tested under linux.
```
l-kent, line 41: Not true, I am using Windows for development, with wsl used for all the compiling and lifting of binaries.

readme.md
```diff
@@ -8,127 +8,95 @@ binaries that have been lifted to intermediate formats. Supported input formats
...
-   - To run the jar inside the docker image `podman-compose run basil-dev $arguments...`
-- `compiler-explorer`
-    - instance of [godbolt.org](godbolt.org) with the tools installed. 
 
-Or enter the dev container manually, mounting the current directory (the same as podman-compose basil-dev).
+The tool is a Scala project, and is therefore OS-independent, but is only tested under linux. 
 
 
-#### 2. Use the images manually  
+The protobuffer compiler is required to build the project (due to ddisasm), this can be installed as the apt `protobuf-compiler` package, or from [here](https://protobuf.dev/downloads/).
```
l-kent, line 44: I don't believe this is correct - it is necessary to have ddisasm etc. installed to lift binaries to .gts files, but that isn't related to building the project. I certainly don't have the protobuf compiler installed on Windows, which I use to build and run the project.

The issue may be that the mill build files haven't been updated to include the protobuf scala library as a dependency?

readme.md
```diff
@@ -8,127 +8,95 @@ binaries that have been lifted to intermediate formats. Supported input formats
...
 
-##### Compiler Explorer Container
+Furthermore, lifting input files from a given AArch64 binary is Linux-specific, and all commands given are for Linux. 
+On Windows, WSL2 may be used to run any Linux-specific tasks, but it is less thoroughly tested.
```
l-kent, line 47: 'less thoroughly tested' is not true

src/main/scala/analysis/ContextTransfer.scala
```diff
@@ -0,0 +1,75 @@
+package analysis
+import ir.*
+import util.Logger
+
+import scala.collection.mutable
+
+/** Abandoned helper analysis for intraproc steensgaard to transfer the context of the constant
+ *  propagation analysis to the call sites.
+  */
```
l-kent, line 9: If this is abandoned, do we need to keep it around?

src/main/scala/analysis/MemoryModelMap.scala
```diff
@@ -86,38 +141,74 @@ class MemoryModelMap {
...
 trait MemoryRegion {
   val regionIdentifier: String
   var extent: Option[RangeKey] = None
+  val content: mutable.Set[BitVecLiteral | MemoryRegion] = mutable.Set()
 }
 
-class StackRegion(override val regionIdentifier: String, val start: BitVecLiteral) extends MemoryRegion {
-  override def toString: String = s"Stack($regionIdentifier, $start)"
+class StackRegion(override val regionIdentifier: String, val start: BitVecLiteral, val parent: Procedure = null) extends MemoryRegion {
+  override def toString: String = s"Stack($regionIdentifier, $start, ${if parent != null then parent.name else "Null"}) -> $content"
```
l-kent, line 211: Don't use null

src/main/scala/analysis/SSAForm.scala
```diff
@@ -0,0 +1,103 @@
...
+import analysis.*
+import ir.{SignExtend, *}
+
+import scala.collection.mutable
+
+/** Set-Based SSA
+ * - Each variable has a set of versions
+ * - New assignments create new versions and replaces any new versions
+ */
+case class SSAForm() {
```
l-kent, line 12: Why is this a case class with no parameters?

src/main/scala/ir/Expr.scala
```diff
@@ -350,6 +351,9 @@ case class Memory(name: String, addressSize: Int, valueSize: Int) extends Expr w
 sealed trait Variable extends Expr {
   val name: String
   val irType: IRType
+  val ssa_id: mutable.Set[Int] = mutable.Set[Int]()
+  var sharedVariable: Boolean = false
```
l-kent, line 355: Making the Variable trait mutable is really not a good way to do this. Exprs are supposed to be immutable. This will be very messy and unpredictable.

src/main/scala/ir/Visitor.scala
```diff
@@ -398,6 +398,20 @@ class ExternalRemover(external: Set[String]) extends Visitor {
...
+//  override def visitLocalAssign(node: LocalAssign): Statement =
+//    node.rhs match {
+//      case r: Register =>
+//        regToMemLoad.addOne(())
+//        super.visitLocalAssign(node)
+//      case _ =>
+//        super.visitLocalAssign(node)
+//    }
+//}
+
```
l-kent, line 414: What's this?

src/main/scala/util/RunUtils.scala
```diff
@@ -336,6 +453,8 @@ object IRTransform {
 /** Methods relating to program static analysis.
   */
 object StaticAnalysis {
+  var memoryRegionAnalysisResults: Map[CfgNode, LiftedElement[Set[MemoryRegion]]] = Map()
+  var mmmResults: MemoryModelMap = MemoryModelMap()
```
l-kent, line 457: We don't want to use global variables to access the results, they should be in StaticAnalysisContext instead now

src/main/scala/analysis/MemoryModelMap.scala
```diff
@@ -86,38 +141,74 @@ class MemoryModelMap {
...
+        println(s"  $range -> $region")
+      }
+    }
+    println("Data:")
+    for ((range, region) <- rangeMap.dataMap) {
+      if (region.content.nonEmpty || !hideEmpty) {
+        println(s"  $range -> $region")
+      }
+    }
+  }
```
l-kent, line 201: This should use the logger rather than println

src/test/scala/PointsToTest.scala
l-kent, line 1: This needs to be updated after the RunUtils merge

src/main/scala/analysis/InterprocSteensgaardAnalysis.scala
```diff
@@ -0,0 +1,395 @@
...
+                    X2_star.foreach(
+                      x => unify(exprToStTerm(x), PointerRef(alpha))
+                    )
+                    unify(alpha, varToStTerm(RegisterVariableWrapper(X1)))
+
+                    println("Memory load: " + memoryLoad)
+                    println("Index: " + memoryLoad.index)
+                    println("X2_star: " + X2_star)
+                    println("X1: " + X1)
+                    println("LocalAssign: " + localAssign)
```
l-kent, line 247: use Logger.debug not println

src/main/scala/analysis/MemoryModelMap.scala
```diff
@@ -1,32 +1,49 @@
...
   private val rangeMap = RegionToRangesMap()
   private val MAX_BIGINT: BigInt = BigInt(Long.MaxValue)
   private val contextStack = mutable.Stack.empty[List[StackRegion]]
-  private val allStacks = mutable.Map[String, List[StackRegion]]()
-
-  // Add a range and object to the mapping
-  def add(offset: BigInt, region: MemoryRegion): Unit = {
+  private val sharedContextStack = mutable.Stack.empty[List[StackRegion]]
+  private val localStacks = mutable.Map[String, List[StackRegion]]()
+  private val sharedStacks = mutable.Map[String, List[StackRegion]]().withDefaultValue(List.empty[StackRegion])
```
l-kent, line 31: withDefaultValue produces some confusing behaviour and should be avoided

src/main/scala/analysis/InterprocSteensgaardAnalysis.scala
```diff
@@ -0,0 +1,395 @@
...
+          case localAssign: LocalAssign =>
+            localAssign.rhs match {
+              case binOp: BinaryExpr =>
+                // X1 = &X2: [[X1]] = ↑[[X2]]
+                exprToRegion(binOp, cmd).foreach(
+                  x => unify(varToStTerm(RegisterVariableWrapper(localAssign.lhs)), PointerRef(allocToTerm(x)))
+                )
+              // TODO: should lookout for global base + offset case as well
+              case _ =>
+                unwrapExpr(localAssign.rhs).foreach {
```
l-kent, line 232: this should have a `case _ =>` too since the cases are not exhaustive

src/main/scala/analysis/UtilMethods.scala
```diff
@@ -54,3 +56,85 @@ def evaluateExpression(exp: Expr, constantPropResult: Map[Variable, FlatElement[
...
+    case e: Extract =>
+      evaluateExpressionWithSSA(e.body, constantPropResult) match {
+        case b: Set[BitVecLiteral] if b.nonEmpty => applySingle(BitVectorEval.boogie_extract(e.end, e.start, _: BitVecLiteral), b)
+        case _               => Set()
+      }
+    case registerVariableWrapper: RegisterVariableWrapper =>
+      constantPropResult.collect({
+        case (k, v) if k == registerVariableWrapper => v
+      }).flatten.toSet
+      constantPropResult(registerVariableWrapper)
```
l-kent, line 110: This is an unreachable case because you are matching on an Expr which RegisterVariableWrapper is not a subclass of

l-kent:
@yousifpatti plenty of feedback for you

src/main/scala/ir/Expr.scala
l-kent: Perhaps a new SSAVariable case class that contains a Variable and an Int would be a better way to implement SSA?

l-kent: This also needs commenting - what does it mean for a variable to have multiple SSA ids?

src/main/scala/analysis/InterprocSteensgaardAnalysis.scala
```diff
@@ -0,0 +1,395 @@
...
+      t1,
+      t2
+    ) // note that unification cannot fail, because there is only one kind of term constructor and no constants
+  }
+
+  type PointsToGraph = Map[Object, Set[VariableWrapper | MemoryRegion]]
+
+  /** @inheritdoc
+    */
+  def pointsTo(): PointsToGraph = {
```
l-kent, line 305: Why does this need its own type definition when this is the only use? Why does it not just correctly have CfgNode instead of Object?

src/main/scala/analysis/MemoryModelMap.scala
```diff
@@ -1,32 +1,49 @@
...
 
 // Define a case class to represent a range
-case class RangeKey(var start: BigInt, var end: BigInt):
+case class RangeKey(var start: BigInt, var end: BigInt) extends Ordered[RangeKey]:
   def size(): BigInt = end - start
+  override def compare(that: RangeKey): Int = {
+    if (start < that.start) -1
+    else if (start > that.start) 1
+    else 0
+  }
```
l-kent, line 15: case class should not contain vars

src/main/scala/analysis/MemoryModelMap.scala
```diff
@@ -1,32 +1,49 @@
...
 
 case class RegionToRangesMap():
-  val stackMap: mutable.Map[RangeKey, StackRegion] = mutable.Map()
-  val heapMap: mutable.Map[RangeKey, HeapRegion] = mutable.Map()
-  val dataMap: mutable.Map[RangeKey, DataRegion] = mutable.Map()
+  val stackMap: mutable.Map[RangeKey, StackRegion] = mutable.TreeMap()
+  val sharedStackMap: mutable.Map[Procedure, mutable.TreeMap[RangeKey, StackRegion]] = mutable.Map[Procedure, mutable.TreeMap[RangeKey, StackRegion]]()
+  val heapMap: mutable.Map[RangeKey, HeapRegion] = mutable.TreeMap()
+  val dataMap: mutable.Map[RangeKey, DataRegion] = mutable.TreeMap()
```
l-kent, line 22: Why is this a case class and why is it even separate from the MemoryModelMap class? these could all just be internal variables to MemoryModelMap?

l-kent:
I tried to fix the PointsToTest but the examples are malformed in some way compared to what the IDESolver expects and since the test just runs all the analyses it gets caught up in that. We should have some way of only running particular analyses but also make it so that the DSL cannot so easily produce malformed examples, or fix whatever the issue in the IDESolver is if it's at fault.

sadrabt:
They just need to run the return unifier on the test program which set the return blocks for procedures 
the return block is how we get function exit points
val returnUnifier = ConvertToSingleProcedureReturn()
program = returnUnifier.visitProgram(program)

src/main/scala/analysis/InterprocSteensgaardAnalysis.scala
```diff
@@ -0,0 +1,395 @@
...
+import analysis.solvers.{Cons, Term, UnionFindSolver, Var}
+import ir.*
+import util.Logger
+import scala.collection.mutable
+import scala.collection.mutable.ListBuffer
+
+/** Wrapper for variables so we can have Steensgaard-specific equals method indirectly */
+sealed trait VariableWrapper {
+  val variable: Variable
+}
```
l-kent, line 12: why does this trait exist when there is only one implementation of it?

src/main/scala/analysis/InterprocSteensgaardAnalysis.scala
```diff
@@ -0,0 +1,395 @@
...
+            variable == other && variable.ssa_id.intersect(other.ssa_id).nonEmpty
+        case _ =>
+            false
+        }
+    }
+
+    override def hashCode(): Int = {
+        variable.hashCode()
+    }
+}
```
l-kent, line 28: This is a very messy way to handle SSA. If we are using this wrapper, why did mutable state need to be added to the Variable trait? Don't override the hashcode for a case class.

src/main/scala/analysis/UtilMethods.scala
```diff
@@ -54,3 +56,85 @@ def evaluateExpression(exp: Expr, constantPropResult: Map[Variable, FlatElement[
...
+            case BVSHL => apply(BitVectorEval.smt_bvshl, lhs, rhs)
+            case BVOR => apply(BitVectorEval.smt_bvor, lhs, rhs)
+            case _ => throw new RuntimeException("Binary operation support not implemented: " + binOp.op)
+          }
+      }
+    case extend: ZeroExtend =>
+      evaluateExpressionWithSSA(extend.body, constantPropResult) match {
+        case b: Set[BitVecLiteral] if b.nonEmpty => applySingle(BitVectorEval.smt_zero_extend(extend.extension, _: BitVecLiteral), b)
+        case _                => Set()
+      }
```
l-kent, line 100: Why does this need to check for emptiness, shouldn't applySingle return an empty set if given an empty set as input?

src/main/scala/analysis/UtilMethods.scala
```diff
@@ -54,3 +56,85 @@ def evaluateExpression(exp: Expr, constantPropResult: Map[Variable, FlatElement[
...
+    case registerVariableWrapper: RegisterVariableWrapper =>
+      constantPropResult.collect({
+        case (k, v) if k == registerVariableWrapper => v
+      }).flatten.toSet
+      constantPropResult(registerVariableWrapper)
+    case variable: Variable =>
+      val registerVariableWrapper = RegisterVariableWrapper(variable)
+      constantPropResult.collect({
+        case (k, v) if k == registerVariableWrapper => v
+      }).flatten.toSet
```
l-kent, line 115: why does this need to use collect instead of just directly accessing the map? Is this a quirk of how the SSA is implemented - registerVariableWrapper equality is only checking if the variables share any SSA IDs? this defeats the whole purpose of a Map and is very unclear

src/main/scala/analysis/InterprocSteensgaardAnalysis.scala
```diff
@@ -0,0 +1,395 @@
...
+    * @param node
+    *   the node for which it generates the constraints
+    * @param arg
+    *   unused for this visitor
+    */
+  def visit(n: CfgNode, arg: Unit): Unit = {
+
+    def varToStTerm(vari: VariableWrapper): Term[StTerm] = IdentifierVariable(vari)
+    def exprToStTerm(expr: MemoryRegion | Expr): Term[StTerm] = ExpressionVariable(expr)
+    def allocToTerm(alloc: MemoryRegion): Term[StTerm] = AllocVariable(alloc)
```
l-kent, line 211: Why do these exist when the constructors for those case classes could be called each time instead?

src/main/scala/analysis/SSAForm.scala
l-kent, line 1: This approach makes no attempt to handle loops and so note should be taken of that

src/main/scala/analysis/SSAForm.scala
```diff
@@ -0,0 +1,99 @@
...
+                val maxVal = varMaxTracker.getOrElseUpdate(localAssign.lhs.name, 0)
+                blockBasedMappings((currentBlock, localAssign.lhs.name)) = mutable.Set(maxVal)
+
+                localAssign.lhs.ssa_id.clear()
+                localAssign.lhs.ssa_id.addAll(blockBasedMappings((currentBlock, localAssign.lhs.name)))
+
+                varMaxTracker(localAssign.lhs.name) = blockBasedMappings((currentBlock, localAssign.lhs.name)).max + 1
+              }
+              case memoryAssign: MemoryAssign => {
+                transformVariables(memoryAssign.lhs.variables, currentBlock, proc)
```
l-kent, line 65: memoryAssign has no Variables in its lhs

src/main/scala/analysis/SSAForm.scala
```diff
@@ -0,0 +1,99 @@
...
+
+/** Set-Based SSA
+ * - Each variable has a set of versions
+ * - New assignments create new versions and replaces any new versions
+ */
+case class SSAForm() {
+
+  val varMaxTracker = new mutable.HashMap[String, Int]()
+  private val blockBasedMappings = new mutable.HashMap[(Block, String), mutable.Set[Int]]().withDefault(_ => mutable.Set())
+  private val context = new mutable.HashMap[(Procedure, String), mutable.Set[Int]]().withDefault(_ => mutable.Set())
```
l-kent, line 16: please avoid using withDefault, it can have confusing behaviour

src/main/scala/analysis/SSAForm.scala
```diff
@@ -0,0 +1,99 @@
...
+              case _ => throw new RuntimeException("No SSA form for " + stmt.getClass + " yet")
+            }
+          }
+          currentBlock.jump match {
+            case directCall: DirectCall =>
+              // TODO: transfers the whole context but it could be using ANR and RNA to transfer only the relevant context
+              varMaxTracker.keys.foreach(varr => {
+                //context((directCall.target, varr)) = context((directCall.target, varr)) ++ blockBasedMappings(block, varr)
+                context.getOrElseUpdate((directCall.target, varr), mutable.Set()) ++= blockBasedMappings((currentBlock, varr))
+              })
```
l-kent, line 83: This is dependent on visiting the procedures in the correct order. If a procedure is iterated over before all the procedures that call it, then its SSA values will be incorrect.

src/main/scala/analysis/SSAForm.scala
```diff
@@ -0,0 +1,99 @@
...
+            case indirectCall: IndirectCall =>
+              transformVariables(indirectCall.target.variables, currentBlock, proc)
+            case goTo: GoTo =>
+              goTo.targets.foreach(b => {
+                varMaxTracker.keys.foreach(varr => {
+                  blockBasedMappings((b, varr)) = blockBasedMappings(b, varr) ++ blockBasedMappings(currentBlock, varr)
+                })
+              })
```
l-kent, line 91: This is dependent on visiting blocks after all of their predecessors, which no attempt is made to guarantee. If a block is visited before all of its predecessors, then its SSA values will be incorrect.

For example, if we have the following blocks:

A goto B, C
C goto B

There are two possible orders the blocks could be visited in (as the next blocks are stored in a set) - either A, B, C or A, C, B. If they are visited in the order A, B, C, then the visit to B will incorrectly miss using C's SSA values.

examples/paper_example_librg/extraspec.bpl
```diff
@@ -0,0 +1,26 @@
...
+var Gamma_malloc_count: bool;
+var Gamma_malloc_base: bool;
+var Gamma_malloc_end: bool;
+var Gamma_malloc_id: bool;
+
+
+// because named wrong in generated file
+function {:inline} in_bounds64(base: bv64, len: bv64, i: bv64) returns (bool) {
+  (if bvule64(base, bvadd64(base, len)) then (bvule64(base, i) && bvult64(i, bvadd64(base, len))) else (bvule64(base, i) || bvult64(i, bvadd64(base, len))))
+}
```
l-kent, line 22: Is this 'named wrong in generated file' thing still a problem?

yousifpatti:
I am doing clean up based on your comments as well Liam

src/main/scala/analysis/RegToMemAnalysis.scala
```diff
@@ -0,0 +1,67 @@
...
+  def eval(cmd: Command, constants:  Map[Variable, FlatElement[BitVecLiteral]], s: Map[RegisterVariableWrapper, FlatElement[Expr]]): Map[RegisterVariableWrapper, FlatElement[Expr]] = {
+    cmd match {
+      case localAssign: LocalAssign =>
+        localAssign.rhs match {
+          case memoryLoad: MemoryLoad =>
+            s + (RegisterVariableWrapper(localAssign.lhs) -> FlatEl(memoryLoad).asInstanceOf[FlatElement[Expr]])
+          case binaryExpr: BinaryExpr =>
+            if (evaluateExpression(binaryExpr.arg1, constants).isEmpty) { // approximates Base + Offset
+              return s + (RegisterVariableWrapper(localAssign.lhs) -> FlatEl(binaryExpr).asInstanceOf[FlatElement[Expr]])
+            }
```
l-kent, line 38: How does this approximate 'base + offset'? evaluateExpression can return None for plenty of other cases that aren't just aren't handled at present, which will just cause cascading problems. This isn't robust at all.

Why do we even care about the concept of 'base + offset' - that is a heuristic which does not apply to ARM64.

src/main/scala/analysis/RegToMemAnalysis.scala
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * Collects all the memory loads and the expressions that are assigned to a register but cannot be evaluated.
+ *
+ * Tracks:
+ * R_x = MemoryLoad[Base + Offset]
+ * R_x = Base + Offset
+ *
+ * Both in which constant propagation mark as TOP which is not useful.
+ */
+trait RegionAccessesAnalysis(cfg: ProgramCfg, constantProp: Map[CfgNode, Map[Variable, FlatElement[BitVecLiteral]]]) {
```
l-kent, line 17: If the issue is with constant propagation not providing useful results, maybe an analysis that can handle both constants and memory accesses should have been implemented instead of needing multiple separate analyses that are later combined?

src/main/scala/analysis/InterprocSteensgaardAnalysis.scala
```diff
@@ -0,0 +1,381 @@
...
+      val pt = unifications(solution(v))
+        .collect({
+          case PointerRef(IdentifierVariable(id)) => id
+          case PointerRef(AllocVariable(alloc))   => alloc
+        })
+        .toSet
+      a + (v.id -> pt)
+    }
+    Logger.debug(s"\nPoints-to:\n${pointsto.map(p => s"${p._1} -> { ${p._2.mkString(",")} }").mkString("\n")}\n")
+    pointsto.asInstanceOf[Map[CfgNode, Set[RegisterVariableWrapper | MemoryRegion]]]
```
l-kent, line 312: pointsto is never actually a Map[CfgNode, _] it is a Map[RegisterVariableWrapper, _] but is incorrectly. Scala doesn't enforce these sorts of collection parameter types at runtime so it all goes through but is incredibly confusing to try to understand.

src/main/scala/analysis/InterprocSteensgaardAnalysis.scala
```diff
@@ -0,0 +1,381 @@
...
+
+  /** @inheritdoc
+    */
+  def mayAlias(): (RegisterVariableWrapper, RegisterVariableWrapper) => Boolean = {
+    val solution = solver.solution()
+    (id1: RegisterVariableWrapper, id2: RegisterVariableWrapper) =>
+      val sol1 = solution(IdentifierVariable(id1))
+      val sol2 = solution(IdentifierVariable(id2))
+      sol1 == sol2 && sol1.isInstanceOf[PointerRef] // same equivalence class, and it contains a reference
+  }
```
l-kent, line 323: Is this actually used anywhere?

src/main/scala/analysis/IDEAnalysis.scala
```diff
@@ -0,0 +1,61 @@
...
+   * Edges for other CFG nodes.
+   */
+  def edgesOther(n: CFGPosition)(d: DL): Map[DL, edgelattice.Element]
+}
+
+
+trait ForwardIDEAnalysis[D, T, L <: Lattice[T]] extends IDEAnalysis[Procedure, Command, DirectCall, GoTo, D, T, L]
+
+
+trait BackwardIDEAnalysis[D, T, L <: Lattice[T]] extends IDEAnalysis[Command, Procedure, GoTo, DirectCall, D, T, L]
```
l-kent, line 61: Shouldn't we be able to have stricter type constraints on these? The exit for forward analysis/entrance for backward analysis should be a return shouldn't it? (which is for now an IndirectCall)

@sadrabt

src/main/scala/analysis/UtilMethods.scala
```diff
@@ -54,3 +55,76 @@ def evaluateExpression(exp: Expr, constantPropResult: Map[Variable, FlatElement[
...
+    case binaryExpr: BinaryExpr =>
+      unwrapExpr(binaryExpr.arg1)
+      unwrapExpr(binaryExpr.arg2)
+    case memoryLoad: MemoryLoad =>
+      buffers.addOne(memoryLoad)
+      unwrapExpr(memoryLoad.index)
+    case _ =>
+  }
+  buffers
+}
```
l-kent, line 130: What is this supposed to do? My impression is that it will either return a ListBuffer containing a MemoryLoad if the input Expr is a MemoryLoad, or an empty ListBuffer if Expr is not a MemoryLoad, and the recursive calls will do nothing?

src/main/scala/analysis/UtilMethods.scala
l-kent: the result here isn't even used for anything? what does this do exactly? @yousifpatti

src/main/scala/analysis/IDEAnalysis.scala
sadrabt: Most likely. I never came across examples where the exit for forward analysis/entrance for backward analysis is not a return/IndirectCall from our examples.

this is not the case in stripped/uncalled functions so I'm not sure if there is a program that will have something else there
Screenshot from 2024-02-29 13-51-39

Either way, the analysis won't cope well if a function doesn't have a returnblock like above, so the type should be tighter 
(indirectCall/ new return node)

src/main/scala/analysis/InterprocSteensgaardAnalysis.scala
l-kent: @yousifpatti

src/main/scala/analysis/UtilMethods.scala
l-kent: @yousifpatti

src/main/scala/analysis/RegToMemAnalysis.scala
l-kent: @yousifpatti

examples/paper_example_librg/extraspec.bpl
l-kent: @ailrst

src/main/scala/analysis/IrreducibleLoops.scala
l-kent: @ailrst do you have any idea what the idea was here?

src/main/scala/util/RunUtils.scala
l-kent: @ailrst

readme.md
l-kent: @ailrst

readme.md
l-kent: @ailrst

src/main/scala/util/RunUtils.scala
l-kent: @ailrst

readme.md
l-kent: @ailrst

l-kent:
@yousifpatti The indirect call resolution still seems quite unreliable. It fails for the following cases.

examples:
- functionpointer/functionpointer - throws exception due to a region having bitvector as its contents
- functionpointer/functionpointer_clang - throws exception due to a region having bitvector as its contents
- jumptable/jumptable_clang - throws exception due to a region having bitvector as its contents
- jumptable2/jumptable2_clang - fails to resolve statements 000004fc (a call to sub_seven) and 000004e6 (a call to add_six)

tests:
- functionpointer/gcc and functionpointer/clang are the same as the examples and throw an exception
- indirect_call/clang - throws exception due to a region having bitvector as its contents
- indirect_call/gcc - throws exception due to a region having bitvector as its contents
- jumptable/clang - throws exception due to a region having bitvector as its contents
- jumptable2/jumptable2_clang - throws exception due to a region having bitvector as its contents

src/main/scala/analysis/ContextTransfer.scala
l-kent: @yousifpatti

src/main/scala/analysis/UtilMethods.scala
yousifpatti: This is supposed to be recursively called and returns a List containing MemoryLoads identified in the expression as it may contain multiple MemoryLoads
I will fix this

l-kent: It is safe to assume that an Expr does not contain multiple MemoryLoads recursively

l-kent: but you also need to actually apply the semantics of anything the MemoryLoad is contained within

yousifpatti: MRA is only concerned about collecting the regions being loaded and not the actual semantics of the expression as it is not evaluated as a whole. Collecting those regions will provide a more fine-grained MemoryModelMap.
The IR language allows for a binary expression to contain multiple MemoryLoads for examples so you cannot assume that multiple loads will not exist in a single expression

src/main/scala/analysis/InterprocSteensgaardAnalysis.scala
yousifpatti: No, but this function is a result of the analysis because steensgaard is a may-alias analysis originally

yousifpatti: do you suggest removing it?

l-kent: If there's a good chance it will be relevant in the future it makes sense to keep it

src/main/scala/analysis/RegToMemAnalysis.scala
yousifpatti: Constant propagation is meant for tracking constants so it is not proper to mix types with expressions

A new analysis that can handle both may not be necessarily useful because not all subsequent analyses make use of the expression result that the analysis will provide, may lead to confusion on the result of the analysis and in some instances approximating the result to Top could be more beneficial that returning unevaluated expression.

If you think merging it could be better then I can work on that

src/main/scala/analysis/InterprocSteensgaardAnalysis.scala
yousifpatti: I believe it will be relevant in future for tracking regions being passed as parameters

src/main/scala/analysis/RegToMemAnalysis.scala
l-kent: I don't think merging it should be a priority but that sort of thing is something we should think about - considering how do all the analyses we have fit together since it's a bit convoluted at present

src/main/scala/analysis/RegToMemAnalysis.scala
yousifpatti: The heuristic of a base + offset is essential for the working out of regions for the MemoryModelMap. While it may not work perfectly for adrp instructions, it still provide a good approximation of the memory layout. What is an alternative approach that you suggest that does not depend on base + offset for stack regions?

The way the approximation happens is by assuming that the base pointer (ie. lhs in a binaryExpr) is an abstract location that cannot be evaluated and as such if it does not evaluate to anything at that point of the program then it may be a base pointer and the lhs is an offset.

l-kent: "can't be evaluated" is currently defined very broadly since evaluateExpression doesn't support many operators. It also isn't even checking that it is an addition or subtraction operation?

src/main/scala/analysis/UtilMethods.scala
yousifpatti: I agree. An alternative lattice that does not use a map for constantPropResult needs to be used to avoid the problem of adding two identical registers which should work as a register tree structure. I will have to rewrite that part of the constantprop

src/main/scala/analysis/RegToMemAnalysis.scala
l-kent: adrp instructions are also the most common way that pointers to the data section are loaded from memory, so it isn't good if we don't handle those well.

src/main/scala/analysis/InterprocSteensgaardAnalysis.scala
```diff
@@ -0,0 +1,389 @@
...
+          }
+        }
+        evaluateExpressionWithSSA(binExpr, constantProp(n)).foreach { b =>
+          val region = mmm.findDataObject(b.value)
+          reducedRegions = reducedRegions ++ region
+        }
+      case _ =>
+    }
+    reducedRegions
+  }
```
l-kent, line 127: This seems to just assume that the BinaryExpr is an addition operation?

src/main/scala/analysis/RegToMemAnalysis.scala
yousifpatti: Will have to add it as an issue for it then [here](https://github.com/UQ-PAC/bil-to-boogie-translator/issues/192)

src/main/scala/analysis/InterprocSteensgaardAnalysis.scala
yousifpatti: It could be replaced with `binExpr.op` I can change that

src/main/scala/analysis/UtilMethods.scala
yousifpatti: #193

readme.md
l-kent: Why did I only get all these comments now, but it says you made them days ago?

The issue might have just been that you didn't reload sbt after switching branches? Regardless, this is a dependency that we should ensure is addressed through the build system dependency management systems and not recommending people manually install protobuf-compiler.

src/main/scala/analysis/MemoryModelMap.scala
l-kent: It is easy to do weird things with it like mutate the default value instead of mutating the value for a key that hadn't been added, and some operations just confusingly ignore the default.

The alternative is to just check if a key exists in the map before attempting to update its value. This may be a little more verbose but it is worth it.

src/main/scala/analysis/RegToMemAnalysis.scala
l-kent: > We don't need to be so pessimistic, it still seems relevant in a lot of [cases](https://nationally-usable-bull.ngrok-free.app/z/KE7Y8f), and makes sense as the way to symbolically treat addresses. Yes its not a heuristic we can naively apply everywhere and we should be more structured about it. But, e.g. `adrp` is not a huge deal because we usually only want the final address value anyway to find a global variable or block address.
> 
> It would be good to more thoroughly look into what heuristics we can use and where, if you have suggestions on how to infer structure from access patterns please share them. Ultimately these types of analysis (VSA/DSA/ASI) need heuristics like this to identify how to group memory accesses into regions and the ARM load/store addressing mode specification should give us something to work with.

What I would suggest is just making better use of the information in the symbol table - that already gives us what the MemoryRegionAnalysis is attempting to infer. The symbol table directly gives us the addresses of objects in the data region. These objects can then be divided up further according to accesses that take place within them.

src/main/scala/analysis/UtilMethods.scala
l-kent: Part of the issue is that the result of this messy map access isn't even used, is that supposed to be the case

src/main/scala/analysis/RegToMemAnalysis.scala
yousifpatti: Currently, the Data Regions are loaded straight from the symbol table and divided based on their addresses. However, we do not have information about stack regions that we can acquire from an external source so the heuristic is an ok approximation for those regions for our purposes

src/main/scala/analysis/UtilMethods.scala
yousifpatti: That is true. The liner under it should have been deleted and the collected set should be the set being returned. I will fix that.

src/main/scala/analysis/RegToMemAnalysis.scala
l-kent: Not all the information that could be retrieved from the symbol table is retrieved currently, so there is definitely room for improvement there.

It isn't really a heuristic for the stack accesses - it is more just a case of ensuring that the value of the stack pointer is tracked effectively so that offsets to it are evaluated correctly. This part is not really an issue.

examples/paper_example_librg/extraspec.bpl
ailrst: I haven't fixed it so I assume so

readme.md
ailrst: I wrote them three days ago but I didn't realise you can't see them until I submit it as a review

src/main/scala/ir/Expr.scala
ailrst: It's supposed to work like a definitions set, the id is incremented and assigned at every definition, and after a join the variable will contain the ids of the last definition in both paths. Hence the intersection of the set between a variable at different points being nonempty tells you they use the same definition, similar to a phi node.  I think at this point it would be simpler and more reliable to implement a regular usedef analysis with the tip framework with all the state stored externally in a map lattice (cfgposition -> variable -> definition set). We can then optionally use that to transform the IR into an SSA form; maybe wrapping variables in an SSAVariable case class as you suggested; that way its immutable. @yousifpatti

src/main/scala/analysis/IrreducibleLoops.scala
ailrst: Don't know, as far as I can see they aren't used?

src/main/scala/util/RunUtils.scala
ailrst: The IR is modified in-place, the top-level 'program' is not copied or replaced in any way.

src/main/scala/util/RunUtils.scala
ailrst: This function isn't a recursion anymore, this is incremented in a while loop so has to be var afaik?

src/main/scala/analysis/UtilMethods.scala
yousifpatti: Converted into a direct map access in the latest commit

src/main/scala/analysis/ContextTransfer.scala
yousifpatti: It can be removed. I will remove it.

examples/paper_example_librg/extraspec.bpl
l-kent: This is related to the --boogie-use-lambda-stores flag so if it is still a problem then you should file an issue

src/main/scala/util/RunUtils.scala
l-kent: the analysisResult doesn't need to be a var is what I wanted to highlight

src/main/scala/analysis/ContextTransfer.scala
l-kent: @yousifpatti What needs to happen here, just the whole file is fine to remove since we don't need it?

src/main/scala/ir/Statement.scala
```diff
@@ -128,7 +130,40 @@ object GoTo:
 
 sealed trait Call extends Jump
 
-class DirectCall(val target: Procedure, var returnTarget: Option[Block],  override val label: Option[String] = None) extends Call {
+abstract trait FallThrough extends HasParent[Block]:
```
l-kent, line 133: why does this need to be separate from Call? `abstract trait` also is not really anything, it isn't any different to `trait`. @ailrst

l-kent:
I would like to merge this today if possible. Before that, I would like to have ContextTransfer sorted out (removed completely? @yousifpatti) and discuss Fallthrough, but those should be doable today hopefully.

The main outstanding issue is the SSA implementation but I am fine to merge this before that is re-done.

yousifpatti:
ContextTransfer can be removed completely. I had it removed in some branch but haven't merged. I am happy for this PR to be merged as the outstanding SSA rewrite has an issue ticket

src/main/scala/ir/Statement.scala
ailrst: Its a clear single piece of functionality, the invariant (parent.fallthrough == this.returntarget, if and when it is set), its separate so its responsibility is clearer. You could also conceivably have a Call that cannot have a fallthrough, e.g nonterminating depending on how we represent that

l-kent: Do you mean non-returning calls? They aren't currently separate in the IR and if they were then Fallthrough shouldn't have an Option at all. I don't see the point in keeping things separate for hypothetical future refactors that we have no idea if we even will want to do, and Call does not have any functionality associated with it at present anyway.

ailrst: I wrote it that way because I thought the code was clearer, if you disagree you can change it however you like.

l-kent:
@ailrst I think this is fine to merge

PR #189, opened 2024-02-25
Liveness/Parameter Analysis updates

sadrabt:
Addresses review comments for #179 

- Creates common class `PushDownWorklistFixPointFunctions` for propagation worklist solvers that doesn't extend type restricted `Dependencies`
- `TwoElementLattice` inherits directly from `Lattice`
- Tests now use `BasilResult`. Removed liveness and parameter analysis result dump
- Improved documentation for `IDEAnalysis`


src/main/scala/analysis/Lattice.scala
```diff
@@ -71,6 +71,24 @@ class LiftLattice[T, +L <: Lattice[T]](val sublattice: L) extends Lattice[Lifted
...
+case object TwoElementTop extends TwoElementLatticeEl
+case object TwoElementBottom extends TwoElementLatticeEl
+
+
+/**
+ * A lattice with only top and bottom
+ */
+class TwoElementLattice extends Lattice[TwoElementLatticeEl]:
+  override val bottom: TwoElementLatticeEl = TwoElementBottom
+  override def top: TwoElementLatticeEl = TwoElementTop
```
l-kent, line 85: This should be a val, not a def

src/main/scala/analysis/solvers/FixPointSolver.scala
```diff
@@ -261,25 +274,21 @@ trait SimplePushDownWorklistFixpointSolver[N, T, L <: Lattice[T]] extends PushDo
...
-  var x: Map[N, T]
+/**
+ * Push down worklist solver that maps first to init
+ *
+ * @tparam N
+ *   type of the elements in the worklist.
+ * @tparam T Type of the elements in the sublattice
+ * @tparam L Sublattice
+ */
+trait InitializingPushDownWorklistFixpointSolver[N, T, L <: Lattice[T]] extends PushDownWorklistFixPointFunctions[N, T, L]:
```
l-kent, line 286: Why doesn't this just extend PushDownWorklistFixPointSolver? This doesn't make sense on its own as it doesn't have a process method. I understand that it is only used by the IDESolver classes which implement their own process methods.

src/main/scala/analysis/solvers/FixPointSolver.scala
```diff
@@ -204,25 +204,26 @@ trait WorklistFixpointSolverWithReachability[N, T, L <: Lattice[T]] extends Work
...
-trait PushDownWorklistFixpointSolver[N, T, L <: Lattice[T]] extends MapLatticeSolver[N, T, L] with LinkedHashSetWorklist[N] with Dependencies[N]:
+/**
+ * Common trait for pushdown worklist solvers
+ *
+ * @tparam N
+ *   type of the elements in the worklist.
+ * @tparam T Type of the elements in the sublattice
+ * @tparam L Sublattice
+ */
+trait PushDownWorklistFixPointFunctions[N, T, L <: Lattice[T]] extends LinkedHashSetWorklist[N]:
```
l-kent, line 215: Why does this need to exist at all? It just seems to make the inheritance more convoluted without adding anything at present. Just because TIP has very convoluted inheritance doesn't mean we should copy that.

l-kent:
It would be ideal to use tests from the test folder and not the examples folder. If you want to use specific examples that do not exist in the test folder, they should be added to a new test suite. It would also probably be good to have at least one less trivial test case for these analyses, something with a few procedures with many more variables?

src/main/scala/analysis/solvers/FixPointSolver.scala
sadrabt: PushDownWorklistFixPointSolver extends the `Dependencies` trait with a type restriction N where N is the elements in the worklist. The worklist elements for IDE analysis is a triple/tuple like (CFGPosition, DL, DL)/(CFGPosition, DL). I could write a forward and backward iterator that detuples first and calls the IL iterator. PushDownWorklistFixPointSolver also extends MapLatticeSolver, but the IDESolver phases that use the worklist don't require MapLatticeSolver members

l-kent: ah, the TIP inheritance is such a nightmare. I would like to try to organise this better but it'll take a bit more thought and isn't an urgent issue, so it's fine the way it is for now

l-kent:
Since this doesn't seem to conflict with anything else in the staging branch it should be fine to just merge it into that unless @ailrst objects?

Issue #200, opened 2024-03-15
Block.jump_= violates invariant


l-kent:
https://github.com/UQ-PAC/bil-to-boogie-translator/blob/b6a2409f60b977985ee0c97959cdcfe04bdb35f0/src/main/scala/ir/Program.scala#L380-L386

`Block.jump_=` deparents the existing Jump and updates the parent for the replacement Jump. This breaks things if a Jump that is already contained within another Block is the input Jump, as the reference to it in its previous Block will not be removed. Then, if the previous Block's Jump is updated, its old Jump (which already has a different parent) will have its parent set to None which further breaks things.

Either `Jump.deParent()` should replace a Block's jump with some dummy Jump (maybe just None?) or `Block.jump_=` should prevent a Jump contained within an existing Block being used as input, though that would probably require a way to safely decouple a Jump from its parent Block too?

`replaceJump` is also redundant given that `jump_=` exists?

l-kent:
fixed in #208

Issue #199, opened 2024-03-14
Call fallthrough assertion


l-kent:
The Call trait requiring that a returnTarget Block must already have a parent Procedure is impractical.

https://github.com/UQ-PAC/bil-to-boogie-translator/blob/b6a2409f60b977985ee0c97959cdcfe04bdb35f0/src/main/scala/ir/Statement.scala#L133-L142

This is an issue for handleUnidentifedIndirectCall:
https://github.com/UQ-PAC/bil-to-boogie-translator/blob/b6a2409f60b977985ee0c97959cdcfe04bdb35f0/src/main/scala/translating/GTIRBToIR.scala#L352-L398

https://github.com/UQ-PAC/bil-to-boogie-translator/blob/b6a2409f60b977985ee0c97959cdcfe04bdb35f0/src/main/scala/translating/GTIRBToIR.scala#L316-L332

as it means that handleUnidentifiedIndirectCall cannot cleanly return an ArrayBuffer of newly created Blocks which are then added to a Procedure and possibly queued in the current iterator. Instead, all this functionality has to be jumbled up - handleUnidentifiedIndirectCall has to then add the Blocks to the Procedure itself but it still has to either return the newly created Blocks separately or queue them itself too. Either approach is messier than what we currently have.

Additionally, are we even going to want to reassign a returnTarget, as this was not previously allowed? This does not seem too likely in the first place - I can only really see the case where a returnTarget is removed completely because the call has been identified as non-returning?

l-kent:
fixed in #208

PR #208, opened 2024-05-06
IntrusiveList/Parent/Block assertion fixes

### LINKED ISSUE #198: IntrusiveList interface issues
The IntrusiveList is very difficult to work with due to the complex set of assertions designed to ensure that every element is only in a single IntrusiveList. The way this is implemented now (as of the recent staging pull request being merged in) conflicts with the design of the splitOn method:
https://github.com/UQ-PAC/bil-to-boogie-translator/blob/b6a2409f60b977985ee0c97959cdcfe04bdb35f0/src/main/scala/util/intrusive_list/IntrusiveList.scala#L241-L257
The splitOn method returns an IntrusiveList that is not part of a Block or Procedure, and was designed to help split one Block into multiple by removing statements past a certain point from an IntrusiveList and adding them to a new, separate IntrusiveList. The issue now is that this new IntrusiveList produced by splitOn cannot be used in the creation of a new Block, as the statements are already in the IntrusiveList produced by splitOn and so cannot be added to the new Block's IntrusiveList. Making this work at present would require some messy workaround like moving all of the result of splitOn's contents to a new List (that is not an IntrusiveList) and then adding that to a new Block, which seems less than ideal.

This should be changed in some way so that the design of IntrusiveList does not contradict itself as it currently does. 

This sort of issue is also why having an earlier intermediate GTIRB/ASLp data structure would be helpful - it means we can avoid having to deal with the intricacies of IntrusiveList while creating the broader program data structure, as I suggested in #183.

l-kent:
Fixes #198, #199 and #200

It wasn't possible to make IntrusiveList.from behave differently just for IntrusiveList inputs, so I made splitOn just return an ArrayBuffer.

### DIFF
--- a/src/main/scala/ir/IRCursor.scala
+++ b/src/main/scala/ir/IRCursor.scala
@@ -46,8 +46,8 @@ object IRWalk:
 
   def commandBegin(pos: CFGPosition) : Option[Command] = {
     pos match {
-      case p: Procedure => p.entryBlock.map(b => b.statements.headOption().getOrElse(b.jump))
-      case b: Block => Some(b.statements.headOption().getOrElse(b.jump))
+      case p: Procedure => p.entryBlock.map(b => b.statements.headOption.getOrElse(b.jump))
+      case b: Block => Some(b.statements.headOption.getOrElse(b.jump))
       case c: Command => Some(c)
     }
   }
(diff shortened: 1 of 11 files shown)


ailrst:
Is there a test for the specific problem this fixes? If not can you add one please

ailrst (approved):

PR #212, opened 2024-05-30
Thread identification

l-kent:
Splits threads into separate boogie output files. Currently only works for pthreads (support for other thread libraries can be added later)

`-t` flag enables this functionality and requires the `--analyse` flag to be used alongside it.


ailrst:
Can you check in the C source for the two examples please

l-kent:
It already exists for basic_assign_pthread and never existed for the 'pthread' example (someone else committed it but without the source) which is why I removed it.

ailrst:
Oh ok my mistake, I thought it was a newly added example

l-kent:
This obviously needs to wait for #204 to be merged in.

src/main/scala/util/RunUtils.scala
```diff
@@ -455,6 +463,70 @@ object IRTransform {
...
+                  val thread = ProgramThread(threadEntrance, mutable.Set(threadEntrance), Some(d))
+                  program.threads.addOne(thread)
+                case _ =>
+                  throw Exception("unexpected non-data region " + threadTargets.head + " as PointsTo result for R2 at " + d)
+              }
+            }
+          case _ =>
+        }
+      }
+    }
```
ailrst, line 513: This could be written in a functional style (since we have the iterator defined on the IR program you can do program.collect{}) but its still pretty interpret-able as is.

src/main/scala/util/RunUtils.scala
```diff
@@ -455,6 +463,70 @@ object IRTransform {
...
+      program.threads.addOne(mainThread)
+
+      val programProcs = program.procedures.toSet
+
+      // do reachability for all threads
+      for (thread <- program.threads) {
+        // TODO this will give the threads their procedures in a non-deterministic/fragile order and this should be changed to maintain the original order
+        val reachable = thread.entry.reachableFrom.intersect(programProcs)
+        thread.procedures.addAll(reachable)
+      }
```
ailrst, line 526: The procedures should be in the order provided by DFS right? And we can sort the threads by entry function name.

src/main/scala/Main.scala
```diff
@@ -45,9 +45,16 @@ object Main {
...
+      analysisResultsDot: Option[String],
+      @arg(name = "threads", short = 't', doc = "Separates threads into multiple .bpl files with given output filename as prefix (requires --analyse flag)")
+      threadSplit: Flag
   )
 
+  // need to parse thread split flag
+  // separate file names for each thread
+  // put method in runutils that splits threads if threads flag is true
+  // detect using the indirect call resolution stuff
+
```
ailrst, line 57: Comment not needed anymore I think?

src/main/scala/util/RunUtils.scala
```diff
@@ -455,6 +463,70 @@ object IRTransform {
...
+            // look up R2 value using points to results
+            val R2 = Register("R2", BitVecType(64))
+            val b = reachingDefs(d)
+            val R2Wrapper = RegisterVariableWrapper(R2, getDefinition(R2, d, reachingDefs))
+            val threadTargets = pointsTo(R2Wrapper)
+
+            if (threadTargets.size > 1) {
+              // currently can't handle case where the thread created is ambiguous
+              throw Exception("can't handle thread creation with more than one possible target")
+            }
```
ailrst, line 493: Since all we currently do is output extra programs to verify we can probably just attempt to resolve all the targets and add them as separate threads.

src/main/scala/Main.scala
```diff
@@ -78,9 +85,9 @@ object Main {
     val q = BASILConfig(
       loading = ILLoadingConfig(conf.inputFileName, conf.relfFileName, conf.specFileName, conf.dumpIL, conf.mainProcedureName, conf.procedureDepth),
       runInterpret = conf.interpret.value,
-      staticAnalysis = if conf.analyse.value then Some(StaticAnalysisConfig(conf.dumpIL, conf.analysisResults, conf.analysisResultsDot)) else None,
+      staticAnalysis = if conf.analyse.value then Some(StaticAnalysisConfig(conf.dumpIL, conf.analysisResults, conf.analysisResultsDot, conf.threadSplit.value)) else None,
```
ailrst, line 88: Could have --threads imply --analyse

ailrst (commented):
Everything looks fine, I left some comments for things to clean up.

src/main/scala/util/RunUtils.scala
l-kent: The DFS does not provide a consistent ordering, and this is not about the order of threads, but about the order of procedures within the threads.

PR #220, opened 2024-06-12
Cil visitor

ailrst:
This is a first pass at a port of the CIL visitor used by asli, it is especially good for modifying immutable expressions, with the main advantage being it does not require faking multiple-dispatch or re-defining the recursion tree for every implementation.


src/main/scala/ir/cilvisitor/CILVisitor.scala
```diff
@@ -0,0 +1,196 @@
...
+  */
+
+enum VisitAction[T]:
+  case SkipChildren()
+  case DoChildren()
+  case ChangeTo(e: T)
+  // changes to e, then visits children of e, then applies f to the result
+  case ChangeDoChildrenPost(e: T, f: T => T)
+
+import VisitAction._
```
katrinafyi, line 19: it would look slightly nicer to me if the enum cases didn't need to be prefixed with VisitAction. no big deal though.

src/main/scala/ir/cilvisitor/CILVisitor.scala
```diff
@@ -0,0 +1,196 @@
...
+  def vexpr(e: Expr): VisitAction[Expr] = VisitAction.DoChildren()
+  // right-expressions
+  def vvar(e: Variable): VisitAction[Variable] = VisitAction.DoChildren()
+  def vmem(e: Memory): VisitAction[Memory] = VisitAction.DoChildren()
+  def vload(e: MemoryLoad): VisitAction[MemoryLoad] = VisitAction.DoChildren()
+
+  // left-expressions
+  // Somewhat confusing that these do not get visited by vexpr
+  def vlvar(e: Variable): VisitAction[Variable] = VisitAction.DoChildren()
+  def vlmem(e: Memory): VisitAction[Memory] = VisitAction.DoChildren()
```
katrinafyi, line 40: unify these with the right-expression methods?

src/main/scala/ir/cilvisitor/CILVisitor.scala
```diff
@@ -0,0 +1,196 @@
...
+
+  def enter_scope(params: ArrayBuffer[Parameter]): Unit = ()
+  def leave_scope(outparam: ArrayBuffer[Parameter]) : Unit = ()
```
katrinafyi, line 43: remark: scopes don't really make sense for a basic-block IR.

src/main/scala/ir/cilvisitor/CILVisitor.scala
```diff
@@ -0,0 +1,196 @@
...
+
+
+class CILVisitor(val v: CILVisitorImpl) {
```
katrinafyi, line 46: suggest renaming "CILVisitorImpl", since the CILVisitor seems to have the main implementation, not the other way around. maybe "CILVisitorActions" for Impl?

src/main/scala/ir/cilvisitor/CILVisitor.scala
```diff
@@ -0,0 +1,196 @@
...
+  }
+
+  def doVisit[T](v: CILVisitorImpl, a: VisitAction[T], n: T, continue: (CILVisitorImpl, T) => T): T = {
+    a match {
+      case VisitAction.SkipChildren()             => n
+      case VisitAction.DoChildren()               => continue(v, n)
+      case VisitAction.ChangeTo(z)                => z
+      case VisitAction.ChangeDoChildrenPost(x, f) => f(continue(v, x))
+    }
+  }
```
katrinafyi, line 64: these are independent of the IR and can sit outside the visitor.

also, the continue function probably doesn't need to have a Visitor parameter. in fact, the corresponding "aux" functions in ASLp just ignore their visitor argument.

i suspect this was done in CIL because of the lack of closures in C. here and in Ocaml, the `continue` function can carry the CILVisitorImpl itself.

src/main/scala/ir/cilvisitor/CILVisitor.scala
```diff
@@ -0,0 +1,196 @@
...
+          case n :: tl => {
+            b.statements.replace(s, n)
+            var p = n
+            for (i <- tl) {
+              b.statements.insertAfter(p, i)
+              p = i
+            }
+          }
+        }
+      })
```
katrinafyi, line 159: a bit ceremonious. i assume this is required by intrusive list.

would this be useful to move into a general function for intrusive lists?

src/test/scala/ir/CILVisitorTest.scala
```diff
@@ -0,0 +1,209 @@
...
+      )
+    )
+
+    class ExprTrace extends CILVisitorImpl {
+      val res = mutable.ArrayBuffer[String]()
+
+      override def vlvar(e: Variable) = {
+        e match {
+          case Register(n, _) => res.append(n);
+          case _              => ???
```
katrinafyi, line 106: is this unreachable? maybe a comment

src/test/scala/ir/CILVisitorTest.scala
```diff
@@ -0,0 +1,209 @@
...
+                res.append(e match {
+                  case v: Variable => v.name
+                  case _           => ???
+                }); e
+              }
+            );
+          case _ => VisitAction.DoChildren()
+        }
+      }
+    }
```
katrinafyi, line 193: are these visitors more useful generally? i guess they are rather trivial.

src/main/scala/ir/Statement.scala
```diff
@@ -178,4 +178,4 @@ class IndirectCall(var target: Variable,
 }
 
 object IndirectCall:
-  def unapply(i: IndirectCall): Option[(Variable, Option[Block], Option[String])] = Some(i.target, i.returnTarget, i.label)
\ No newline at end of file
+  def unapply(i: IndirectCall): Option[(Variable, Option[Block], Option[String])] = Some(i.target, i.returnTarget, i.label)
```
katrinafyi, line 181: accidental change?

katrinafyi (commented):
A good first start, I think it could be tidied up a bit.

katrinafyi (approved):
Looks agreeable, thanks for the intrusive list changes.

src/main/scala/util/intrusive_list/IntrusiveList.scala
```diff
@@ -298,6 +298,47 @@ final class IntrusiveList[T <: IntrusiveListElement[T]] private (
...
+          p = insertAfter(p, i)
+        }
+        Some(p)
+    }
+  }
+
+
+  /**
+   * Insert an element before another element in the list.
+   * @param intrusiveListElement The element in the list to insert after, or None to indicate the end of the list.
```
katrinafyi, line 324: comment is wrong ("after" should be "before").

src/main/scala/ir/cilvisitor/CILVisitor.scala
ailrst: This is here for procedure calls mainly, it shouldnt get in the way

ailrst:
Merging as this doesn't impact any existing functionality

PR #221, opened 2024-06-13
Fix issue with memory coalescing

### LINKED ISSUE #201: Memory consolidation doesn't happen
In some cases, such for as the binary [here](https://github.com/UQ-PAC/bil-to-boogie-translator/files/14610421/cntlm-2024-03-15.tar.gz) (regardless of whether BAP or GTIRB/ASLp are used), initial memory isn't consolidated into 64-bit accesses and remains represented as 8-bit accesses in the Boogie output, which causes the output file to be impractically large. I have not yet investigated why.

l-kent:
Fixes #201

I have an idea for a more sophisticated fix too but it will take more effort so this will work for the time being.


src/main/scala/translating/IRToBoogie.scala
```diff
@@ -475,27 +475,57 @@ class IRToBoogie(var program: Program, var spec: Specification) {
     def coalesced(): List[BExpr] = {
       val sections = memory.flatMap { s =>
         // Phrase the memory condition in terms of 64-bit operations, as long as the memory
-        // region is a multiple of such operations and appropriately aligned
-        if (s.address % 8 == 0 && s.bytes.size % 8 == 0) {
-          for (b <- s.bytes.indices by 8) yield {
-            // Combine the byte constants into a 64-bit value
-            val sum: BigInt =
-              (0 until 8).foldLeft(BigInt(0))((x, y) => x + (s.bytes(b + y).value * (BigInt(2).pow(y * 8))))
+        // section is a multiple of 64-bits and appropriately aligned
```
ailrst, line 478: This doesn't seem accurate? It will coalesce any internal 64-bit bit multiple aligned subsection and concatenate the unaligned prefix and suffix.

src/main/scala/translating/IRToBoogie.scala
```diff
@@ -475,27 +475,57 @@ class IRToBoogie(var program: Program, var spec: Specification) {
...
+
+        val aligned = s.address % 8
+        val alignedStart = s.address + aligned
+        val alignedSizeMultiple = (s.bytes.size - aligned) % 8
+        val alignedEnd = s.bytes.size - alignedSizeMultiple
+
+        // Section that is safe to convert to 64-bit values
+        val section = for (b <- aligned until alignedEnd by 8) yield {
+          // Combine the byte constants into a 64-bit value
+          val sum: BigInt =
```
ailrst, line 488: This is a concat not really a sum although it uses sums

src/main/scala/translating/IRToBoogie.scala
```diff
@@ -475,27 +475,57 @@ class IRToBoogie(var program: Program, var spec: Specification) {
...
-              (0 until 8).foldLeft(BigInt(0))((x, y) => x + (s.bytes(b + y).value * (BigInt(2).pow(y * 8))))
+        // section is a multiple of 64-bits and appropriately aligned
+
+        val aligned = s.address % 8
+        val alignedStart = s.address + aligned
+        val alignedSizeMultiple = (s.bytes.size - aligned) % 8
+        val alignedEnd = s.bytes.size - alignedSizeMultiple
+
+        // Section that is safe to convert to 64-bit values
+        val section = for (b <- aligned until alignedEnd by 8) yield {
```
ailrst, line 486: Should this be alignedStart until alignedEnd by 8?

ailrst (changes requested):
Might be a problem with the aligned section loop if I'm reading it right, it probably deserves a comment on the whole function as well. It probably wouldn't be too hard to add a unit test for this function but I'm not requiring it.

src/main/scala/translating/IRToBoogie.scala
l-kent: This name was just taken from the original implementation

PR #222, opened 2024-07-01
Data Structure Analysis 

sadrabt:
- [x] implement DSA with union find
- [x] implement points-to visualiser functionality
- [x] Document tests 
- [x] extend DSAUnionFindSolver from the UnionFindSolver
- [x] replace tuples with case classes
- [x] single instance of stack symbolic access for each variable violated by indirect_call and malloc_with_local3
- [x]  bu(proc) in DSA throws no such elemenet exception for tests correct/indirect_call/clang_O2:GTIRB and correct/malloc_with_local3/clang_O2:GTIRB
- [ ] storeExitJump(exitToEntry(exit), d1, d2) throws cast error for correct/syscall/gcc_O2:GTIRB and correct/syscall/clang_O2:GTIRB
- [ ] resolve merge conflicts

includes symbolic access analysis and a reaching definitions analysis that works slightly differently from the one already implemented. (registers changed by a procedure call are defined by the call instruction)


l-kent:
I'm a bit confused by the commit history here - some of Yousif's commits that are not ready to merge in yet have been included, but then later reverted? Is there a neater way to do things than that?

src/main/scala/translating/ReadELFLoader.scala
```diff
@@ -76,7 +76,7 @@ object ReadELFLoader {
   }
 
   def visitSymbolTableRow(ctx: SymbolTableRowContext): Option[SpecGlobal] = {
-    if (ctx.entrytype.getText == "OBJECT" && ctx.bind.getText == "GLOBAL" && ctx.vis.getText == "DEFAULT") {
+    if ((ctx.entrytype.getText == "OBJECT" || ctx.entrytype.getText == "FUNC") && ctx.bind.getText == "GLOBAL" && ctx.vis.getText == "DEFAULT") {
```
l-kent, line 79: We really should be distinguishing between objects and functions in the symbol table - it will confuse things if we just treat them the same like this.

src/main/scala/analysis/SymbolicAccessAnalysis.scala
```diff
@@ -0,0 +1,148 @@
...
+
+case class HeapLocation(override val regionIdentifier: String, proc: Procedure, size: BigInt) extends MemoryLocation {
+  override def toString: String = s"Heap($regionIdentifier, $size)"
+}
+
+case class DataLocation(override val regionIdentifier: String, start: BigInt, size: BigInt) extends MemoryLocation {
+  override def toString: String = s"Data($regionIdentifier, $start, $size)"
+}
+
+case class UnkownLocation(override val regionIdentifier: String, proc: Procedure) extends MemoryLocation {
```
l-kent, line 31: typo in `Unkown`

src/main/scala/analysis/solvers/IDESolver.scala
```diff
@@ -245,7 +245,10 @@ abstract class BackwardIDESolver[D, T, L <: Lattice[T]](program: Program)
 
   protected def callToReturn(call: GoTo): DirectCall = call.parent.jump.asInstanceOf[DirectCall]
 
-  protected def returnToCall(ret: DirectCall): GoTo = ret.parent.fallthrough.get
+  protected def returnToCall(ret: DirectCall): GoTo = 
+    if ret.parent.fallthrough.isEmpty then
+      print("")
```
l-kent, line 250: Why does this print an empty string?

src/main/scala/util/RunUtils.scala
```diff
@@ -484,6 +487,7 @@ object StaticAnalysis {
       config: StaticAnalysisConfig,
       iteration: Int
   ): StaticAnalysisContext = {
+    val before = System.nanoTime()
```
l-kent, line 490: Is this ever used?

src/test/scala/RegionBuilderTests.scala
l-kent, line 1: What is this supposed to be testing?

src/main/scala/analysis/InterprocSteensgaardAnalysis.scala
```diff
@@ -108,7 +108,8 @@ class InterprocSteensgaardAnalysis(
     var reducedRegions = Set.empty[MemoryRegion]
     binExpr.arg1 match {
       case variable: Variable =>
-        evaluateExpressionWithSSA(binExpr, constantProp(n), n, reachingDefs).foreach { b =>
+        val a = evaluateExpressionWithSSA(binExpr, constantProp(n), n, reachingDefs)
+          a.foreach { b =>
```
l-kent, line 112: What's the purpose of this? Was this just to help with debugging or something?

src/main/scala/analysis/Local.scala
```diff
@@ -0,0 +1,300 @@
...
+          val global = isGlobal(rhs, n).get
+          graph.mergeCells(lhsCell, global)
+        else if isStack(rhs, n).isDefined then // just in case stack can't be recognised in after this assignment
+          val stack = isStack(rhs, n).get
+          graph.mergeCells(lhsCell, stack)
+        else
+          expr match
+            case BinaryExpr(op, arg1: Variable, arg2) if op.equals(BVADD) && arg1.equals(stackPointer)
+              && evaluateExpression(arg2, constProp(n)).isDefined && evaluateExpression(arg2, constProp(n)).get.value >= BITVECNEGATIVE =>
+              val size = twosComplementToDec(decToBinary(evaluateExpression(arg2, constProp(n)).get.value))
```
l-kent, line 201: It's not good to execute evaluateExpression three times like this.

src/main/scala/analysis/Local.scala
```diff
@@ -0,0 +1,300 @@
...
+              val node = DSN(Some(graph))
+              node.allocationRegions.add(StackLocation("Stack_"+proc.name, proc, -size))
+              node.flags.stack = true
+              graph.mergeCells(lhsCell, node.cells(0))
+
+            case BinaryExpr(op, arg1: Variable, arg2) if /*varToSym.contains(n) &&  varToSym(n).contains(arg1) && */ evaluateExpression(arg2, constProp(n)).isDefined =>
+              val offset = evaluateExpression(arg2, constProp(n)).get.value
```
l-kent, line 208: It's not good to execute evaluateExpression twice like this

src/main/scala/translating/ReadELFLoader.scala
sadrabt: Do you have any advice on the best way to do this? I changed SpecGlobal to a trait and extended it with two 
case classes SpecObejct and SpecFunc. But I haven't changed the behaviour of either case class from the trait.

l-kent: I don't think splitting SpecGlobal into two separate case classes really does anything to help, I'd have to look at what it's actually used for in the DSA but that's quite complex.

sadrabt: Sorry I assumed you wanted to distinguish between the two for other parts of BASIL not the DSA itself. 
I don't believe we need to distinguish between global objects and functions for DSA. It's convenient to treat them the same since they are accessed similarly (constant base address + constant offset) and we can reuse the globals recognition functionality. 
I can only see two reason we might want to do this:

1. if we extend DSA to perform call resolution it may be useful but even then we can recognise functions by the fact that their corresponding DS node is used in an indirect call
2.  we can also check that object and function DS nodes are not used in wrong manner (e.g. an object  is used in an indirect call), however, I'm not sure how useful this would be

l-kent:
It'll take some time to get across all the details of this, so detailed comments would really help to make it easier to understand.

One thing that makes the code more difficult to follow at the moment is the widespread use of tuples - is it possible that replacing those with case classes would make sense?

src/main/scala/translating/ReadELFLoader.scala
l-kent: Rather than splitting SpecGlobal, it would make more sense to make a new parent trait which contains both SpecGlobal and the new case class for the symbol table function entries. We should probably rename SpecGlobal too since it's very disconnected from its original use now.

src/main/scala/analysis/SymbolicAccessAnalysis.scala
```diff
@@ -0,0 +1,151 @@
...
+        expr match
+          case BinaryExpr(op, arg1: Variable, arg2) =>
+            evaluateExpression(arg2, constProp(n)) match
+              case Some(v) =>
+                if op.equals(BVADD) && arg1.equals(stackPointer) && v.value >= BITVECNEGATIVE then
+                  d match
+                    case Left(value) if value.accessor == variable => Map()
+                    case Left(value) => Map(d -> IdEdge())
+                    case Right(_) =>
+                      val size = twosComplementToDec(decToBinary(v.value))
```
ailrst, line 105: As far as I can tell this is just implementing conversion from bv64 to signed integer, it would be better to add this to bitVectorEval as `bv2int`, and maintain bitvector-width type-safety rather than assuming 64 bits. :)

src/main/scala/analysis/Local.scala
```diff
@@ -230,7 +231,7 @@ class Local(
 
               if op.equals(BVADD) && arg1.equals(stackPointer)
               && arg2Offset.isDefined && arg2Offset.get.value >= BITVECNEGATIVE then
```
l-kent, line 233: BitVectorEval has the method 'isNegative' that can be used instead of comparing with BITVECNEGATIVE?

src/main/scala/analysis/BitVectorEval.scala
```diff
@@ -24,6 +24,32 @@ object BitVectorEval {
...
+    var counter: Int = 0
+    binaryNum.foreach(
+      n =>
+        if counter == binaryNum.length - 1 && n == 1 then
+          result = result - BigInt(2).pow(counter)
+        else if n == 1 then
+          result = result + BigInt(2).pow(counter)
+        counter += 1
+    )
+    result
```
l-kent, line 50: This is an inefficient way to convert from bitvector to signed integer. The simplest way would be to just check `isNegative`, and if `isNegative` returns true then return `b.value - BigInt(2).pow(b.size)`, otherwise just return `b.value`. 

I also think that 'signed' should be in the name of this method to avoid confusion.

src/main/scala/analysis/DSAUtility.scala
```diff
@@ -0,0 +1,785 @@
...
+    var global: Option[((BigInt, BigInt), (DSN, BigInt))]  = None
+    breakable {
+      for (elem <- globalMapping) {
+        val range = elem._1
+        if address >= range._1 && (address < range._2 || (range._1 == range._2 && range._2 == address)) then
+          global = Some(elem)
+          break
+      }
+    }
+    global
```
l-kent, line 171: This is another instance of tuples hindering the clarity of the code.

src/main/scala/analysis/DSAUtility.scala
```diff
@@ -0,0 +1,785 @@
...
+      case MemoryAssign(mem, index: Expr, value: Expr, endian, size: Int, label) =>
+        visitStackAccess(pos, index, size, m)
+      case _ => m
+
+  }
+
+
+  // make all globals
+  private val swappedOffsets = globalOffsets.map(_.swap)
+  val globalMapping: mutable.Map[(BigInt, BigInt), (DSN, BigInt)] = mutable.Map[(BigInt, BigInt), (DSN, BigInt)]()
```
l-kent, line 109: Another instance of tuples hindering the clarity of the code.

src/test/scala/LocalTest.scala
```diff
@@ -0,0 +1,856 @@
+import analysis.{AddressRange, DSC, DSG, DSN, DataLocation, HeapLocation}
+import ir.Endian.BigEndian
+import ir.{Assign, BVADD, BinaryExpr, BitVecLiteral, ConvertToSingleProcedureReturn, DirectCall, Memory, MemoryAssign, MemoryLoad, SharedMemory}
+import org.scalatest.funsuite.AnyFunSuite
+import test_util.TestUtil
+import ir.dsl.*
+import specification.Specification
+import util.{BASILConfig, BoogieGeneratorConfig, ILLoadingConfig, IRContext, RunUtils, StaticAnalysisConfig}
+
+class LocalTest extends AnyFunSuite, TestUtil {
```
l-kent, line 10: Rather than 'LocalTest' wouldn't the name DSATest be clearer for this file & class?

src/test/scala/LocalTest.scala
```diff
@@ -0,0 +1,856 @@
...
+//    assert(dsg.stackMapping.isEmpty)
+//    assert(dsg.pointTo(dsg.globalMapping(AddressRange(69680, 69684))._1.cells(0))._1.node.get.collapsed)
+  }
+
+
+
+
+  ignore("interproc unsafe pointer arithmetic") {
+    // test interproc unification with points-to that have internal offsets into cells
+  }
```
l-kent, line 216: Are these needed for anything? Should they be removed?

src/main/scala/analysis/DSAUtility.scala
```diff
@@ -0,0 +1,837 @@
...
+      node.flags.global = true
+      node.flags.incomplete = true
+      globalMapping.update(AddressRange(external.offset, external.offset), Field(node, 0))
+  )
+
+
+
+  // determine if an address is a global and return the corresponding global if it is.
+  def isGlobal(address: BigInt): Option[(AddressRange, Field)] =
+    var global: Option[(AddressRange, Field)]  = None
```
l-kent, line 174: This is another case that would benefit from using a case class instead of a tuple

src/main/scala/util/RunUtils.scala
```diff
@@ -948,8 +967,51 @@ object RunUtils {
...
+      paramResults = paramResults, //analysisResult.last.paramResults,
+      steensgaardResults = analysisResult.last.steensgaardResults,
+      mmmResults = analysisResult.last.mmmResults,
+      memoryRegionContents = analysisResult.last.memoryRegionContents,
+      symbolicAccessess = symResults, // analysisResult.last.symbolicAccessess,
+      locals = Some(dsa.locals.toMap),
+      bus = Some(dsa.bu.toMap),
+      tds = Some(dsa.td.toMap),
+      reachingDefs = analysisResult.last.reachingDefs
+    )
```
l-kent, line 1014: Why is this separate to the rest of the static analysis?

sadrabt: it relies on resolved procedure calls

src/main/scala/util/RunUtils.scala
```diff
@@ -948,8 +967,51 @@ object RunUtils {
...
+
     Logger.info(s"[!] Finished indirect call resolution after $iteration iterations")
-    analysisResult.last
+    StaticAnalysisContext(
+      cfg = analysisResult.last.cfg,
+      constPropResult = analysisResult.last.constPropResult,
+      IRconstPropResult = analysisResult.last.IRconstPropResult,
+      memoryRegionResult = analysisResult.last.memoryRegionResult,
+      vsaResult = analysisResult.last.vsaResult,
+      interLiveVarsResults = analysisResult.last.interLiveVarsResults,
```
l-kent, line 1004: I don't think this analysis is ever actually run so you may want to re-enable it where it makes sense to

src/main/scala/analysis/Local.scala
```diff
@@ -0,0 +1,337 @@
...
+  // variables to symbolic access map for each cfg position
+  val varToSym: Map[CFGPosition, Map[Variable, Set[SymbolicAccess]]] = symResults.foldLeft(Map[CFGPosition, Map[Variable, Set[SymbolicAccess]]]()) {
+    (outerMap, syms) =>
+      val position = syms._1
+      val innerMap = syms._2.foldLeft(Map[Variable, Set[SymbolicAccess]]()) {
+        (m, access) =>
+          if m.contains(access._1.accessor) then
+            // every variable pointing to a stack region ONLY has one symbolic access associated with it.
+            m(access._1.accessor).foreach(
+              sym => assert(!sym.symbolicBase.isInstanceOf[StackLocation])
```
l-kent, line 50: This assertion fails for a number of test cases when run with the analysis on:
- correct/indirect_call/clang:GTIRB
- correct/indirect_call/clang_no_plt_no_pic:GTIRB
- correct/indirect_call/clang_pic:GTIRB
- correct/indirect_call/gcc:GTIRB
- correct/indirect_call/gcc_no_plt_no_pic:GTIRB
- correct/indirect_call/gcc_pic:GTIRB
- correct/malloc_with_local3/clang:GTIRB
- correct/malloc_with_local3/clang_no_plt_no_pic:GTIRB
- correct/malloc_with_local3/clang_pic:GTIRB
- correct/malloc_with_local3/gcc:GTIRB
- correct/malloc_with_local3/gcc_no_plt_no_pic:GTIRB
- correct/malloc_with_local3/gcc_pic:GTIRB

src/main/scala/analysis/DSA.scala
```diff
@@ -0,0 +1,221 @@
...
+        val preds : Set[Procedure] = CallGraph.pred(proc)
+        queue.enqueueAll(CallGraph.pred(proc).diff(visited).intersect(domain))
+    )
+
+    // bottom up phase
+    while queue.nonEmpty do
+      val proc = queue.dequeue()
+      visited += proc
+      queue.enqueueAll(CallGraph.pred(proc).diff(visited))
+      val buGraph = bu(proc)
```
l-kent, line 90: NoSuchElementException gets thrown here for the tests correct/indirect_call/clang_O2:GTIRB and correct/malloc_with_local3/clang_O2:GTIRB when run with the analysis on.

l-kent:
https://github.com/UQ-PAC/BASIL/blob/c806362d0f7e867968c39808c2bc559eb06543cd/src/main/scala/analysis/solvers/IDESolver.scala#L105

The call to exitToEntry here causes a `ClassCastException: class ir.Procedure cannot be cast to class ir.IndirectCall` for the tests correct/syscall/clang_O2:GTIRB and correct/syscall/gcc_O2:GTIRB when run with the analysis on.

src/main/scala/analysis/DSAUtility.scala
```diff
@@ -0,0 +1,975 @@
...
+    _pointee.get
+
+  def growSize(size: BigInt): Boolean =
+    if size > largestAccessedSize then
+      largestAccessedSize = size
+      true
+    else false
+
+  override def toString: String = s"Cell(${if node.isDefined then node.get.toString else "NONE"}, $offset)"
+}
```
l-kent, line 908: This should be a class, not a case class, as it contains mutable variables.

Is the idea that `_pointee` is private? You have not made it private.

l-kent:
Is there an option to output the .dot files you demonstrated today? I can't find that anywhere, and when I attempted to manually add that functionality, none of the nodes in the .dot files I got it to output had any contents beyond a number. 

Rather than using .dot graphs though, it would be more useful for debugging if it could print the mappings in plain text, in an order that made sense - relating back to the order of statements in the IR?

src/main/scala/analysis/DSA.scala
```diff
@@ -0,0 +1,230 @@
...
+            globals: Set[SymbolTableEntry], globalOffsets: Map[BigInt, BigInt],
+            externalFunctions: Set[ExternalFunction],
+            reachingDefs: Map[CFGPosition, Map[Variable, Set[CFGPosition]]],
+            writesTo: Map[Procedure, Set[Register]],
+            params:  Map[Procedure, Set[Variable]]
+         ) extends Analysis[Map[Procedure, DSG]] {
+
+  val locals : mutable.Map[Procedure, DSG] = mutable.Map()
+  val bu: mutable.Map[Procedure, DSG] = mutable.Map()
+  val td: mutable.Map[Procedure, DSG] = mutable.Map()
```
l-kent, line 34: Is `td` the output of the analysis as such?

src/main/scala/analysis/DSAUtility.scala
sadrabt: Sometimes I want to check if _pointee is defined or not and othertimes I want to get a cell's pointee or create one for it if it doesn't have one. getPointee is not meant as an accessor  to a private variable but a helper that checks if a pointee is already defined and if not creates  one. I still want to read and write to _pointee outside of DSC class. I can make it private and create accessor methods for it but I don't really want to restrict other classes from accessing _pointee

sadrabt:
> Is there an option to output the .dot files you demonstrated today? I can't find that anywhere, and when I attempted to manually add that functionality, none of the nodes in the .dot files I got it to output had any contents beyond a number.

I pushed my most recent changes on, now you should be able to produce dot graphs from a DSG using
writeToFile(dsg.toDot, "test.dot")


> Rather than using .dot graphs though, it would be more useful for debugging if it could print the mappings in plain text, in an order that made sense - relating back to the order of statements in the IR?
Yeah this makes sense, I'll work on outputting text format

l-kent:
That's pretty much what I added in, but the DOT graph it produced didn't have anything in the nodes except their number, it didn't contain all the information that was in the graph you demonstrated today.

l-kent:
[mainTopDown.txt](https://github.com/user-attachments/files/16925266/mainTopDown.txt)
This is one of the outputs I got, for `src\test\correct\jumptable\clang\jumptable.adt`

sadrabt:
> That's pretty much what I added in, but the DOT graph it produced didn't have anything in the nodes except their number, it didn't contain all the information that was in the graph you demonstrated today.

could you please pull and try again

src/main/scala/analysis/WriteToAnalysis.scala
```diff
@@ -0,0 +1,50 @@
...
+              writtenTo.++=(getWritesTos(proc))
+            case _ =>
+      )
+
+      writesTo.update(proc, writtenTo.toSet)
+      writesTo(proc)
+  }
+
+  def analyze(): Map[Procedure, Set[Register]] =
+    program.procedures.foreach(getWritesTos)
```
l-kent, line 48: Just iterating over the procedures like this isn't going to produce correct results, since the result of getWritesTos for a procedure depends on the results for all the procedures that it calls. You need to do this in a depth-first order, so that you do the analysis for a procedure only after all the procedures it calls have been analysed.

src/main/scala/util/RunUtils.scala
l-kent: Why does it rely on resolved indirect calls? This analysis is doing something similar enough to the Steensgaard analysis that is currently used to resolve indirect calls, so why does it need the indirect calls to be resolved first?

l-kent:
Thanks, that works.

The DSA seems to run into problems with overlapping accesses. The following is from `src\test\correct\jumptable\clang\jumptable.bir` with comments added:

```
000003fe: R8 := 0x10000
// 0x10DC0 contains a pointer to the procedure add_two
// 0x10DC8 contains a pointer to the procedure add_six
00000404: R8 := R8 + 0xDC0
// loading 128 bits from 0x10DC0 loads both the 64-bit pointer at address 0x10DC0
// and the 64-bit pointer at address 0x10DC8 to V0, concatenated together
0000040b: V0 := mem[R8, el]:u128
// storing V0 to the stack stores the pointer to add_two at R31 + 0x10 and the pointer to add_six at R31 + 0x18
00000413: mem := mem with [R31 + 0x10, el]:u128 <- V0
0000041a: R8 := mem[R8 + 0x10, el]:u64
00000422: mem := mem with [R31 + 0x20, el]:u64 <- R8
00000429: R8 := mem[R31 + 0x10, el]:u64
0000042e: R30 := 0x7D0
00000431: call R8 with return %00000433

00000433:
// the DSA says that R8 at this assignment points to node 434 which is empty, but it should point to add_six
// or alternatively some node that represents that the pointers to add_two and add_six were loaded together and stored on the stack together
00000438: R8 := mem[R31 + 0x18, el]:u64
0000043d: R30 := 0x7D8
00000440: call R8 with return %00000442
```

l-kent:
Flags like CF, VF, etc. are inherently 1-bit values and can't point to anything, I don't think there is a reason to include them in this?

src/main/scala/util/RunUtils.scala
sadrabt: DSA can resolve indirect calls, however, this functionality was not implemented. Due to the lack of type info from the lifter, SAA (symbolic access analysis) is used to find possible pointers. SAA is inter procedural so it requires call resolution first. SAA results are heavily used throughout DSA. Most of that info probably can be gathered through local reaching definitions (to find stack locations), and the symbol table (for globals) but removing SAA causes the analysis to lose some information about heap objects and unknowns (loaded from memory) . 

I think it's possible completely remove SAA and fully rely on DSA for call resolution but it is beyond this pull request

l-kent: Resolving indirect calls is necessarily going to be an iterative process, so I don't think SAA has to depend on indirect call resolution at all. DSA is ultimately a form of points-to analysis so it doesn't really make sense to make it depend on the Steensgaard points-to analysis. This doesn't matter for now but is worth further consideration when we try to actually do anything with the DSA results.

src/test/scala/DSATest.scala
```diff
@@ -0,0 +1,707 @@
...
+    val stack8 = dsg.find(dsg.stackMapping(8).cells(0)) //  R31 + 8
+    val stack40 = dsg.find(dsg.stackMapping(40).cells(0))//  R31 + 40
+    val stack32 = dsg.find(dsg.stackMapping(32).cells(0)) //  R31 + 32
+    val stack24 = dsg.find(dsg.stackMapping(24).cells(0)) //  R31 + 24 and Malloc
+
+    assert(dsg.adjust(stack0.getPointee).equals(R29formal)) // R31 points to the frame pointer
+    assert(dsg.adjust(stack8.getPointee).equals(dsg.adjust(dsg.formals(R30)))) // R31 + 8 points to the link register
+
+
+    assert(dsg.adjust(stack32.getPointee).equals(stack24))  // R31 + 32 points to R31 + 24, 00000457
```
l-kent, line 61: R31 + 32 is later set to point to a heap allocation (at 00000466), but I can't see this in the DSA results at all. How does the analysis work when a stack location is assigned multiple values during a procedure's execution?

src/test/scala/DSATest.scala
```diff
@@ -0,0 +1,707 @@
...
+    assert(stack24.node.get.collapsed) // 00000497 collapses stack24 concatenation is currently unhandled, any objects referenced in an unhandled operation are collapsed
```
l-kent, line 62: Maybe it would be better to test the analysis on examples that can be handled properly?

src/test/scala/DSATest.scala
```diff
@@ -0,0 +1,707 @@
...
+    assert(dsg.adjust(stack24.getPointee).equals(stack24)) // 00000466, R31 + 32 and R31 + 24 pointees are merged
```
l-kent, line 63: What does this actually mean? My assumption would be that this is testing if the stack location R31 + 24 points to itself, but I don't think that's correct. What does this have to do with statement 00000466, where the heap pointer that is located at R31 + 24 is stored in R31 + 32?

src/test/scala/DSATest.scala
```diff
@@ -0,0 +1,707 @@
...
+    // test that all three calles have the same local graph
+    val callees = Set("sub_seven", "add_two", "add_six")
+    callees.foreach(
+      callee =>
+        val dsg = results.analysis.get.locals.get(program.procs(callee))
+        assert(dsg.stackMapping.isEmpty) // stack is not used in either callee
+        assertJumptable2Globals(dsg) // globals should be the same everywhere unused in callees
+        // x should point to a collapsed object, in all 3 functions
+        // all three load value of x
+        // the analysis doesn't know if x is a pointer or not therefore assumes it is for soundness
```
l-kent, line 118: Can the analysis not tell that x is not a pointer because it is a 32 bit value and therefore not a pointer?

l-kent:
In general it would be preferable if you do not use the examples folder for any tests - the existing contents of the examples folder are very out of date and we really should delete most of them. Any new test files should be put in their own folder in the tests folder - move the /unsafe_pointer_arithmetic and /interproc_pointer_arithmetic folders from /examples to /src/test/dsa.

src/test/scala/DSATest.scala
```diff
@@ -0,0 +1,707 @@
...
+    assert(dsg.adjust(dsg.formals(R1)).equals(dsg.adjust(dsg.formals(R2))))
+
+    // R6 and R7 address the same cell
+    assert(dsg.find(dsg.varToCell(locAssign1)(R6)).cell.equals(dsg.find(dsg.varToCell(locAssign2)(R7)).cell))
+    // however, they address different internal offets in those cells
+    assert(dsg.find(dsg.varToCell(locAssign1)(R6)).internalOffset == 0)
+    assert(dsg.find(dsg.varToCell(locAssign2)(R7)).internalOffset == 1)
+    // R6 points to input R1
+    assert(dsg.adjust(dsg.varToCell(locAssign1)(R6))._pointee.isDefined)
+    assert(dsg.adjust(dsg.adjust(dsg.varToCell(locAssign1)(R6)).getPointee).equals(dsg.adjust(dsg.formals(R1))))
```
l-kent, line 316: Is this because the analysis has no concept of most of the value of R1 (at address R0 + 5) being overwritten by the write of the value of R2 to address R0 + 4? The analysis just says that both R1 and R2 were written to R0 + 4 at some point? If this is the case, it should be explicit in the comments.

src/test/scala/DSATest.scala
```diff
@@ -0,0 +1,707 @@
...
+    assert(stack8.equals(dsg.adjust(dsg.formals(R0))))
+    assert(stack8.offset == 0)
+    // stack 24 points to the formal argument object at offset 16, instead
+    assert(stack24.equals(dsg.adjust(dsg.formals(R0)).node.get.cells(16)))
+  }
+
+
+  test("internal merge") {
+    // this is an internal merge (two cells of the same node overlap and are merged together)
+    val mem = SharedMemory("mem", 10000, 10000)
```
l-kent, line 286: These are not correct parameters - the addressSize should be 64 (since addresses are 64-bit) and the valueSize should be 8 (since 8 bits are stored at every offset).

src/test/scala/DSATest.scala
```diff
@@ -0,0 +1,707 @@
...
+
+    val results = RunUtils.staticAnalysis(StaticAnalysisConfig(None, None, None), IRContext(Set.empty, Set.empty, Set.empty, Map.empty, Specification(Set(), Set(), Map(), List(), List(), List(), Set()), program))
+    val dsg: DSG = results.locals.get(program.mainProcedure)
+    // check that R5 points to separate cell at offset 13
+    assert(dsg.find(dsg.varToCell(locAssign3)(R5)).offset == 13)
+  }
+
+  test("offsetting from middle of cell to the same cell") {
+    // similar to above except instead of creating new cell the last assign
+    // points R7's cell at an internal offset of 8
```
l-kent, line 350: This isn't what the test actually tests, it tests that R7 has an internal offset of 1 and that R5 has an internal offset of 8.

src/test/scala/DSATest.scala
```diff
@@ -0,0 +1,707 @@
...
+    // however, they address different internal offets in those cells
+    assert(dsg.find(dsg.varToCell(locAssign1)(R6)).internalOffset == 0)
+    assert(dsg.find(dsg.varToCell(locAssign2)(R7)).internalOffset == 1)
+    // R6 points to input R1
+    assert(dsg.adjust(dsg.varToCell(locAssign1)(R6))._pointee.isDefined)
+    assert(dsg.adjust(dsg.adjust(dsg.varToCell(locAssign1)(R6)).getPointee).equals(dsg.adjust(dsg.formals(R1))))
+
+  }
+
+  test("offsetting from middle of cell to a new cell") {
```
l-kent, line 320: Looking at the dot graph for this test, it doesn't seem like it has the node that R0/R5/R6/R7 point to (all at varying offsets) point to anything - is this intentional?

src/main/scala/analysis/DSAUtility.scala
```diff
@@ -0,0 +1,983 @@
...
+    }
+
+    structs ++= varToCell.foldLeft(Set[DotStruct]()) {
+      (s, r) =>
+        val pos = r._1
+        val mapping = r._2
+        s ++ mapping.foldLeft(Set[DotStruct]()) {
+          (k, n) =>
+            val variable = n._1.name
+            k + DotStruct(s"SSA_${pos.toShortString.slice(1, 9)}_$variable", s"SSA_${pos}_$variable", None, false)
```
l-kent, line 241: This cuts off the first character of `pos.toShortString` which can cause problems for readability.

src/main/scala/analysis/Local.scala
```diff
@@ -0,0 +1,351 @@
...
+            )
+            assert(!access._1.symbolicBase.isInstanceOf[StackLocation])
```
l-kent, line 52: This assertion fails for the input examples\indirect_call_outparam\indirect_call_outparam.adt

src/main/scala/analysis/DSAUtility.scala
```diff
@@ -0,0 +1,983 @@
...
+  /**
+   * this function takes a stackMapping and updates it based on a memory load and memory store
+   * @param pos memory load or store IL position
+   * @param index memory location of load or store
+   * @param size size of the load or store
+   * @param m stack mapping
+   * @return updated stack mapping
+   */
+  private def visitStackAccess(pos: CFGPosition, index: Expr, size: Int, m: Map[BigInt, DSN]) : Map[BigInt, DSN] =
+    val byteSize = (size.toDouble / 8).ceil.toInt
```
l-kent, line 67: what is the point of converting to bytes like this, and why bother casting to double and back to int? A memory access size must always be a multiple of 8 bits.

src/main/scala/analysis/DSAUtility.scala
```diff
@@ -0,0 +1,983 @@
...
+}
+
+/**
+ * a cell in DSA
+ * @param node the node this cell belongs to
+ * @param offset the offset of the cell
+ */
+case class DSC(node: Option[DSN], offset: BigInt)
+{
+  var largestAccessedSize: BigInt = 0
```
l-kent, line 893: Why does this need to be a BigInt?

src/main/scala/analysis/DSAUtility.scala
l-kent: This is a bit unclear in the current code and better names or comments would probably help.

What's with the _ in the name then? In some languages that's used for private variables in a class?

src/main/scala/analysis/DSAUtility.scala
```diff
@@ -0,0 +1,983 @@
...
+    case literal: Literal => literal
+    case Repeat(repeats, body) => Repeat(repeats, unwrapPaddingAndSlicing(body))
+    case SignExtend(extension, body) => SignExtend(extension, unwrapPaddingAndSlicing(body))
+    case UnaryExpr(op, arg) => UnaryExpr(op, arg)
+    case BinaryExpr(op, arg1, arg2) => BinaryExpr(op, unwrapPaddingAndSlicing(arg1), unwrapPaddingAndSlicing(arg2))
+    case MemoryLoad(mem, index, endian, size) => MemoryLoad(mem, unwrapPaddingAndSlicing(index), endian, size)
+    case variable: Variable => variable
+    case Extract(end, start, body) /*if start == 0 && end == 32*/ => unwrapPaddingAndSlicing(body) // this may make it unsound
+    case ZeroExtend(extension, body) => unwrapPaddingAndSlicing(body)
+    case _ => expr
```
l-kent, line 966: Ignoring extract and extend operations is not going to produce correct results.

src/main/scala/analysis/Local.scala
```diff
@@ -0,0 +1,351 @@
...
+        val lhsCell = graph.adjust(graph.varToCell(n)(variable))
+        var global = isGlobal(rhs, n)
+        var stack = isStack(rhs, n)
+        if global.isDefined then // Rx = global address
+          graph.mergeCells(lhsCell, global.get)
+        else if stack.isDefined then // Rx = stack address
+          graph.mergeCells(lhsCell, stack.get)
+        else
+          expr match
+            case BinaryExpr(op, arg1: Variable, arg2) => // Rx = Rx + c
```
l-kent, line 246: Why does this treat Rx = Rx + c differently to Rx = c + Rx (which is not handled)?

src/main/scala/analysis/Local.scala
```diff
@@ -0,0 +1,351 @@
...
+                () // the stack is handled prior to this
+//                val size = bv2SignedInt(arg2Offset.get)
+//                val node = DSN(Some(graph))
+//                node.allocationRegions.add(StackLocation("Stack_"+proc.name, proc, -size))
+//                node.flags.stack = true
+//                graph.mergeCells(lhsCell, node.cells(0))
+              else if /*varToSym.contains(n) &&  varToSym(n).contains(arg1) && */ arg2Offset.isDefined then
+                // merge lhs with cell(s) corresponding to (arg1 + arg2) where arg1 is cell and arg2 is an offset
+                val offset = evaluateExpression(arg2, constProp(n)).get.value
+                visitPointerArithmeticOperation(n, lhsCell, arg1, 0, false, offset)
```
l-kent, line 259: This is incorrectly assuming that the operation is BVADD.

src/main/scala/analysis/Local.scala
```diff
@@ -0,0 +1,351 @@
...
+              else if stack.isDefined then
+                graph.mergeCells(lhsCell, graph.adjust(graph.find(stack.get).getPointee))
+              else
+                index match
+                  case BinaryExpr(op, arg1: Variable, arg2) =>
+                    evaluateExpression(arg2, constProp(n)) match
+                      case Some(v) =>
+//                        assert(varToSym(n).contains(arg1))
+                        val offset = v.value
+                        visitPointerArithmeticOperation(n, lhsCell, arg1, byteSize, true, offset)
```
l-kent, line 283: This is incorrectly assuming that the operation is BVADD

src/main/scala/analysis/Local.scala
```diff
@@ -0,0 +1,351 @@
...
+                      case None =>
+//                        assert(varToSym(n).contains(arg1))
+                        visitPointerArithmeticOperation(n, lhsCell, arg1, byteSize, true, 0, true)
```
l-kent, line 286: Is it correct to just assume that the value of arg2 is 0 if it couldn't be evaluated?

src/main/scala/analysis/Local.scala
```diff
@@ -0,0 +1,351 @@
...
+          else if stack.isDefined then
+            graph.adjust(graph.find(stack.get).getPointee)
+          else
+            index match
+              case BinaryExpr(op, arg1: Variable, arg2) =>
+                evaluateExpression(arg2, constProp(n)) match
+                  case Some(v) =>
+//                    assert(varToSym(n).contains(arg1))
+                    val offset = v.value
+                    visitPointerArithmeticOperation(n, DSN(Some(graph)).cells(0), arg1, byteSize, true, offset)
```
l-kent, line 314: This is once again assuming that the operation is addition, incorrectly.

src/main/scala/analysis/Local.scala
```diff
@@ -0,0 +1,351 @@
...
+                  case None =>
+//                    assert(varToSym(n).contains(arg1))
+                    visitPointerArithmeticOperation(n, DSN(Some(graph)).cells(0), arg1, byteSize, true, 0, true)
```
l-kent, line 317: This is assuming that the operation is addition and that the value of arg2 is 0 if it could not be evaluated.

src/main/scala/analysis/DSAUtility.scala
```diff
@@ -0,0 +1,983 @@
...
+
+
+  /**
+   * Collapses the node causing it to lose field sensitivity
+   */
+  def collapseNode(n: DSN): DSN =
+
+
+    val (term, offset) = solver.find(n.term)
+    val node: DSN = term.asInstanceOf[Derm].node
```
l-kent, line 325: If you are using asInstanceOf like this, it likely means you have not used types correctly. If the first output from solver.find is guaranteed to be a Derm, that should be reflected by the types.

src/main/scala/analysis/DSAUtility.scala
```diff
@@ -0,0 +1,983 @@
...
+      else
+        cell1._pointee = cell2._pointee
+    val internalOffsetChange = cell2.offset - cell1.offset
+    cell2.node.get.cells.remove(cell2.offset)
+    cell1.growSize((cell2.offset - cell1.offset) + cell2.largestAccessedSize) // might cause another collapse
+    cell1
+
+
+  //  private val parent = mutable.Map[DSC, DSC]()
+  val solver: DSAUnionFindSolver[UniTerm] = DSAUnionFindSolver()
```
l-kent, line 416: Why is this not just DSAUnionFindSolver[Derm]?

src/main/scala/analysis/DSAUtility.scala
```diff
@@ -0,0 +1,983 @@
...
+
+
+
+/** Terms used in unification.
+ */
+sealed trait UniTerm
+
+/** A term variable in the solver
+ */
+case class Derm(node: DSN) extends UniTerm with Var[UniTerm] {
```
l-kent, line 976: This could really use a clearer name

src/main/scala/analysis/solvers/DSAUnionFindSolver.scala
```diff
@@ -0,0 +1,72 @@
...
+
+import scala.collection.mutable
+
+class DSAUnionFindSolver[A] {
+
+
+  val parent = mutable.Map[Term[A], Term[A]]()
+  val offsets = mutable.Map[Term[A], BigInt]()
+
+  def unify(t1: Term[A], t2: Term[A], offset: BigInt): Unit = {
```
l-kent, line 11: What does the addition of the offset parameter mean in the UnionFindSolver?

src/main/scala/analysis/DSAUtility.scala
l-kent: Oh I see this is an issue caused by the use of types from TIP, so it's not straight-forward to fix. Hmm

src/test/scala/DSATest.scala
sadrabt: image

This is from "MACROSCOPIC DATA STRUCTURE ANALYSIS AND OPTIMIZATION" (Lattner, 2005) Figure 3.5, "local analysis function". 

A memory assignment is *Y = X is handled by merge(Ev(X), E(Ev(Y)) where "Ev: virtual register -> slice"  gives the pointee of virtual register (in this example r0 or X is replaced by tmp1 as registers are treated as SSA form and tmp2 is a implicit virtual register). E: cell -> slice gives the pointee of a cell. 
Every stack position (R31 + C) is treated as a special virtual register here.
normalised it would look like 
pos := R31 + C
*pos := tmp
merge(Ev(tmp), E(Ev(pos)), where tmp and pos  are SSA variables

now E(Ev(pos)) is either a single region/node for the whole stack such as:
Node R31 with cells at offsets: 0, 8 ... 24, 32
or treated as separate node for each offset used on the stack Node R31 + 0, Node R31 + 8, ... Node R31 + 32,
The stack access patterns are known so we can create the nodes (know their size and whether two nodes have overlapped accesses in order to merge them to one region/node prior). 

if a stack address is stored in another position in memory
mem = 0x000001
*mem = R31 + 8

we will have to to merge(Ev(R31 + 8), E(Ev(mem))), if the stack is treated as a single node, the node pointed to by the memory location mem is unified with the node of the stack. This will cause the stack to lose all field sensitivity eventually since the cell E(Ev(mem) is unlikely to have node that structurally is equivalent to the stack.
However, if multiple regions are created this is avoided. This is still sound if we have correctly found all the stack accesses and merged any overlapping R31 + C accesses into one. 

Therefore tmp := R31 + C is equivalent to X = alloca in the figure above.  that is we have 
Ev(tmp) = new node(R31 + C).0 where node(R31 + C).0 is the cell at offset 0 of the new node which is marked as a stack region. Any reads/writes to the stack position R31 + C, now is represented by E(Ev(tmp))

Position 0000044F: tmp1 := R31 + 24 // Ev(tmp1) = new Node(R31 + 24).0
// implicit normalisation: tmp2 := R31 + 32  // Ev(tmp2) = new Node(R31 + 32).0
Position 00000457:  *tmp2  := tmp1 // merge(Ev(tmp1), E(Ev(tmp2)))
// Therefore, Node(R31 + 32).0.pointee is merged with  Node(R31 + 24).0, making E(Node(R31 + 32).0) == Node(R31 + 24).0

at position 00000446 Ev(R0) == Malloc_Node  == E(Node(R31 + 24).0)
we have 
mem := mem with [R31 + 0x20, el]:u64 <- R0
*(R31 + 32) := R0
merge(Ev(R0), E(Ev(R31+ 32)) 
== merge(E(Node(R31 + 24).0),  E(Node(R31 + 32).0)) 
== merge(E(Node(R31 + 24).0), Node(R31 + 24).0)
which merges make the stack + 24 point to itself

In general any memory location which has multiple values assigned to it, is has its pointee merged with all of them.
The analysis is flow insensitive, therefore, the order of assignment don't matter

src/test/scala/DSATest.scala
sadrabt: This is part of DSA's approach to handling more instructions while remaining sound.  it's the last instruction in figure 3.5 of the Lattner dissertation.

sadrabt: https://github.com/UQ-PAC/BASIL/pull/222#discussion_r1751272452

src/test/scala/DSATest.scala
sadrabt: https://github.com/UQ-PAC/BASIL/pull/222#discussion_r1751272452

src/test/scala/DSATest.scala
l-kent: I guess what I didn't understand is why only some stack stores seem to appear in nodes. Is it just that the first stack store at each offset has a node created for it - if so, why refer to a particular MemoryStore in the node?

src/test/scala/DSATest.scala
sadrabt: the test comments says
`// object of formals R1 and R2 are written to overlapping fields of the same node? causing them to be merged together assert(dsg.adjust(dsg.formals(R1)).equals(dsg.adjust(dsg.formals(R2))))`
*R7 := R1 // equivalent to *(R6 + 1) := R1
*R6 := R2 
Since R6 and R7 point to the overlapping cells from the same node (Node pointed to offset 0 by R0) at offsets 4 and 5 respectively, the overlapping cells are merged together to form a cell encompassing both. This will need to point to two cells (the outgoing edges of cells at offset 4 and 5 respectively). So these two outgoing edges are unified. 
After the analysis is done, R6 and R7  both point to the same cell at different internal offsets, that cell will have one outgoing edge. Outgoing edge is a cell in the node resulting from the unification of the nodes of R1 and R2.

The test checks that R1 and R2 are pointing to the same cell since they have been unified and states that in comment.
it checks R6 and R7 are the slices with the same cell and different internal offsets.
then it checks that R6 points to R1
this is the same as R7 points to R2
or R6 points to R2
or R7 points to R1

l-kent: It would be helpful if this sort of thing, explaining what you are actually trying to test, was more explicit in the test comments - 'R6 points to input R1' is just not very clear on its own, and talking about something pointing to a register is also unintuitive.

src/test/scala/DSATest.scala
l-kent: In general:
- 64-bit values: possibly pointers
- 128-bit values: potentially two 64-bit pointers concatenated together
- anything else: definitely not a pointer

src/test/scala/DSATest.scala
l-kent: More detailed explanation in the comments for tests like this where you are testing something not immediately obvious would help.

src/test/scala/DSATest.scala
l-kent: The concatenation is only because the input you are using was incorrectly lifted by a very old version of BAP. If you want to test this behaviour I suggest writing an example designed to test it.

src/main/scala/analysis/DSAUtility.scala
l-kent: Since pointers are 64-bit values, if a SignExtend/ZeroExtend/Repeat/Extract is applied to a pointer, the result will not be a  pointer, so I assume that means the nodes should be collapsed? It may be possible to refine that though, some of the cases we need to care about don't exist in higher level code so they weren't considered in the DSA paper.

src/test/scala/DSATest.scala
sadrabt: This is allows us to know which CFGPosittion(s) created the current node and was mainly used for debugging

src/main/scala/analysis/WriteToAnalysis.scala
sadrabt: getWritesTos  is recursive

src/main/scala/analysis/DSAUtility.scala
```diff
@@ -0,0 +1,983 @@
...
+          case _ => m
+      case MemoryAssign(mem, index: Expr, value: Expr, endian, size: Int, label) =>
+        visitStackAccess(pos, index, size, m)
+      case _ => m
+
+  }
+
+  private def createStackMapping(label: String, offset: BigInt, m:  Map[BigInt, DSN], byteSize: Int) :  Map[BigInt, DSN]=
+    if m.contains(offset) then
+      assert(!m(offset).cells(0).growSize(byteSize))
```
l-kent, line 105: This assertion can fail unexpectedly, though I foolishly didn't keep track of the example where it does - I'll try to recreate it later

src/main/scala/analysis/DSAUtility.scala
```diff
@@ -0,0 +1,983 @@
...
+    index match
+      case BinaryExpr(op, arg1: Variable, arg2) if varToSym.contains(pos) && varToSym(pos).contains(arg1) &&
+        evaluateExpression(arg2, constProp(pos)).isDefined =>
+        var offset = evaluateExpression(arg2, constProp(pos)).get.value
+        varToSym(pos)(arg1).foldLeft(m) { // go through all the symbolic accesses tied to arg1 at pos
+          (m, sym) =>
+            sym match
+              case SymbolicAccess(accessor, StackLocation(regionIdentifier, proc, size), symOffset) => // only consider stack accesses
+                offset = offset + symOffset
+                createStackMapping(pos.toShortString, offset, m, byteSize)
```
l-kent, line 77: `offset` being a var here seems incorrect - if varToSym(pos)(arg1) contained multiple stack accesses (which it doesn't seem to in practice but I can't be sure about that), then any subsequent offsets would be incorrect as they would not relate to the base offset?

src/main/scala/analysis/SymbolicAccessAnalysis.scala
```diff
@@ -0,0 +1,152 @@
...
+                    case Right(_) =>
+                      val size = bv2SignedInt(v)
+                      Map(d -> IdEdge(), Left(SymbolicAccess(variable, StackLocation(s"Stack_${procedure(n).name}", procedure(n), -size), 0)) -> ConstEdge(TwoElementTop))
+                else
+                  d match
+                    case Left(value) if value.accessor == arg1 =>
+                      val offsetUpdate = evaluateExpression(arg2, constProp(n)).get.value
+                      val result: Map[DL, EdgeFunction[TwoElement]] = Map(Left(SymbolicAccess(variable, value.symbolicBase, value.offset + offsetUpdate)) -> ConstEdge(TwoElementTop))
+                      if value.accessor != variable then
+                        result + (d -> IdEdge())
```
l-kent, line 114: This seems to assume that op == BVADD even though that isn't checked?

src/main/scala/analysis/SymbolicAccessAnalysis.scala
```diff
@@ -0,0 +1,152 @@
...
+import analysis.BitVectorEval.{bv2SignedInt, isNegative}
+import analysis.solvers.ForwardIDESolver
+import ir.IRWalk.procedure
+import ir.{Assign, BVADD, BinaryExpr, BitVecLiteral, BitVecType, CFGPosition, DirectCall, Expr, Extract, GoTo, IndirectCall, Literal, Memory, MemoryLoad, Procedure, Program, Register, Repeat, SignExtend, UnaryExpr, Variable, ZeroExtend}
+
+import java.math.BigInteger
+
+case class SymbolicAccess(accessor: Variable, symbolicBase: MemoryLocation, offset: BigInt) {
+  override def toString: String = s"SymbolicAccess($accessor, $symbolicBase, $offset)"
+}
```
l-kent, line 12: These probably need to include the size of the access as well to give fully correct results

src/main/scala/analysis/SymbolicAccessAnalysis.scala
```diff
@@ -0,0 +1,152 @@
...
+case class UnknownLocation(override val regionIdentifier: String, proc: Procedure) extends MemoryLocation {
+  override def toString: String = s"Unknown($regionIdentifier)"
+}
+
+/**
+ * environment transformers for SAA or symbolic access analysis
+ * Combination of reaching definitions and constant propagation
+ * elements in D are symbolic accesses of the form (variable, symbolic base, concrete offset)
+ * lattice L is a binary lattice with top and bottom
+ */
```
l-kent, line 41: What do the results of this analysis actually mean, exactly? It produces a Map[CFGPosition, Map[SymbolicAccess, TwoElement]] but I don't really follow what each part means here, especially the TwoElement.

src/main/scala/analysis/Local.scala
l-kent: These assertions just don't seem to hold up much in practice, what's the assumption behind them? There seems to be an issue with it

src/main/scala/analysis/DSAUtility.scala
sadrabt: you are right. I'll rename it

src/test/scala/DSATest.scala
sadrabt: no, this is due to an issue in the DSG::toDot.it should be resolved now.

src/main/scala/analysis/DSAUtility.scala
sadrabt: addressing is done based on bytes, so to be consistent access sizes are converted to bytes to determine overlapping accesses based on the address of the accesses and their known size. the casting is unnecessary and will be removed.

src/main/scala/analysis/Local.scala
sadrabt: do you know of an example of Rc = c + Rx?

l-kent: It's better practice to handle all inputs that we can, rather than only matching a very narrow set of inputs that we're familiar with. Making assumptions like this can cause problems later on - the less assumptions about what the input looks like, the better.

src/main/scala/analysis/Local.scala
sadrabt: yes, this is should be more clear but in this case the resulting node is collapsed with the second true passed to the function, therefore, the offset doesn't matter

src/main/scala/analysis/Local.scala
sadrabt: https://github.com/UQ-PAC/BASIL/pull/222#discussion_r1770753140

src/main/scala/analysis/SymbolicAccessAnalysis.scala
sadrabt: TwoElement is binary lattice element (top or bottom).
It's similar to live vars where there is a mapping from CFGPositions (flow-sensitive) to symbolic bases that are alive at that position.

src/main/scala/analysis/Local.scala
l-kent: I think this is still an issue - visitPointerArithmeticOperation assumes that the operation is addition and this is not a case that has collapse = true

src/main/scala/analysis/SymbolicAccessAnalysis.scala
l-kent: I know what a TwoElement is, what I'm wondering is what Top and Bottom represent in this context - what does it mean that a SymbolicAccess maps to Top or to Bottom?

src/main/scala/analysis/Local.scala
l-kent: https://github.com/UQ-PAC/BASIL/pull/222#discussion_r1770765651

src/main/scala/analysis/Local.scala
l-kent: https://github.com/UQ-PAC/BASIL/pull/222#discussion_r1770765651

src/main/scala/analysis/Local.scala
l-kent: Is it the case that any other operation is equivalent to Rx = Ry + 0 though? The result isn't even necessarily offset from Ry if it's not addition or subtraction?

l-kent: Shouldn't it be unsupportedPointerArithmeticOperation that's used for this case (where the operation isn't addition) not visitPointerArithmeticOperation?

src/main/scala/analysis/DSAUtility.scala
sadrabt: I think this would fail if there are accesses of different sizes to the same stack position. This can be handled by collapsing stack positions ie (if R31 + 8 is written to using different sizes we assume the size of R31 + 8 is the maximum size of the accesses.

src/main/scala/analysis/SymbolicAccessAnalysis.scala
sadrabt: this is a reaching definitions + constant propagation combined

all this track looks something like this
cfgposition -> [(r0 == R31 + c) -> Top],

it conveys that at position cfgposition, R0 may be an alias for R31 + C where C is a known constant. 
Note that not all accesses have size associated with them (e.g. in the example above r0 or R31 + c may never be used as index into memory (in memory read or write with a size). 

I think what may confusing here is the they are named SymbolicAccess but it is really a Symbolic Address not an access

src/main/scala/analysis/DSAUtility.scala
sadrabt: You are correct. I think there is an implicit assumption here from the assertion in local that there is only on stack location associated with each variable at each cfgposition. However, that assertions doesn't always hold.

src/main/scala/analysis/SymbolicAccessAnalysis.scala
l-kent: That makes sense, thanks, maybe just rename it to 'SymbolicAddress' then?

src/main/scala/analysis/Local.scala
sadrabt: Sorry I made changes locally to check for op.bvadd and marked this as resolved. It is enough to analyse some complex examples. The other operations are unclear specially, since one of the operands is unknown. I'm not sure if can represent subtractions soundly since we can't have negative offsets and we have to be able to handle the program flow insensitively

src/main/scala/analysis/Local.scala
sadrabt: yes we can only handle addition precisely, I added a guard to the binary expression case to check for add op. otherwise unsupportedPointerArithmetic should be called

src/main/scala/analysis/Local.scala
l-kent: I think it makes sense if this case is unsupportedPointerArithmeticOperation since arg2 could not be evaluated?

src/main/scala/analysis/SymbolicAccessAnalysis.scala
sadrabt: top is that definition of  Rx is valid (is alive)  and  bottom means that it no longer affects the environment (the definition is dead)

l-kent: Thanks, if you could add that as a comment that would be very helpful

src/main/scala/analysis/DSA.scala
sadrabt: this was due to inlined procedures being added to the queue for analysis. it's now resolved

src/main/scala/analysis/solvers/DSAUnionFindSolver.scala
```diff
@@ -0,0 +1,72 @@
...
+    val rep2 = findWithOffset(t2)._1
+
+    if (rep1 == rep2) return
+
+      (rep1, rep2) match {
+        case (t1: DSAUniTerm, t2: DSAUniTerm) =>
+          mkUnion(t1, t2, offset)
+        case (x, y) =>
+          throw new UnificationFailure(s"Cannot unify $t1 and $t2 (with representatives $x and $y)")
+      }
```
l-kent, line 30: this can just be mkUnion(t1, t2, offset) without the match statement since t1 and t2 are always of type DSAUniTerm

src/main/scala/analysis/Local.scala
sadrabt: I think an issue here are direct calls to procedures that don't have a return block. The ide solver treats these as branches instead of procedure calls. the function entry will be the join of all the direct calls for this procedure. therefore there will be multiple (one for each caller) stack locations associated with R31 in the callee. The simplest way around it is to not propagate stack symbolic addresses through direct calls (this already takes place for direct calls which the ide solver recognizes as a call (ones with a return block)). Otherwise will have to merge stack memory regions of different procedures together. (this is different to merging of stack elements in the same procedures stack which 

I have implemented the former and ran systemtests and the assertion no longer fails. I still think that the assertion may fail in some rare cases but not for any variations of the examples so far. I'm going to try an construct an example where it fails and work on it from there

l-kent: There are definitely other cases where the assertions can fail. See this case: https://github.com/UQ-PAC/BASIL/tree/points-to-analysis-alternate-merge-stackambiguity/examples/stackambiguity

l-kent: In this example, R9 and R10 are assigned different stack pointer offsets depending on branch conditions, which causes them to have multiple `SymbolicAddress`s associated with them after the branching.

sadrabt: refactored the stack recognition functionality to allow for  multiple stack symbolic addresses associated with a register. It first finds all the stack offsets and their maximum accessed size. It then builds the stack regions by merging overlapping accesses to a single Node (each offset in the overlapping access will be a cell of this node).
Offsets which don't have any overlapping accesses remain as separate nodes

sadrabt: okay i think there is still the possibility of an indirect overlapping access though. ie store stack address in heap, load it, use loaded address to do a 128 bit load. Not sure how to handle such cases. Ian iterative approach where we find indirect stack references and redo the analysis?

l-kent: That seems like something that should be handled during the DSA itself - you need to deal with it for non-stack accesses as well. You need some sort of check to find if an access is accessing multiple nodes (at once) that you would then concatenate & merge.

l-kent:
I've merged this and from my testing it seemed to be working fine, but please create issues for any problems that occur.

PR #223, opened 2024-07-09
action: fix SystemTests and upload csv files

katrinafyi:
the previous SystemTests string failed to match any tests, since it was missing the wildcard.


ailrst:
Thanks

PR #225, opened 2024-07-10
Set up versioning from git tags using mill

ailrst:
This adds a version number to BASIL's help text based on `git describe --tags`. It shouldn't cause issues when git isn't available, falling back to whatever is in the VERSION file. The VERSION file stores the version number of the last tagged release.

- Every build it generates a Scala file defining a variable with the content of `git describe --tags` if git is available at build-time, or the contents of the VERSION file if not.

The Scala version file is generated each build is not checked in.

```scala
val BASILVersion = "0.1.2-alpha"
val BASILBuildVersion = "0.1.2-alpha-8-g4d319937"
```

This means the process for making a release is
1. increment number in VERSION file, and commit change
2. add tag: `git tag -a "$(cat VERSION)"`
3. push commit and tags
4. Build the release; doing this earlier will clobber the version file with the previous tag

Alternatively, just adding a git tag will bump the version file on the next build, which will mean the the tagged commit has the old version file, but builds with the correct version when git is present.


l-kent:
This breaks the sbt build, so if the idea is to deprecate the sbt build then that should be done fully.

This doesn't seem to generate the version Scala file automatically upon building with mill, only if the `updateVersion` mill command is run. Is that the intended behaviour? It wasn't clear from what you've written here that I had to do that.

ailrst:
Its supposed to generate the version Scala file automatically when building, and it seems to have compiled successfully using mill in CI with `mill build`.

I wasn't intending to break sbt build, but was deliberately not implementing the whole version updating logic in sbt. Just the Version.scala generation could be implemented in sbt based on the VERSION file, I think that would be enough. Alternatively we could just check in Version.scala, but would probably want to leave it in the gitignore.

l-kent:
Neither `./mill.bat compile` or `./mill.bat run` will create Version.scala for me, only `./mill.bat updateVersion` works.

PR #227, opened 2024-07-17
Function summaries

b-paul:
Generates ensures clauses about the gammas of output variables in procedures using a taint analysis (ish). As an example

```c
int c; // L(c) = true

int f(int a, int b) {
    if (c) {
        return a;
    } else {
        return b;
    }
}
```
we know that the return value of this function could be affected by `a` or `b`, so we say `ensures Gamma_out == Gamma_a || Gamma_out == Gama_b || Gamma_out == true`.

This analysis does not currently scale to run on the `cntlm-new` binary. This is probably because of the time complexity of the IDE solver.


ailrst:
To clarify the approximation in the taint analysis, this runs the taint analysis separately for each each input variable being tainted, and treats taint as a fixed gamma value, so that taint-flow is security-level information-flow. It over-approximates towards high (where?), so for each variable the ensures condition ensures either low, equal to the input variables which taint it = possibly high.

b-paul:
Yeah, the initial implementation would perform a taint analysis for each input/global variable, where taint is the variable's gamma at the start of the procedure. Doing this over every variable tells us precisely which input gammas affect the gamma of each output variable. The current implementation does this more efficiently though, instead working over a powerset lattice to determine the set of input variables which taint an output variable, in a single analysis.

I'm not 100% sure what you mean by over-approximates to high, are you referring to how the results can miss cases where a gamma is definitely high depending on the input (For example `int f(int x, int y) { return x + y; }` and `int f(int x, int y) { return c ? x : y; }` generate the same ensures clause, however the first one could be saying `ensures Gamma_out == Gamma_x && Gamma_y`)?

b-paul:
This change improves the performance of variable dependency analysis by breaking the call graph into its strongly connected components using Tarjan's algorithm, and analyzing those individually. We get a reverse topological sort of the components from Tarjan's algorithm, which allows us to analyze our procedures in an order such that we can get "summaries" of the analysis for all procedures called by a particular procedure, which significantly reduces duplicated work. (Unfortunately the cntlm-new binary has a really big strongly connected component in it which makes this analysis still infeasible to run on the whole binary)

b-paul:
(Credit to @ j-tobler) Ensures clauses might be able to be rewritten as $(\bigwedge_{v \in \text{taint}(out)} \Gamma_v) \Rightarrow \Gamma_{out}$, the gamma of the output variable is lower than the join of the gammas of the tainters.

src/main/scala/analysis/solvers/FixPointSolver.scala
```diff
@@ -247,8 +249,10 @@ trait PushDownWorklistFixpointSolver[N, T, L <: Lattice[T]] extends MapLatticeSo
   def process(n: N): Unit =
     val xn = x(n)
     val y = transfer(n, xn)
+    if (y != xn) {
+      for succ <- outdep(n) do propagate(y, succ)
+    }
```
l-kent, line 254: @yousifpatti didn't you find that doing this caused some bugs?

src/main/scala/util/RunUtils.scala
```diff
@@ -697,6 +733,19 @@ object StaticAnalysis {
...
+    val taintResults = TaintAnalysis(IRProgram, specGlobalAddresses, constPropResult,
+        IRProgram.procedures.foldLeft(Map[CFGPosition, Set[analysis.Taintable]]()) {
+          (m, p) => m + (p -> Set(analysis.UnknownMemory()))
+        }
+      ).analyze()
+
+    config.analysisResultsPath.foreach(s =>
+      writeToFile(printAnalysisResults(IRProgram, taintResults), s"${s}_taint$iteration.txt")
+    )
+    */
```
l-kent, line 747: Does it make sense to add a flag for this if we don't want it enabled by default, or is it not yet working?

b-paul: There wasn't any particular reason for commenting this out apart from just not wanting it to run so adding a flag should be fine

src/main/scala/analysis/SummaryGenerator.scala
```diff
@@ -0,0 +1,186 @@
...
+    varDepsSummaries: Map[Procedure, Map[Taintable, Set[Taintable]]],
+) {
+  val rnaResults = RNATaintableSolver(program, globals, constProp).analyze()
+
+  // TODO should the stack/link/frame pointers propagate taint?
+  val variables: Set[analysis.Taintable] = 0.to(28).map { n =>
+    Register(s"R$n", 64)
+  }.toSet ++ specGlobals.map { g =>
+    analysis.GlobalVariable(dsl.mem, BitVecLiteral(g.address, 64), g.size, g.name)
+  }
```
l-kent, line 93: Only registers R0-R18 and R30 are allowed under the AArch64 standard to change values as the result of a procedure call. This means that there isn't a need to create summaries for the gamma values for R19-R29 and R31 - we already have those.

src/main/scala/analysis/EdgeFunctionLattice.scala
```diff
@@ -62,16 +62,38 @@ class EdgeFunctionLattice[T, L <: Lattice[T]](val valuelattice: L) extends Latti
     def composeWith(e: EdgeFunction[T]): EdgeFunction[T] = this
 
     def joinWith(e: EdgeFunction[T]): EdgeFunction[T] =
-      if (e == this || c == valuelattice.top) this
-      else if (c == valuelattice.bottom) e
```
l-kent, line 66: Why remove this?

src/test/scala/SystemTests.scala
```diff
@@ -162,16 +165,18 @@ trait SystemTests extends AnyFunSuite {
...
     summary("testresult-BAP.csv")
   }
 }
 
 class SystemTestsGTIRB extends SystemTests  {
-  runTests(correctPrograms, correctPath, "correct", true, false)
-  runTests(incorrectPrograms, incorrectPath, "incorrect", false, false)
+  runTests(correctPrograms, correctPath, "correct", true, false, false)
+  runTests(incorrectPrograms, incorrectPath, "incorrect", false, false, false)
+  runTests(correctWithAnalysisPrograms, correctWithAnalysisPath, "correct", true, false, true)
```
l-kent, line 179: This (and the BAP one) should use "correct-analysis" since that's the name of the folder the tests are in.

l-kent:
There seems to be room for optimisation here in relation to which variables this generates summaries for. In some of the tests, it will generate summaries for global variables even when the procedure does not modify memory at all, or only modifies memory because it calls rely (which just says that it does not modify memory).

l-kent:
I'm a bit confused about why it even generates summaries for the globals x and y in the get_two procedure in the function_summary test. get_two doesn't at all access those globals, or even memory at all, so why does the taint analysis say they're tainted?

src/main/scala/util/RunUtils.scala
```diff
@@ -931,6 +981,8 @@ object RunUtils {
         result.reachingDefs,
         ctx.program
       )
+      Logger.info("[!] Generating Function Summaries")
+      modified = modified | IRTransform.generateFunctionSummaries(ctx, ctx.program, result.constPropResult, result.varDepsSummaries)
```
l-kent, line 985: I don't think it makes sense to update `modified` here, at least at the moment. The procedure summaries don't have any impact (and can't have any impact) on any of the other analyses right now so there isn't any need to do another iteration.

src/main/scala/util/RunUtils.scala
l-kent: If the taint analysis gets run as part of generating the function summaries I don't think it makes sense to have it as a separate thing, so this bit could probably just be removed actually?

src/main/scala/analysis/SummaryGenerator.scala
```diff
@@ -0,0 +1,186 @@
...
+        case _ => true
+      }
+    }
+
+    // Use rnaResults to find stack function arguments
+    val tainters = relevantVars.map {
+      v => (v, Set())
+    }.toMap ++ getTainters(procedure, variables ++ rnaResults(procedure.begin) + UnknownMemory()).filter { (variable, taints) =>
+      relevantVars.contains(variable)
+    }
```
l-kent, line 145: A comment explaining what this represents and how the sets of tainters for each variable have been derived would help.

src/main/scala/analysis/SummaryGenerator.scala
```diff
@@ -0,0 +1,186 @@
...
+              // its gamma is low. Note that if a variable was unchanged, it would be tainted by itself.
+              Some(varGamma)
+            } else {
+              // Else, we must consider the variables which have tainted this variable. We can soundly assert
+              // that this variable's gamma must be equal to one of its tainter's gammas, or it is set to
+              // low.
+              //
+              // In a correct procedure, we never have L(x) = true and Gamma_x = false, so we can safely
+              // assume L(x) || Gamma_x == Gamma_x. This means that the gamma of any expression is equal
+              // to the conjunction of the gammas of the variables and the loads in the expression.
```
l-kent, line 166: This doesn't seem necessarily correct to me. Since a variable's security level can change throughout a procedure, its final Gamma value isn't necessarily going to directly relate to the input Gamma values of one of the variables that tainted it?

For example, if we have R0 and R1 as low and R2 as high at the start of a procedure and then the body of the procedure is the following:
```
R0 := R2 + 1
R1 := R0 + 1
```
Would the analysis be able to tell that R1's gamma value actually depends on the initial gamma value of R2, not R1?

l-kent:
The summaries for function_summary3 seem incorrect. For f we get:
```
ensures gamma_load32(Gamma_mem, $x_addr);
ensures Gamma_R1;
ensures Gamma_R0;
```
and for g we get:
```
ensures (gamma_load32(Gamma_mem, $x_addr) || (gamma_load32(Gamma_mem, $x_addr) == old(gamma_load32(Gamma_mem, $x_addr))));
ensures ((Gamma_R0 || (Gamma_R0 == old(Gamma_R1))) || (Gamma_R0 == old(Gamma_R0)));
```
Neither of these procedures modify memory at all, so I'm unclear on why summaries for the global variable x are generated. 

For f, how is it that R1 and R0 are ensured to be low? Shouldn't R1 and R0 both have security levels that are dependent on the input security level of R0? 

These issues may be related to this example using a non-returning call which is not really properly supported at present.

src/main/scala/util/RunUtils.scala
b-paul: Actually yeah these results won't be used for anything you're right, this can be removed

src/main/scala/analysis/SummaryGenerator.scala
b-paul: Oh that's helpful, thanks! I was unaware

src/main/scala/analysis/EdgeFunctionLattice.scala
b-paul: These cases will be accounted for by the join edge, but removing this means we don't need to do call top (which isn't defined on some lattices such as the powerset lattice)

src/main/scala/analysis/SummaryGenerator.scala
b-paul: > Since a variable's security level can change throughout a procedure, its final Gamma value isn't necessarily going to directly relate to the input Gamma values of one of the variables that tainted it?

Maybe I've missed something, but I thought that a change in security level doesn't change the gamma of any variables

> Would the analysis be able to tell that R1's gamma value actually depends on the initial gamma value of R2, not R0?

It will be able to tell, yes. After the first assignment, R0 will depend on {R2}, and after the second assignment, R1 will depend on whatever R0 depends on, so {R2}.

b-paul:
> I'm a bit confused about why it even generates summaries for the globals x and y in the get_two procedure in the function_summary test. get_two doesn't at all access those globals, or even memory at all, so why does the taint analysis say they're tainted?

You're right in that get_two shouldn't need to say anything about x and y. The reason it mentions those is because x and y are initially marked as tainting themselves, so at the end of the procedure they still taint themselves. It shouldn't be too hard to add a check for modifying memory (hopefully) though

> For f, how is it that R1 and R0 are ensured to be low? Shouldn't R1 and R0 both have security levels that are dependent on the input security level of R0?

That is concerning, it certainly shouldn't be doing that :grimacing: I'm confused by how that even passed the test to be perfectly honest... My first guess was that it is something to do with the summaries generated by the strongly connected components, or something like that


I'll hopefully find time to work on this soon

src/main/scala/analysis/SummaryGenerator.scala
l-kent: I was using 'security level' and 'gamma' interchangeably there, not referring to the security policy for a variable, sorry for the confusion.

l-kent:
The test passes because the 'assume false' (generated by the non-returning call to g() in f()) allows for any post condition of f() to be verified. That's part of what I mean by non-returning calls not being properly supported at the moment.

src/main/scala/analysis/solvers/FixPointSolver.scala
yousifpatti: This caused the results of constantprop analysis to not be propagated properly. The effects may not be noticeable because evaluate expression may return TOP and you deal with it the same way. The best way to check is by checking the dot file of constant prop to see if values are being propagated.

l-kent: Thanks, I remember that made the analyses resolve some indirect calls incorrectly. It would be good if we could figure out how to make this optimisation work but it should be reverted for now.

src/main/scala/analysis/solvers/FixPointSolver.scala
```diff
@@ -225,10 +225,12 @@ trait PushDownWorklistFixPointFunctions[N, T, L <: Lattice[T]] extends LinkedHas
...
-    val t = lattice.sublattice.lub(xm, y)
-    if (t != xm) {
-      add(m)
-      x += m -> t
+    if (xm != y) {
+      val t = lattice.sublattice.lub(xm, y)
+      if (t != xm) {
+        add(m)
+        x += m -> t
+      }
```
l-kent, line 233: I think this will require close inspection of its impact, given the issue with the similar change to the process method.

l-kent: This seems to be fine in practice, and makes sense theoretically as long as the lub method isn't doing anything it shouldn't.

l-kent:
I'm just going to clean this up and merge it in, putting it behind a flag for now.

PR #231, opened 2024-08-13
Adding Memory Regions to IR

yousifpatti:
Before merging there are two main issues

- [x] ~~Some initial regions of a function are ignored (they are not collected by MRA) and thus are not renamed. This may have been fixed in the MRA file of the other branch but has other dependencies that needs merging.~~

- [x] Some Data section regions that are in initial memory and read only memory are not found. (Those are not loaded by the ReadELFLoader and cause this issue as MMM relies on results from the loader. Notable mentions for "arrays" example:
[WARN]   No region found for memory section 0 [transformMemorySections@RegionInjector.scala:297]
[WARN]   No region found for memory section 0 [transformMemorySections@RegionInjector.scala:297]
[WARN]   No region found for memory section 0 [transformMemorySections@RegionInjector.scala:297]
[WARN]   No region found for memory section 0 [transformMemorySections@RegionInjector.scala:297]
[WARN]   No region found for memory section 568 [transformMemorySections@RegionInjector.scala:297]
[WARN]   No region found for memory section 596 [transformMemorySections@RegionInjector.scala:297]
[WARN]   No region found for memory section 632 [transformMemorySections@RegionInjector.scala:297]
[WARN]   No region found for memory section 664 [transformMemorySections@RegionInjector.scala:297]
[WARN]   No region found for memory section 696 [transformMemorySections@RegionInjector.scala:297]
[WARN]   No region found for memory section 960 [transformMemorySections@RegionInjector.scala:297]
[WARN]   No region found for memory section 1148 [transformMemorySections@RegionInjector.scala:297]
[WARN]   No region found for memory section 1176 [transformMemorySections@RegionInjector.scala:297]
[WARN]   No region found for memory section 1240 [transformMemorySections@RegionInjector.scala:297]
[WARN]   No region found for memory section 1528 [transformMemorySections@RegionInjector.scala:297]

Anything bigger than these offsets is resolved. These offsets dont seem to be loaded with externalFunctions or globalOffsets


l-kent:
Memory sections in the initial memory/read only memory that are not actually accessed in the program don't need to be given their own region, it's fine if they just use the default `mem` region for now. Eventually we want to go through the initial memory definitions per-procedure so that any memory definitions that relate to regions that aren't in the procedure are not included.

l-kent:
> This may have been fixed in the MRA file of the other branch but has other dependencies that needs merging.

Which other branch are you talking about?

yousifpatti:
> > This may have been fixed in the MRA file of the other branch but has other dependencies that needs merging.
> 
> Which other branch are you talking about?

I am talking about yousif-vsa branch were I made changes to MRA. This issue has been resolved tho in my latest commit here so no need to worry about that anymore.

yousifpatti:
@l-kent In that case can we look into merging the changes please?

src/main/scala/util/RunUtils.scala
```diff
@@ -25,13 +26,15 @@ import org.antlr.v4.runtime.BailErrorStrategy
...
 import util.Logger
+
 import java.util.Base64
 import spray.json.DefaultJsonProtocol.*
 import util.intrusive_list.IntrusiveList
 import analysis.CfgCommandNode
 
 import scala.annotation.tailrec
 import scala.collection.mutable
+import scala.sys.exit
```
l-kent, line 37: Is this used for anything?

yousifpatti:
The last commit should fix the issue with over approximating to "_abort_"
The issue was that abort was the last region loaded and thus any offsets after were approximated to "_abort_".
The other global offsets were missing "_globalAddresses_" variable not passed to MMM, so I have just added those offsets and it is working properly,

src/main/scala/analysis/RegionInjector.scala
```diff
@@ -0,0 +1,321 @@
...
+            case call: DirectCall => // ignore DirectCall
+            case call: IndirectCall => // ignore IndirectCall
+    case _ => // ignore other kinds of nodes
+  }
+
+  def transformMemorySections(memorySegment: ArrayBuffer[MemorySection]): ArrayBuffer[MemorySection] = {
+    val newArrayBuffer = ArrayBuffer.empty[MemorySection]
+    for (elem <- memorySegment) {
+      elem match {
+        case mem: MemorySection =>
```
l-kent, line 307: there isn't a need to match this, it's always going to be a MemorySection?

src/main/scala/analysis/RegionInjector.scala
```diff
@@ -0,0 +1,321 @@
...
+          val regions = mmm.findDataObject(mem.address)
```
l-kent, line 308: This doesn't work at present. `MemorySection.address` is just the address of the start of the memory section - for example `.plt` `.got` and `.data` are all MemorySections, and their contents are just listed as bytes, which is very crude. We want to extract the bytes relevant to any regions that the memory region analysis has determined are accessed, which requires much more work than just attempting to match the addresses like this.

Currently, all this method does is remove some data sections entirely, even if they contain addresses that we care about.

l-kent:
Currently, the initial memory pre/post-conditions are not generated with the identified regions. An example of this is the test correct/basic_function_call_reader/clang_pic. The following snippets are from the output currently produced for this test:

```
procedure main();
modifies ...
  free requires (memory_load64_le(mem, 69568bv64) == 69688bv64);
```

```
R8, Gamma_R8 := 65536bv64, true;
call rely();
R8, Gamma_R8 := memory_load64_le(x, bvadd64(R8, 4032bv64)), (gamma_load64(Gamma_x, bvadd64(R8, 4032bv64)) || L(x, bvadd64(R8, 4032bv64)));
```

In the second snippet, the address 69568 (65536 + 4032) is loaded from the region `x`. However, in the precondition created from initialised memory in the first snippet, the default region `mem` is used for the address 69568. We want the regions to be aligned, so that the same region is used for both.

l-kent:
The rely() procedures created from the user's specification do not include the new regions. An example of this is the test correct/basicassign_gamma0/clang. The following snippets are all from the output produced for it:

```
axiom ($secret_addr == 69684bv64);
```

```
procedure {:extern} rely();
  modifies Gamma_mem, mem;
  ensures (forall i: bv64 :: (((mem[i] == old(mem[i])) ==> (Gamma_mem[i] == old(Gamma_mem[i])))));
  ensures (memory_load32_le(mem, $secret_addr) == old(memory_load32_le(mem, $secret_addr)));
```

```
R8, Gamma_R8 := 69632bv64, true;
call rely();
R8, Gamma_R8 := zero_extend32_32(memory_load32_le(secret, bvadd64(R8, 52bv64))), (gamma_load32(Gamma_secret, bvadd64(R8, 52bv64)) || L(secret, bvadd64(R8, 52bv64)));
```

In the third snippet, the address 69684 (69632 + 52) is loaded from the region `secret`, but in the rely() procedure, the address 69684 is loaded from the default region `mem`. We want this to be replaced in the rely procedure so it matches the identified region.

l-kent:
The test correct/initialisation/gcc demonstrates three issues - not using identified regions in the generated require/ensures clauses from the user specifications, incorrectly removing relevant sections of initialised memory, and not using the identified regions in initialised memory requires/ensures clauses. I've already discussed the latter so I'll detail the first two for this test.

On main, the following snippet is produced for this test:

```
procedure main();
  modifies ...
  free requires (memory_load64_le(mem, 69632bv64) == 0bv64);
  free requires (memory_load64_le(mem, 69640bv64) == 69640bv64);
  free requires (memory_load64_le(mem, 69648bv64) == 416611827717bv64);
  free requires (memory_load64_le(mem, 69656bv64) == 68719476735bv64);
  free requires (memory_load64_le(mem, 69664bv64) == 8589934593bv64);
```

These requires clauses are all removed from the output when using this branch's memory region analysis. However, the addresses 69648, 69656, 69664, 69668 and 69652 are all accessed, so we want to keep the initialisations for 69648, 69656 and 69664 around. We want the regions in the requires clause to match as well, so that would require splitting up the initialisations further - the region y would need the address 69652 initialised with a 32-bit access, and the region x would need the address 69648 initialised with a 32-bit access, instead of the 64-bit initialisation covering both as-is.

```
R0, Gamma_R0 := 69632bv64, true;
R0, Gamma_R0 := bvadd64(R0, 16bv64), Gamma_R0;
call rely();
R0, Gamma_R0 := zero_extend32_32(memory_load32_le(x, R0)), (gamma_load32(Gamma_x, R0) || L(x, R0));
```
```
R0, Gamma_R0 := 69632bv64, true;
R0, Gamma_R0 := bvadd64(R0, 24bv64), Gamma_R0;
call rely();
R0, Gamma_R0 := memory_load64_le(z, R0), (gamma_load64(Gamma_z, R0) || L(z, R0));
```
```
R0, Gamma_R0 := 69632bv64, true;
R0, Gamma_R0 := bvadd64(R0, 32bv64), Gamma_R0;
call rely();
R0, Gamma_R0 := zero_extend32_32(memory_load32_le(a, R0)), (gamma_load32(Gamma_a, R0) || L(a, R0));
```
```
R0, Gamma_R0 := 69632bv64, true;
R0, Gamma_R0 := bvadd64(R0, 32bv64), Gamma_R0;
call rely();
assert (L(a, bvadd64(R0, 4bv64)) ==> Gamma_R1);
a, Gamma_a := memory_store32_le(a, bvadd64(R0, 4bv64), R1[32:0]), gamma_store32(Gamma_a, bvadd64(R0, 4bv64), Gamma_R1);
```
```
R0, Gamma_R0 := 69632bv64, true;
R0, Gamma_R0 := bvadd64(R0, 20bv64), Gamma_R0;
call rely();
R0, Gamma_R0 := zero_extend56_8(memory_load8_le(y, R0)), (gamma_load8(Gamma_y, R0) || L(y, R0));
```

The output also includes the following ensures clauses in the procedure main():
``` 
ensures (memory_load32_le(mem, $x_addr) == 6bv32);
ensures (memory_load32_le(mem, bvadd64($a_addr, 4bv64)) == 4bv32);
ensures (memory_load32_le(mem, bvadd64($a_addr, 0bv64)) == 1bv32);
```
The issue is that mem is not replaced with identified regions for these, but we want it to.

l-kent:
Currently, the memory regions for pointers and pointers-to-pointers are unnecessarily combined. This doesn't cause any issues at present but it is unnecessary. 

An example of this is the output for the test case correct/basic_arrays_read/clang_pic:
```
R8, Gamma_R8 := 65536bv64, true;
call rely();
R8, Gamma_R8 := memory_load64_le(arr, bvadd64(R8, 4056bv64)), (gamma_load64(Gamma_arr, bvadd64(R8, 4056bv64)) || L(arr, bvadd64(R8, 4056bv64)));
call rely();
assert (L(arr, R8) ==> true);
arr, Gamma_arr := memory_store32_le(arr, R8, 0bv32), gamma_store32(Gamma_arr, R8, true);
```

The important context here is that the global variable `arr` is located at address 69638, and this address is stored in memory at address 69592. This means that the first memory access loads 69638 (the address of global variable `arr`) from address 69592 (65536 + 4056) to R8, and the second memory access writes 0 to the global variable `arr`.

Both accesses are given the region `arr` even though they are completely distinct memory locations, and it would be ideal for the analysis to be able to give these separate regions.

l-kent:
This PR causes some of the IndirectCallsTests to fail - for some of those tests, the indirect calls get incorrectly resolved to the global variable `x` or `_ITM_registerTMCloneTable`.

Indirect call resolution also fails for the correct/jumptable/clang (including variants) tests and for the correct/functionpointer/gcc_O2 and /clang_O2 tests. 

The correct/jumptable/clang test is a particularly interesting example, and stack region identification is currently incorrect for it too. These are all the stack accesses produced in main() (in order):

```
#5, Gamma_#5 := bvadd64(R31, 64bv64), Gamma_R31;
stack_25, Gamma_stack_25 := memory_store64_le(stack_25, #5, R29), gamma_store64(Gamma_stack_25, #5, Gamma_R29); // R31 + 64 (64-bit access)

stack_25, Gamma_stack_25 := memory_store64_le(stack_25, bvadd64(#5, 8bv64), R30), gamma_store64(Gamma_stack_25, bvadd64(#5, 8bv64), Gamma_R30); // R31 + 72 (64-bit access)

R29, Gamma_R29 := bvadd64(R31, 64bv64), Gamma_R31;

stack_15, Gamma_stack_15 := memory_store32_le(stack_15, bvadd64(R31, 12bv64), R8[32:0]), gamma_store32(Gamma_stack_15, bvadd64(R31, 12bv64), Gamma_R8); // R31 + 12 (32-bit access)

stack_28, Gamma_stack_28 := memory_store32_le(stack_28, bvadd64(R29, 18446744073709551612bv64), 0bv32), gamma_store32(Gamma_stack_28, bvadd64(R29, 18446744073709551612bv64), true); // R31 + 60 (32-bit access)

stack_28, Gamma_stack_28 := memory_store32_le(stack_28, bvadd64(R29, 18446744073709551608bv64), R0[32:0]), gamma_store32(Gamma_stack_28, bvadd64(R29, 18446744073709551608bv64), Gamma_R0); // R31 + 56 (32-bit access)

stack_21, Gamma_stack_21 := memory_store64_le(stack_21, bvadd64(R29, 18446744073709551600bv64), R1), gamma_store64(Gamma_stack_21, bvadd64(R29, 18446744073709551600bv64), Gamma_R1); // R31 + 48 (64-bit access)

stack_21, Gamma_stack_21 := memory_store128_le(stack_21, bvadd64(R31, 16bv64), V0), gamma_store128(Gamma_stack_21, bvadd64(R31, 16bv64), Gamma_V0); R31 + 16 (128-bit access)

stack_25, Gamma_stack_25 := memory_store64_le(stack_25, bvadd64(R31, 32bv64), R8), gamma_store64(Gamma_stack_25, bvadd64(R31, 32bv64), Gamma_R8); R31 + 32 (64-bit access)

R8, Gamma_R8 := memory_load64_le(stack_21, bvadd64(R31, 16bv64)), gamma_load64(Gamma_stack_21, bvadd64(R31, 16bv64)); R31 + 16 (64-bit access)

R8, Gamma_R8 := memory_load64_le(stack_28, bvadd64(R31, 24bv64)), gamma_load64(Gamma_stack_28, bvadd64(R31, 24bv64)); R31 + 24 (64-bit access)

R8, Gamma_R8 := memory_load64_le(stack_25, bvadd64(R31, 32bv64)), gamma_load64(Gamma_stack_25, bvadd64(R31, 32bv64)); R31 + 32 (64-bit access)

R0, Gamma_R0 := zero_extend32_32(memory_load32_le(stack_15, bvadd64(R31, 12bv64))), gamma_load32(Gamma_stack_15, bvadd64(R31, 12bv64)); R31 + 12 (32-bit access)
#6, Gamma_#6 := bvadd64(R31, 64bv64), Gamma_R31;
R29, Gamma_R29 := memory_load64_le(stack_25, #6), gamma_load64(Gamma_stack_25, #6); R31 + 64 (64-bit access)
R30, Gamma_R30 := memory_load64_le(stack_25, bvadd64(#6, 8bv64)), gamma_load64(Gamma_stack_25, bvadd64(#6, 8bv64)); R31 + 72 (64-bit access)
```

Each stack region is associated with the following accesses:

stack_15: R31 + 12 (32-bit access)
stack_21: R31 + 48 (64-bit access), R31 + 16 (128-bit access), R31 + 16 (64-bit access)
stack_25: R31 + 64 (64-bit access), R31 + 72 (64-bit access), R31 + 32 (64-bit access)
stack_28: R31 + 60 (32-bit access), R31 + 56 (32-bit access), R31 + 24 (64-bit access)

It is incorrect that R31 + 16 (128-bit access) and R31 + 24 (64-bit access) are in different regions, despite being overlapping accesses - the analysis needs to consider the size of accesses as well as their address. 

While the way the other regions are combined does not cause problems, it is imprecise and it would be ideal to fix it so that distinct accesses have distinct regions.

l-kent:
Another issue is the regions in assertions, specifically the assertions that check if writes to memory are secure. 

These are updates to the global variables z and x, respectively from the test correct/secret_write/clang, with annotations:

```
assert (L(z, bvadd64(R9, 52bv64)) ==> true); // checking that the memory store doesn't violate security properties
z_old := memory_load32_le(z, $z_addr); // getting current value of global variable z for comparison after the store
Gamma_x_old := (gamma_load32(Gamma_z, $x_addr) || L(z, $x_addr)); // getting current value of security level of global variable x, but using the wrong region
z, Gamma_z := memory_store32_le(z, bvadd64(R9, 52bv64), 0bv32), gamma_store32(Gamma_z, bvadd64(R9, 52bv64), true); // writing to global variable z
assert ((bvadd64(R9, 52bv64) == $z_addr) ==> (L(z, $x_addr) ==> Gamma_x_old)); // checking that if this memory store was to global variable z then it hasn't violated the security policy for global variable x, which depends on global variable z, but using the wrong region as input to the L function
assert bvsge32(memory_load32_le(mem, $z_addr), z_old); // checking the guarantee, that the global variable z is only increasing, but using the wrong region
```

```
assert (L(x, bvadd64(R8, 60bv64)) ==> Gamma_R10);
z_old := memory_load32_le(x, $z_addr); // storing current value of global variable z for comparison after the store, but using incorrect region 
Gamma_x_old := (gamma_load32(Gamma_x, $x_addr) || L(x, $x_addr));  // getting current value of security level of global variable x
x, Gamma_x := memory_store32_le(x, bvadd64(R8, 60bv64), R10[32:0]), gamma_store32(Gamma_x, bvadd64(R8, 60bv64), Gamma_R10); // writing to global variable x
assert ((bvadd64(R8, 60bv64) == $z_addr) ==> (L(x, $x_addr) ==> Gamma_x_old)); // checking that if this memory store was to global variable z then it hasn't violated the security policy for global variable x, which depends on global variable z. This assertion is now redundant since we have determined that this memory store was to the region x, which does not contain the global variable z which this assertion is relevant for. 
assert bvsge32(memory_load32_le(mem, $z_addr), z_old); // checking the guarantee, that the global variable z is only increasing, but using the wrong region. This assertion is now completely redundant since we have determined that this is a write to the region x which does not contain the global variable z which this assertion is relevant for.
```

There are number of issues we need to address here. Most straight-forwardly - there is still a reference to `mem` in the last assertion generated, which checks the user-specified guarantee. We need to make sure all accesses to global variables use the correct region for each global variable.

The more difficult issue is that we need to redesign this secure update check to take into account that not all global variables share the same region now, which means many checks can be removed. 

We want to change the output so that it now looks something like this:

```
assert (L(z, bvadd64(R9, 52bv64)) ==> true);
z_old := memory_load32_le(z, $z_addr);
Gamma_x_old := (gamma_load32(Gamma_x, $x_addr) || L(x, $x_addr));
z, Gamma_z := memory_store32_le(z, bvadd64(R9, 52bv64), 0bv32), gamma_store32(Gamma_z, bvadd64(R9, 52bv64), true);
assert ((bvadd64(R9, 52bv64) == $z_addr) ==> (L(x, $x_addr) ==> Gamma_x_old));
assert bvsge32(memory_load32_le(z, $z_addr), z_old);
```

```
assert (L(x, bvadd64(R8, 60bv64)) ==> Gamma_R10);
x, Gamma_x := memory_store32_le(x, bvadd64(R8, 60bv64), R10[32:0]), gamma_store32(Gamma_x, bvadd64(R8, 60bv64), Gamma_R10); // writing to global variable x
```

The differences are:
- the correct regions are used in relation to global variables
- assertions for guarantees are only generated if a global variable referred to in a guarantee is contained in a region accessed by a memory store
- secure update assertions are only generated for a control variable (in this case, z is the control variable) if a region containing that control variable is accessed by a memory store

It may also make sense to split the L function so that each region has its own L function? I'm not sure if this matters at all either way though.

l-kent:
I think that should be pretty much everything. I realise this is a lot. I think the most important issues to focus on at first are the ones in this comment: https://github.com/UQ-PAC/BASIL/pull/231#issuecomment-2299908156 since those are issues with the analysis itself, rather than just more work that needs to be done with the analysis results.

src/main/scala/analysis/MemoryModelMap.scala
```diff
@@ -254,6 +405,29 @@ class MemoryModelMap {
...
+    uf.find(region.asInstanceOf[MemoryRegion]).asInstanceOf[StackRegion]
+  }
+
+  private def returnRegion(region: DataRegion): DataRegion = {
+    uf.find(region.asInstanceOf[MemoryRegion]).asInstanceOf[DataRegion]
+  }
+
+  private def returnRegion(region: HeapRegion): HeapRegion = {
+    uf.find(region.asInstanceOf[MemoryRegion]).asInstanceOf[HeapRegion]
+  }
```
l-kent, line 430: If all the region subtypes are treated completely separately as this implies, shouldn't each region subtype have its own UnionFind to enforce this?

l-kent:
Another issue: the stack identification doesn't seem to work correctly when heap pointers are involved.

This is from test/correct/malloc_with_local/gcc/malloc_with_local.bir, with annotations:

```
0000036d: call @malloc with return %0000036f // call malloc, putting fresh heap pointer in R0

0000036f:
00000375: mem := mem with [R31 + 0x20, el]:u64 <- R0 // store fresh heap pointer on stack at R31 + 0x20
0000037a: R0 := 4
0000037f: R30 := 0x830
00000381: call @malloc with return %00000383 // call malloc, putting a second fresh heap pointer in R0

00000383:
00000389: mem := mem with [R31 + 0x28, el]:u64 <- R0 // store second heap pointer on stack at R31 + 0x28
0000038e: R0 := 0xA
00000396: mem := mem with [R31 + 0x1C, el]:u32 <- 31:0[R0] 
0000039d: R0 := mem[R31 + 0x20, el]:u64 // load first heap pointer from stack to R0
000003a2: R1 := 0x41
000003aa: mem := mem with [R0] <- 7:0[R1] // write 0x41 to first heap location, which R0 points to
000003b1: R0 := mem[R31 + 0x28, el]:u64 // load second heap pointer from stack to R0
000003b6: R1 := 0x2A
000003be: mem := mem with [R0, el]:u32 <- 31:0[R1] // write 0x2A to second heap location, which R0 points to
000003c5: R0 := mem[R31 + 0x20, el]:u64 // load first heap pointer from stack to R0
000003cc: R0 := pad:64[mem[R0]] // load value contained in first heap location to R0
```

The corresponding Boogie produced is the following (with annotations and cleaned up a little):

```
  // call malloc, putting fresh heap pointer in R0
  call malloc();  
  goto l0000036f;
l0000036f:
  // store fresh heap pointer on stack at R31 + 32 (== 0x20)
  stack_15, Gamma_stack_15 := memory_store64_le(stack_15, bvadd64(R31, 32bv64), R0), gamma_store64(Gamma_stack_15, bvadd64(R31, 32bv64), Gamma_R0);
  R0, Gamma_R0 := 4bv64, true;
  R30, Gamma_R30 := 2096bv64, true;
  // call malloc, putting a second fresh heap pointer in R0
  call malloc();
  goto l00000383;
l00000383:
  // store second heap pointer on stack at R31 + 40 (== 0x28) 
  stack_18, Gamma_stack_18 := memory_store64_le(stack_18, bvadd64(R31, 40bv64), R0), gamma_store64(Gamma_stack_18, bvadd64(R31, 40bv64), Gamma_R0); 
  R0, Gamma_R0 := 10bv64, true;
  stack_20, Gamma_stack_20 := memory_store32_le(stack_20, bvadd64(R31, 28bv64), R0[32:0]), gamma_store32(Gamma_stack_20, bvadd64(R31, 28bv64), Gamma_R0);
  // load first heap pointer from stack to R0
  R0, Gamma_R0 := memory_load64_le(stack_15, bvadd64(R31, 32bv64)), gamma_load64(Gamma_stack_15, bvadd64(R31, 32bv64)); 
  R1, Gamma_R1 := 65bv64, true;
  // write 65 (== 0x41) to first heap location, which R0 points to - incorrectly uses region 'stack_15' instead of the first heap region
  stack_15, Gamma_stack_15 := memory_store8_le(stack_15, R0, R1[8:0]), gamma_store8(Gamma_stack_15, R0, Gamma_R1);
  // load second heap pointer from stack to R0
  R0, Gamma_R0 := memory_load64_le(stack_18, bvadd64(R31, 40bv64)), gamma_load64(Gamma_stack_18, bvadd64(R31, 40bv64));
  R1, Gamma_R1 := 42bv64, true;
  // write 42 (== 0x2A) to second heap location, which R0 points to  - incorrectly uses region 'stack_18' instead of the second heap region
  stack_18, Gamma_stack_18 := memory_store32_le(stack_18, R0, R1[32:0]), gamma_store32(Gamma_stack_18, R0, Gamma_R1);
  // load first heap pointer from stack to R0
  R0, Gamma_R0 := memory_load64_le(stack_15, bvadd64(R31, 32bv64)), gamma_load64(Gamma_stack_15, bvadd64(R31, 32bv64));
  // load value contained in first heap location to R0 - incorrectly uses region 'stack_15' instead of the first heap region
  R0, Gamma_R0 := zero_extend56_8(memory_load8_le(stack_15, R0)), gamma_load8(Gamma_stack_15, R0);
```

The big issue is that accesses to the heap are incorrectly treated as accesses to the stack regions in which the heap pointers are stored. Every heap allocation should have its own region.

l-kent:
If it would help, I am happy to work on all the bits that relate to getting the new regions to be properly used in the Boogie output - everything except the bugs in the analyses. I have a good understanding of those parts of the code so that would speed things up.

yousifpatti:
> If it would help, I am happy to work on all the bits that relate to getting the new regions to be properly used in the Boogie output - everything except the bugs in the analyses. I have a good understanding of those parts of the code so that would speed things up.

That would be good if you could please. I will have a look at the bugs in the analyses. Thanks Liam

src/main/scala/analysis/RegionInjector.scala
yousifpatti: You are right. I think this is left over from the changes to stmts and exprs.

src/main/scala/util/RunUtils.scala
yousifpatti: Used for debugging only. I will remove.

src/main/scala/analysis/RegionInjector.scala
yousifpatti: What is the relationship between those small 8 byte addresses and the memory section? Are they loaded from the relf file?

l-kent: The MemorySections are loaded from the .adt or .gts file and each represent a single section of the binary - such as the `.data` section. The `address` of a MemorySection is the address of the start of the section, and each byte in the array of bytes is located at a subsequent address. So MemorySection.bytes[0] is the 8-bit value stored at address MemorySection.address, MemorySection.bytes[1] is the 8-bit value stored at address MemorySection.address + 1, and so on.

I will handle this part though, since this relates to fixing the Boogie output.

src/main/scala/analysis/RegionInjector.scala
```diff
@@ -0,0 +1,317 @@
...
+        case dataRegion: DataRegion =>
+          return SharedMemory(dataRegion.regionIdentifier, mem.addressSize, mem.valueSize)
+        case _ =>
+      }
+    } else {
+      Logger.warn(s"Mem CMD is: ${cmd}")
+      Logger.warn(s"No region found for mem")
+    }
+    mem
+  }
```
l-kent, line 276: Merging regions at the same time as we replace the `mem` variables in the IR will not work because it means that earlier variables that are replaced will not take into account any later merges.

For instance, if we have a program that the analysis determines accesses regions in the following order:
```
access region A
access region B
access region A or B
```
then when we encounter the access to region A or B, they should be merged into a single region, but the earlier accesses to A and B will not be updated to take into account that A and B are now the same region.

There needs to be two separate passes - a pass where regions are merged, and a pass where the memory variables are replaced.

l-kent:
There's a regression compared to main - the input jumptable2\gcc_pic\jumptable2.adt is no longer resolved correctly - the first call should be resolved to add_two() not "jump_table()"

This is caused by commits 20f9cf65cf8c309e510906911f8b7e5601fba1fc and ac906b90f89ff60bf40c86f4d3150887a6899b87

l-kent:
When running IndirectCallTests, the following tests do not pass in main:
- functionpointer/clang_O2:BAP
- functionpointer/clang_O2:GTIRB
- functionpointer/gcc_O2:BAP
- indirect_call_outparam/clang:BAP
- indirect_call_outparam/clang:GTIRB
- indirect_call_outparam/gcc:BAP
- indirect_call_outparam/gcc:GTIRB
- jumptable/clang:BAP
- jumptable/clang:GTIRB
- jumptable/gcc:BAP (only due to verification failure - indirect calls are resolved correctly)
- jumptable/gcc:GTIRB (only due to verification failure - indirect calls are resolved correctly)
- jumptable3/clang:GTIRB
- jumptable3/clang_O2:GTIRB
- switch2/clang:GTIRB

jumptable2/gcc_pic:BAP is only an issue for this branch and should be fixed

These require more work to handle indirect calls to blocks:
- jumptable3/clang:GTIRB
- jumptable3/clang_O2:GTIRB
- switch2/clang:GTIRB

These require more sophisticated interprocedural analysis so much more work is required so I'm not expecting them to pass yet:
- indirect_call_outparam/clang:BAP
- indirect_call_outparam/clang:GTIRB
- indirect_call_outparam/gcc:BAP
- indirect_call_outparam/gcc:GTIRB

These may require more work but ideally we would be able to handle them right now:
- functionpointer/gcc_O2:BAP
- functionpointer/clang_O2:GTIRB
- functionpointer/clang_O2:BAP
- jumptable/clang:BAP
- jumptable/clang:GTIRB

l-kent:
Currently there is a problem where the analysis doesn't seem to resolve regions for all accesses. For the functionpointer/clang:BAP test, for instance, the following is part of the bpl output in main:
```
    R29, Gamma_R29 := bvadd64(R31, 32bv64), Gamma_R31;
    call rely();
    assert (L(mem, bvadd64(R29, 18446744073709551612bv64)) ==> true);
    mem, Gamma_mem := memory_store32_le(mem, bvadd64(R29, 18446744073709551612bv64), 0bv32), gamma_store32(Gamma_mem, bvadd64(R29, 18446744073709551612bv64), true);
    assume {:captureState "%000003ee"} true;
    call rely();
    assert (L(mem, bvadd64(R29, 18446744073709551608bv64)) ==> Gamma_R0);
    mem, Gamma_mem := memory_store32_le(mem, bvadd64(R29, 18446744073709551608bv64), R0[32:0]), gamma_store32(Gamma_mem, bvadd64(R29, 18446744073709551608bv64), Gamma_R0);
    assume {:captureState "%000003f6"} true;
```
The accesses to `mem` here should clearly be stack accesses, but they aren't replaced by the RegionInjector. The idea is that every access should have an associated region.

src/main/scala/ir/IRCursor.scala
```diff
@@ -148,27 +148,21 @@ object IntraProcBlockIRCursor extends IntraProcBlockIRCursor
...
           }
-
-        }
-        case c: Procedure                              => c.incomingCalls().toSet.asInstanceOf[Set[CFGPosition]]
-        case _                                         => Set.empty
-      )
+        case c: Procedure => c.incomingCalls().toSet.asInstanceOf[Set[CFGPosition]]
+        case _            => IntraProcIRCursor.pred(pos)
   }
 }
```
l-kent, line 167: Changing the behaviour of InterProcIRCursor may break other analyses that rely on it.

src/main/scala/analysis/Analysis.scala
```diff
@@ -22,7 +22,7 @@ trait Analysis[+R]:
 
 /** Base class for value analysis with simple (non-lifted) lattice.
   */
-trait ConstantPropagation(val program: Program) {
+trait ConstantPropagation(val program: Program, val assumeR31: Boolean) {
```
l-kent, line 25: What is the purpose of assumeR31? It doesn't seem to actually be used anywhere, and seems like it would just confuse anything involving the stack if it were to be used?

src/main/scala/analysis/MemoryModelMap.scala
```diff
@@ -22,7 +22,7 @@ class MemoryModelMap {
   private val MAX_BIGINT: BigInt = BigInt(Long.MaxValue)
   private val contextStack = mutable.Stack.empty[String]
   private val sharedContextStack = mutable.Stack.empty[List[StackRegion]]
-  private val localStacks = mutable.Map[String, List[StackRegion]]()
+  private val localStacks = mutable.Map[String, List[StackRegion]]().withDefaultValue(List.empty)
```
l-kent, line 25: It's really best to avoid using withDefaultValue where possible since it has some strange properties that can cause unintuitive bugs.

src/main/scala/analysis/MemoryRegionAnalysis.scala
```diff
@@ -15,8 +15,9 @@ trait MemoryRegionAnalysis(val program: Program,
                            val constantProp: Map[CFGPosition, Map[Variable, FlatElement[BitVecLiteral]]],
                            val ANRResult: Map[CFGPosition, Set[Variable]],
                            val RNAResult: Map[CFGPosition, Set[Variable]],
-                           val regionAccesses: Map[CFGPosition, Map[RegisterVariableWrapper, FlatElement[Expr]]],
-                           reachingDefs: Map[CFGPosition, (Map[Variable, Set[Assign]], Map[Variable, Set[Assign]])]) {
+                           val reachingDefs: Map[CFGPosition, (Map[Variable, Set[Assign]], Map[Variable, Set[Assign]])],
+                           val offSetApproximation: Boolean = false,
```
l-kent, line 19: What does offsetApproximation do? You should comment this

yousifpatti: Testing. It will be removed

src/main/scala/ir/IRCursor.scala
yousifpatti: It was not producing the expected behaviour. It was returning the edge from the call to the after call which is not correct

src/main/scala/analysis/MemoryRegionAnalysis.scala
```diff
@@ -195,26 +235,93 @@ trait MemoryRegionAnalysis(val program: Program,
 //          }
           if (directCall.target.name == "malloc") {
             evaluateExpression(mallocVariable, constantProp(n)) match {
-              case Some(b: BitVecLiteral) => regionLattice.lub(s, Set(HeapRegion(nextMallocCount(), b, IRWalk.procedure(n))))
+              case Some(b: BitVecLiteral) =>
+                val negB = if isNegative(b) then b.value - BigInt(2).pow(b.size) else b.value
+                val newHeapRegion = HeapRegion(nextMallocCount(), negB, IRWalk.procedure(n))
+                addReturnHeap(directCall, newHeapRegion)
+                regionLattice.lub(s, Set(newHeapRegion))
               case None => s
```
l-kent, line 243: The heap identification currently crashes for a lot of the tests that have malloc in them. correct/malloc_with_local2/clang:BAP is one of these.

This appears to be because of this section of code - heap regions are not created unless the size parameter of malloc can be determined by constant propagation. We need to be able to handle cases where a heap region's size is not known (at least initially). In the example correct/malloc_with_local2/clang:BAP constant propagation fails to determine the size because the size parameter is put on the stack and then loaded from the stack.

src/main/scala/analysis/AVLTree.scala
l-kent, line 1: This doesn't seem to actually be used for anything?

src/main/scala/analysis/BACKUPLAttice
l-kent, line 1: Please don't commit unnecessary files like this.

src/main/scala/analysis/ActualVSA.scala
l-kent, line 1: Please don't commit unnecessary files like this

src/main/scala/analysis/GlobalRegionAnalysis.scala
l-kent, line 1: What does this analysis actually do & why do we need it to be separate from the MRA?

src/main/scala/analysis/IrreducibleLoops.scala
```diff
@@ -24,7 +24,7 @@ private def label(p: CFGPosition) = {
  *
  */
 case class LoopEdge(from: CFGPosition, to: CFGPosition) {
-  override def toString: String = s"(${label(from)}, ${label(to)})"
+  override def toString: String = s"(${from}, ${to})"
```
l-kent, line 27: Why make this change?

src/main/scala/analysis/LAST_VSA_BACKUP.scala
l-kent, line 1: Please don't commit unnecessary files like this

src/main/scala/analysis/MemoryModelMap.scala
```diff
@@ -4,67 +4,156 @@ import analysis.*
...
+      case None =>
+        // If no memory is stored at the requested address, return zero
+        BigInt(0) // TODO: may need to be sm else
+    }
+  }
 
   /** Add a range and object to the mapping
    *
-   * @param offset the offset of the range
+   * @param offset the offset of the range, if a heap region is given, the offsets controls the shift of regions from the start
```
l-kent, line 93: What does this mean?

src/main/scala/analysis/SteensgaardAnalysis.scala
l-kent, line 1: Unneeded files should be deleted

src/main/scala/analysis/SSAForm.scala
l-kent, line 1: Unneeded files should be deleted

build.sbt
```diff
@@ -10,6 +10,7 @@ val scalactic = "org.scalactic" %% "scalactic" % "3.2.10"
 val antlrRuntime = "org.antlr" % "antlr4-runtime" % "4.9.3"
 val sourceCode = "com.lihaoyi" %% "sourcecode" % "0.3.0"
 val mainArgs = "com.lihaoyi" %% "mainargs" % "0.5.1"
+val parralelCollections = "org.scala-lang.modules" %% "scala-parallel-collections" % "1.0.4"
```
l-kent, line 13: What is this used for?

src/main/scala/analysis/GlobalRegionAnalysis.scala
yousifpatti: This analysis only evaluates data regions and MRA evaluates the remaining stack and heap regions. MRA uses the results of GRA to skip statements that are already evaluated to data regions

src/main/scala/analysis/Lattice.scala
```diff
@@ -25,6 +26,794 @@ trait Lattice[T]:
...
+//      val l = bitVec_min(x)
+//      val u = bitVec_max(x)
+//      val initialStride = smt_bvsub(u, l)
+//      val stride = x.foldLeft(initialStride) {
+//          case (acc, v) => bitVec_gcd(smt_bvsub(v, l), acc)
+//      }
+//      SI(stride, l, u)
+//    }
+//  }
+//}
```
l-kent, line 164: Please don't include large commented out sections like this

src/main/scala/analysis/GlobalRegionAnalysis.scala
l-kent: Why do we need separate analyses for that then?

build.sbt
yousifpatti: Intellij might have added it automatically. I will remove

src/main/scala/analysis/GlobalRegionAnalysis.scala
yousifpatti: It was to separate the eval and transfer functions. They can be merged by bringing the functions over. I will work on that

src/main/scala/analysis/IrreducibleLoops.scala
yousifpatti: That was for testing. I will remove

src/main/scala/util/RunUtils.scala
```diff
@@ -539,13 +559,16 @@ object RunUtils {
     val analysisResult = mutable.ArrayBuffer[StaticAnalysisContext]()
     while (modified) {
       Logger.debug("[!] Running Static Analysis")
-      val result = StaticAnalysis.analyse(ctx, config, iteration)
+      val result = StaticAnalysis.analyse(ctx, config, iteration, analysisResult.lastOption)
```
l-kent, line 562: It's not safe to feed the previous iteration's results in like this - since the IR has been modified, the previous results will not be valid for many cases.

yousifpatti: At this point only IndirectCalls would be replaced by DirectCalls and the domain recalculated so any other node would have the result of the analysis. Why would it not be valid?

src/main/scala/analysis/SteensgaardAnalysis.scala
yousifpatti: This file is deleted not sure why it is showing here

src/main/scala/analysis/SSAForm.scala
yousifpatti: This file is deleted not sure why it is showing here

src/main/scala/analysis/MemoryModelMap.scala
yousifpatti: It is a way to give heap regions an assumed "offset". This should mean the offset passed is alway 0 but not explained properly. I will modify the function and the comments to be clearer

src/main/scala/util/RunUtils.scala
l-kent: Anything after a resolved call would not be valid

src/main/scala/analysis/MemoryModelMap.scala
l-kent: Why do you give heap regions an assumed offset at all if you always use 0?

l-kent:
There is a regression in resolving indirect calls in that the indirect_call_outparam tests no longer pass.

src/main/scala/analysis/MemoryRegionAnalysis.scala
l-kent: This is still an issue

yousifpatti: Fixed in the coming push

src/main/scala/util/RunUtils.scala
yousifpatti: The results fed are intraprocedural so they should not change. They will be undefined for new functions that have not been analysed but this case is handled

l-kent:
This seems to be working much better than the previous version - region results seem to generally be correct with the stack and data regions, though I haven't had a close look at anything complicated with multiple procedures yet.

The big remaining issue is heap identification, which doesn't seem to be working at all right now?

src/main/scala/analysis/MemoryRegionAnalysis.scala
```diff
@@ -1,29 +1,32 @@
...
-  val stackMap: mutable.Map[Procedure, mutable.Map[Expr, StackRegion]] = mutable.Map()
+  val stackMap: mutable.Map[Procedure, mutable.Map[BigInt, StackRegion]] = mutable.Map()
 
-  private def nextMallocCount() = {
-    mallocCount += 1
-    s"malloc_$mallocCount"
+  private def nextMallocCount(size: BigInt) = {
+    val start = mallocCount
+    mallocCount += (size.toDouble/8).ceil.toInt + 1
+    (s"malloc_$mallocCount", start)
```
l-kent, line 29: I don't understand what this is trying to do. Why does the size have anything to do with giving each malloc a unique identifier?

l-kent:
There is a problem with the regions analysis for pointers-to-pointers. This is an example of this from correct/secret_write/clang_pic:BAP:
```
    assume {:captureState "%0000031d"} true;
    R8, Gamma_R8 := 65536bv64, true;
    call rely();
    R8, Gamma_R8 := memory_load64_le(data_11, bvadd64(R8, 4048bv64)), (gamma_load64(Gamma_data_11, bvadd64(R8, 4048bv64)) || L(data_11, bvadd64(R8, 4048bv64)));
    call rely();
    R10, Gamma_R10 := zero_extend32_32(memory_load32_le(data_11, R8)), (gamma_load32(Gamma_data_11, R8) || L(data_11, R8));
``` 
Here, the contents of address 0x10FB8 (69560 in decimal), which is the value 0x11034 (69684 in decimal), is loaded into R8. The analysis gives this access the region data_11. 0x11034 is the address of global variable z. The contents of global variable z is then loaded into R10 in the subsequent load, which is also given the region data_11. 

There is no reason that these two accesses should have the same region - they are completely separate. This same pattern occurs in many other places.

src/main/scala/analysis/MemoryModelMap.scala
```diff
@@ -254,20 +522,125 @@ class MemoryModelMap {
...
   val regionIdentifier: String
+  val start: BigInt
+  val subAccesses: mutable.Set[BigInt] = mutable.Set()
 }
 
-case class StackRegion(override val regionIdentifier: String, start: BitVecLiteral, parent: Procedure) extends MemoryRegion {
-  override def toString: String = s"Stack($regionIdentifier, $start, ${parent.name})"
+case class StackRegion(override val regionIdentifier: String, start: BigInt, parent: Procedure) extends MemoryRegion {
+  override def toString: String = s"Stack($regionIdentifier, $start, ${parent.name}, $subAccesses)"
 }
```
l-kent, line 579: Why hasn't a size been added to this?

l-kent: `start` also doesn't override the trait's `start` variable because you haven't made it `override val`. This applies to the other region case classes too.

src/main/scala/analysis/MemoryModelMap.scala
```diff
@@ -254,20 +522,125 @@ class MemoryModelMap {
...
 
-case class HeapRegion(override val regionIdentifier: String, size: BitVecLiteral, parent: Procedure) extends MemoryRegion {
+case class HeapRegion(override val regionIdentifier: String, start: BigInt, size: BigInt, parent: Procedure) extends MemoryRegion {
   override def toString: String = s"Heap($regionIdentifier, $size)"
 }
```
l-kent, line 583: What does it mean for a HeapRegion to have a `start`? They don't have any base address, only an abstract one that is returned from malloc.

src/main/scala/analysis/MemoryModelMap.scala
```diff
@@ -254,20 +522,125 @@ class MemoryModelMap {
...
+
+  def getMergedName(regions: Set[MemoryRegion]): String = {
+    mergedRegions(regions)
+  }
 }
 
 trait MemoryRegion {
   val regionIdentifier: String
+  val start: BigInt
+  val subAccesses: mutable.Set[BigInt] = mutable.Set()
```
l-kent, line 574: It's bad practice to put mutable parameters like this in a case class (or in this case, a trait that case classes extend).

src/main/scala/analysis/MemoryModelMap.scala
```diff
@@ -254,20 +522,125 @@ class MemoryModelMap {
...
+case class HeapRegion(override val regionIdentifier: String, start: BigInt, size: BigInt, parent: Procedure) extends MemoryRegion {
   override def toString: String = s"Heap($regionIdentifier, $size)"
 }
 
-case class DataRegion(override val regionIdentifier: String, start: BitVecLiteral) extends MemoryRegion {
-  override def toString: String = s"Data($regionIdentifier, $start)"
+case class DataRegion(override val regionIdentifier: String, start: BigInt, size: BigInt) extends MemoryRegion {
+  override def toString: String = s"Data($regionIdentifier, $start, $size, ($relfContent))"
+  def end: BigInt = start + size - 1
+  val relfContent: mutable.Set[String] = mutable.Set[String]()
```
l-kent, line 588: It's bad practice to put mutable parameters like this in a case class (or in this case, a trait that case classes extend).

l-kent:
I need a way of resolving an address (ideally with access size too) to a region or set of regions and was previously using findDataObject in the MemoryModelMap for this, but it does not seem to provide accurate information about regions now. 

This seems to be because the MMM is not accurately updated to match the GRA results, or it could be resulting from a bug in the GRA. 

An example of this is for the test correct/basicassign_gamma0/clang_pic:BAP:

```
    R8, Gamma_R8 := 65536bv64, true;
    call rely();
    R8, Gamma_R8 := memory_load64_le(data_8, bvadd64(R8, 4048bv64)), (gamma_load64(Gamma_data_8, bvadd64(R8, 4048bv64)) || L(bvadd64(R8, 4048bv64)));
    call rely();
    R8, Gamma_R8 := zero_extend32_32(memory_load32_le(data_8, R8)), (gamma_load32(Gamma_data_8, R8) || L(R8));
```
The analysis gives that the accesses 69568 (0x10FD0) and 69684 (0x11034, as it is the value stored at 0x10FD0) have region data_8. This is the same issue described above with correct/secret_write/clang_pic:BAP where a pointer and a pointer to pointer are incorrectly said to have the same region. The further issue here is that the address 69684 is not added to the MMM's dataMap at all. This seems to be because the data_8 DataRegion object produced by the GRA only refers to address 69568, not 69684, so the MMM doesn't know anything about 69684. You need to make sure that the GRA results and MMM results are completely aligned.

l-kent:
For the test correct/initialisation/clang:BAP, the analysis gives regions with incorrect size for accesses %00000367 and %00000359.
```
000002fd: R8 := 0x11000
00000302: R9 := 0x11000
00000308: R9 := R9 + 0x40
0000030d: R11 := 0x11000
00000314: R10 := pad:64[mem[R11 + 0x30, el]:u32]
0000031a: R10 := pad:64[31:0[R10] + 1]
00000322: mem := mem with [R11 + 0x30, el]:u32 <- 31:0[R10]
00000327: R11 := 0x11000
0000032e: R10 := mem[R11 + 0x38, el]:u64
00000334: #4 := R10 - 2
00000339: VF := extend:65[#4 + 1] <> extend:65[R10] - 1
0000033e: CF := pad:65[#4 + 1] <> pad:65[R10] + 0xFFFFFFFFFFFFFFFF
00000342: ZF := #4 + 1 = 0
00000346: NF := 63:63[#4 + 1]
0000034a: R10 := #4 + 1
00000352: mem := mem with [R11 + 0x38, el]:u64 <- R10
00000359: R8 := pad:64[mem[R8 + 0x40, el]:u32]
0000035f: R8 := pad:64[31:0[R8] + 3]
00000367: mem := mem with [R9 + 4, el]:u32 <- 31:0[R8]
0000036c: R9 := 0x11000
00000373: R8 := pad:64[mem[R9 + 0x34]]
00000379: R8 := pad:64[31:0[R8] + 1]
00000381: mem := mem with [R9 + 0x34] <- 7:0[R8]
00000386: R0 := 0
0000038b: call R30 with noreturn
```
Here, the access at %00000367 is to the address 0x11044 (69700 in decimal). This is a 32-bit access or 4-byte access. The GRA gives this access the region data_11, which starts at 69700 and has a size of 8 bytes. This is an incorrect size - it should only be 4 bytes as this is the only time this address is accessed.

The 32-bit (4-byte) access at %00000359 is to the address 0x11040 (69696 in decimal). The GRA gives this access the region data_14, which starts at 69696 and has a size of 8 bytes. This is also an incorrect size - this is also the only time this address is accessed, so it should only have a size of 4 bytes. In addition, if both those sizes were correct, the regions should have been merged due to being overlapping.

src/main/scala/analysis/Lattice.scala
l-kent, line 1: Most or all of the additions to this file are unused - please don't commit unneeded & unfinished work to a branch we are in the process of merging. Keep that to your own separate branches.

src/main/scala/analysis/GlobalRegionAnalysis.scala
```diff
@@ -0,0 +1,246 @@
...
+
+import scala.collection.mutable
+
+trait GlobalRegionAnalysis(val program: Program,
+                           val domain: Set[CFGPosition],
+                           val constantProp: Map[CFGPosition, Map[Variable, FlatElement[BitVecLiteral]]],
+                           val reachingDefs: Map[CFGPosition, (Map[Variable, Set[Assign]], Map[Variable, Set[Assign]])],
+                           val mmm: MemoryModelMap,
+                           val globalOffsets: Map[BigInt, BigInt],
+                           val vsaResult: Option[Map[CFGPosition, LiftedElement[Map[Variable | MemoryRegion, Set[Value]]]]]) {
```
l-kent, line 14: it doesn't really make sense to have a Map inside an Option - you can just use an empty Map as the default instead of None.

src/main/scala/analysis/GlobalRegionAnalysis.scala
```diff
@@ -0,0 +1,246 @@
...
+                            case LiftedBottom =>
+                            case _ =>
+                          case None =>
+                      case literalValue: LiteralValue =>
+                    }
+                    case None =>
+                  case LiftedBottom =>
+                  case _ =>
+                case None =>
+            }
```
l-kent, line 137: This seems overly convoluted

src/main/scala/analysis/GlobalRegionAnalysis.scala
```diff
@@ -0,0 +1,246 @@
...
+          return Set(region.get)
+      }
+    }
+    exp match
+      case literal: BitVecLiteral => tryCoerceIntoData(literal, n, subAccess)
+      case Extract(end, start, body) => tryCoerceIntoData(body, n, subAccess)
+      case Repeat(repeats, body) => tryCoerceIntoData(body, n, subAccess)
+      case ZeroExtend(extension, body) => tryCoerceIntoData(body, n, subAccess)
+      case SignExtend(extension, body) => tryCoerceIntoData(body, n, subAccess)
+      case UnaryExpr(op, arg) => tryCoerceIntoData(arg, n, subAccess)
```
l-kent, line 81: None of these are robust ways of trying to identify which data regions an expression points to. You can't just ignore these operations like this, it will give incorrect results.

src/main/scala/analysis/GlobalRegionAnalysis.scala
```diff
@@ -0,0 +1,246 @@
...
+    while (globalOffsets.contains(tableAddress) && globalOffsets.contains(globalOffsets(tableAddress)) && !exitLoop) {
+      val newAddress = globalOffsets.getOrElse(tableAddress, tableAddress)
+      if (newAddress == tableAddress) {
+        exitLoop = true
+      } else {
+        tableAddress = newAddress
+      }
+    }
+    tableAddress
+  }
```
l-kent, line 65: I think this is the cause of the issue with pointers-to-pointers being given the same region as the pointers they point to. This is not correct - you can't just replace a pointer-to-a-pointer with said pointer like this.

src/main/scala/analysis/BitVectorEval.scala
```diff
@@ -328,4 +329,29 @@ object BitVectorEval {
...
+
+  @tailrec
+  def bitVec_gcd(a: BitVecLiteral, b: BitVecLiteral): BitVecLiteral = {
+    if (b.value == 0) a else bitVec_gcd(b, smt_bvsmod(a, b))
+  }
+
+  def bitVec_interval(lb: BitVecLiteral, ub: BitVecLiteral, step: BitVecLiteral): Set[BitVecLiteral] = {
+    require(smt_bvule(lb, ub) == TrueLiteral, "Lower bound must be less than or equal to upper bound")
+    (lb.value to ub.value by step.value).map(BitVecLiteral(_, lb.size)).toSet
+  }
```
l-kent, line 356: Is any of this needed for anything actually in use? If not, it shouldn't be included in this pull request

src/main/scala/analysis/ReachingDefinitionsAnalysis.scala
```diff
@@ -3,7 +3,7 @@ package analysis
 import ir.*
 import analysis.solvers.SimpleWorklistFixpointSolver
 
-case class ReachingDefinitionsAnalysis(program: Program) {
+case class ReachingDefinitionsAnalysis(program: Program, inter: Boolean = false) {
```
l-kent, line 6: This inter flag doesn't actually do anything?

src/main/scala/analysis/InterprocSteensgaardAnalysis.scala
```diff
@@ -189,15 +183,16 @@ class InterprocSteensgaardAnalysis(
...
-            val regions = mmm.findSharedStackObject(b.value)
+            val regions = mmm.findSharedStackObject(b.get.value)
             Logger.debug("found: " + regions)
             res ++= regions
           } else {
-            val region = mmm.findStackObject(b.value)
+            val region = mmm.findStackObject(b.get.value)
             if (region.isDefined) {
               res = res + region.get
             }
```
l-kent, line 198: This is assuming that the operation is addition

src/main/scala/analysis/InterprocSteensgaardAnalysis.scala
```diff
@@ -245,98 +242,130 @@ class InterprocSteensgaardAnalysis(
...
+        return Set.empty[MemoryRegion] // TODO: IT SHOULD THROW AN EXCEPTION
+        //throw Exception(s"Could not find region for MemLoad: $memLoad, Command: $cmd, Eval: $isGlobal, Global: $globalRegion")
+      }
+      memLoad.index match // treats case where the index is a region and is loaded again like in jumptable2/clang_pic
+        case variable: Variable =>
+          val ctx = getUse(variable, cmd, reachingDefs)
+          for (i <- ctx) {
+            i.rhs match {
+              case load: MemoryLoad =>
+                return memLoadToRegion(load, i)
```
l-kent, line 264: Having a return inside a for loop like this is deprecated in Scala and gives a warning. Since getUse returns a set, why do we only return a single result here?

l-kent:
The branch `memory-region-boogie` now has adding regions to Boogie working properly, except where there are issues with the underlying analyses that I've mentioned. I will wait for those to be addressed before merging at this stage.

l-kent:
Can you push your latest set of fixes to this branch so I can work on merging into main?

yousifpatti:
I pushed up to the working things

l-kent:
For correct/basic_lock_read/gcc:BAP, there is an inconsistency between the GRA result and the MMM. The GRA correctly gives a region with start address 69652 (0x11014) and size 4 to accesses to that address, but this region does not appear in the MMM's dataMap. I haven't yet identified why.

This was an issue before your most recent changes but I had previously missed it.

l-kent:
For correct/initialisation/clang_pic:BAP, there are some region size issues in relation to array accesses as previously discussed, but there is another error in the analysis as well.

This is the relevant parts of the .bir file:
```
00000305: R9 := 0x10000
0000030c: R9 := mem[R9 + 0xFC8, el]:u64

0000036b: R8 := pad:64[mem[R9, el]:u32]

00000379: mem := mem with [R9 + 4, el]:u32 <- 31:0[R8]
```

The access at 0000036b is incorrectly given a region with a size of 8 instead of 4 (it only has a 32-bit access) - this is the same region size issue with array accesses.

The access at 00000379 is incorrectly given a region with size 8 and start address 0x10FCC (69580). It seems to have gotten this by doing 0x10000 + 0xFC8 + 4, but the fact that 0x10FC8 (69576) is loaded from memory and the result has 4 added to it has been missed. The correct address is 0x11044 (69700) as mem[0x10FC8] == 0x11040 (69696)

l-kent:
For correct/secret_write/clang:BAP, there is a new error introduced by the most recent changes. The GRA correctly gives accesses to address 0x11034 (69684, which is the address of x) a region, but this region is not included in the MMM's dataMap.

l-kent:
For correct/functionpointer/clang:GTIRB, there is an issue with the stack regions. This is an excerpt of the output:

```
    R29, Gamma_R29 := bvadd64(R31, 32bv64), Gamma_R31;
    stack_9, Gamma_stack_9 := memory_store32_le(stack_9, bvadd64(R29, 18446744073709551612bv64), 0bv32), gamma_store32(Gamma_stack_9, bvadd64(R29, 18446744073709551612bv64), true);
    assume {:captureState "1872$0"} true;
    stack_20, Gamma_stack_20 := memory_store32_le(stack_20, bvadd64(R29, 18446744073709551608bv64), R0[32:0]), gamma_store32(Gamma_stack_20, bvadd64(R29, 18446744073709551608bv64), Gamma_R0);
    assume {:captureState "1876$0"} true;
    R8, Gamma_R8 := zero_extend32_32(memory_load32_le(stack_8, bvadd64(R29, 18446744073709551608bv64))), gamma_load32(Gamma_stack_8, bvadd64(R29, 18446744073709551608bv64));
```

The accesses to regions stack_20 and stack_8 both have the same address (R29 - 8 == R31 + 24) but have different regions. This is incorrect.

This was an issue before your most recent changes but I had missed it.

src/main/scala/util/RunUtils.scala
```diff
@@ -537,15 +558,18 @@ object RunUtils {
     var iteration = 1
     var modified: Boolean = true
     val analysisResult = mutable.ArrayBuffer[StaticAnalysisContext]()
-    while (modified) {
+    while (modified || analysisResult.size < 2) {
```
l-kent, line 561: This isn't a good way to make use of the VSA results, as we don't need to run every analysis twice. The way the analyses feed into each other could be significantly improved - having a circular dependency between them really doesn't seem ideal.

src/main/scala/analysis/MemoryModelMap.scala
```diff
@@ -4,212 +4,460 @@ import analysis.*
...
+          } else {
+            result
+          }
+          BigInt(1, paddedResult) // Convert the byte array back to BigInt
+        }
+      case None =>
+        // If no memory is stored at the requested address, return zero
+        BigInt(0) // TODO: may need to be sm else
+    }
+  }
```
l-kent, line 90: Is any of this actually used for anything?

src/main/scala/analysis/MemoryModelMap.scala
```diff
@@ -254,20 +521,126 @@ class MemoryModelMap {
...
   override def toString: String = s"Heap($regionIdentifier, $size)"
 }
 
-case class DataRegion(override val regionIdentifier: String, start: BitVecLiteral) extends MemoryRegion {
-  override def toString: String = s"Data($regionIdentifier, $start)"
+case class DataRegion(override val regionIdentifier: String, override val start: BigInt, size: BigInt) extends MemoryRegion {
+  override def toString: String = s"Data($regionIdentifier, $start, $size, ($relfContent))"
+  def end: BigInt = start + size - 1
+  val relfContent: mutable.Set[String] = mutable.Set[String]()
+  val isPointerTo: Option[DataRegion] = None
```
l-kent, line 588: What is this for?

src/main/scala/analysis/MemoryModelMap.scala
```diff
@@ -4,212 +4,460 @@ import analysis.*
...
+    }
+  }
+
+  /* All regions that either:
+   * 1. starts at value but size less than region size
+   * 2. starts at value but size more than region size (add both regions ie. next region)
+   * 3. starts between regions (start, end) and (value + size) => end
+   * 4. starts between regions (start, end) and (value + size) < end (add both regions ie. next region)
+   */
+  def findStackPartialAccessesOnly(value: BigInt, size: BigInt): Set[StackRegion] = {
```
l-kent, line 312: Is this used for anything?

src/main/scala/analysis/MemoryModelMap.scala
```diff
@@ -4,212 +4,460 @@ import analysis.*
...
+            nextStart = nextRange.end
+          }
+        }
+      }
+    }
+
+    matchingRegions.toSet.map(returnRegion)
+  }
+
+  def getRegionsWithSize(size: BigInt, function: String, negateCondition: Boolean = false): Set[MemoryRegion] = {
```
l-kent, line 355: Is this used for anything?

src/main/scala/analysis/MemoryRegionAnalysis.scala
```diff
@@ -195,58 +215,109 @@ trait MemoryRegionAnalysis(val program: Program,
...
+                val newHeapRegion = HeapRegion(name, start, negB, IRWalk.procedure(n))
+                addReturnHeap(directCall, newHeapRegion)
+                s
+              case None =>
+                // Assume heap region size is at least 1 TODO: must approximate size of heap
+                val negB = 1
+                val (name, start) = nextMallocCount(negB)
+                val newHeapRegion = HeapRegion(name, start, negB, IRWalk.procedure(n))
+                addReturnHeap(directCall, newHeapRegion)
+                s
```
l-kent, line 230: You don't need to approximate the size of a heap allocation if you don't know it. The size of each heap allocation is not important except for detecting out-of-bounds accesses (which is not the most important). Each heap allocation needs to be its own separate address space, with the R0 returned from malloc being like the stack pointer for that heap.

src/main/scala/util/RunUtils.scala
```diff
@@ -386,25 +395,38 @@ object StaticAnalysis {
...
+      writeToFile(
+        toDot(IRProgram, IRProgram.filter(_.isInstanceOf[Command]).map(b => b -> graResult(b).toString).toMap),
+        s"${s}_GRA$iteration.dot"
+      )
     })
 
     Logger.debug("[!] Running MMM")
-    val mmm = MemoryModelMap()
-    mmm.convertMemoryRegions(mraResult, mergedSubroutines, globalOffsets, mraSolver.procedureToSharedRegions)
+    mmm.convertMemoryRegions(mraSolver.procedureToStackRegions, mraSolver.procedureToHeapRegions, mraSolver.mergeRegions, mraResult, mraSolver.procedureToSharedRegions, graSolver.getDataMap, graResult)
```
l-kent, line 406: mraSolver.mergeRegions is always empty, isn't it?

src/main/scala/analysis/VSA.scala
```diff
@@ -45,123 +35,81 @@ trait ValueSetAnalysis(program: Program,
...
-    } else if (globalOffsets.contains(bitVecLiteral.value)) {
-      GlobalAddress(bitVecLiteral, resolveGlobalOffset(bitVecLiteral.value))
-    } else if (subroutines.contains(bitVecLiteral.value)) {
-      GlobalAddress(bitVecLiteral, subroutines(bitVecLiteral.value))
-    } else {
-      LiteralValue(bitVecLiteral)
-    }
+  def canCoerceIntoDataRegion(bitVecLiteral: BitVecLiteral, size: Int): Option[DataRegion] = {
+    mmm.findDataObject(bitVecLiteral.value)
   }
```
l-kent, line 55: Why does this have an unused size parameter?

src/main/scala/analysis/GlobalRegionAnalysis.scala
```diff
@@ -0,0 +1,247 @@
...
+
+  def tryCoerceIntoData(exp: Expr, n: Command, subAccess: BigInt, loadOp: Boolean = false): Set[DataRegion] = {
```
l-kent, line 74: I don't really understand how this method works, can you explain it? I don't really follow why BinaryExprs are such a special case for it?

src/main/scala/analysis/InterprocSteensgaardAnalysis.scala
```diff
@@ -60,11 +60,6 @@ class InterprocSteensgaardAnalysis(
 
   def getMemoryRegionContents: Map[MemoryRegion, Set[BitVecLiteral | MemoryRegion]] = memoryRegionContents.map((k, v) => k -> v.toSet).toMap
```
l-kent, line 61: This is always just empty now isn't it?

src/main/scala/analysis/GlobalRegionAnalysis.scala
```diff
@@ -0,0 +1,247 @@
...
+        dataMap(offset) = DataRegion(dataMap(offset).regionIdentifier, offset, (size.toDouble / 8).ceil.toInt)
+        Some(dataMap(offset))
+      } else {
+        Some(dataMap(offset))
+      }
+    } else {
+      dataMap(offset) = DataRegion(nextDataCount(), offset, (size.toDouble / 8).ceil.toInt)
+      Some(dataMap(offset))
+    }
+  }
```
l-kent, line 47: Why does this need to return an Option when it never returns None?

l-kent:
I'm going to merge this into main and file issues for remaining bugs now.

l-kent:
The --memory-regions flag can be used to enable the regions output

src/main/scala/util/RunUtils.scala
l-kent: The VSA depends on the ConstantPropagation results though, and the ConstantPropagation is interprocedural.

PR #233, opened 2024-08-21
Add docs page on scala programming advice

ailrst:


docs/development/scala.md
```diff
@@ -0,0 +1,163 @@
...
+Below is an unorganised list of common pitfalls encountered when programming Scala.
+
+- Statement lists are newline-separated sequences of expressions, which simply evaluate to the last expression in the list. It is easy to accidentally silently 
+  throw away a computed value.
+  - Throwing away values in the middle of the list isn't a warning by default, but can be made one with `-Wnonunit-statement`. 
+    Unfortunately there is no standard library `ignore: _ => Unit` function, and the design of `scala.collections` in always 
+    returning non-unit, even purely side-effecting operations makes enabling this warning require somewhat more convoluted code.
+- Scala contains syntax to encourage procedural programming; if/else statements, for, while statements; 
+  the fact this is implemented through complex syntax sugar on top of functional constructs makes them surprising to both procedural and functional programmers, 
+  and best avoided.
```
l-kent, line 33: I am a bit confused by this - is this suggesting that if/else statements and for and while loops should all be avoided? I don't think that would be useful advice at all.

docs/development/scala.md
```diff
@@ -0,0 +1,163 @@
...
+- There are at least two ways to define an ADT (Algebraic Data Type), using an `enum` or `trait` and `case class`es. By default `enum`s make you to prefix every case with 
+  the enum name, to avoid this you can use `export` to add them to enclosing scope. Scala's case classes are overloaded, they are both the standard record type if defined without parent trait, and the constructor for ADT variants. 
+  - Note that `trait`/`case class`-based ADTs are open by default hence don't enable case-exhaustion checking on pattern matches (instead default to `sealed trait`)
+  ```scala
+  // enum style
+
+  enum Error(r: String):
+    case Permission extends Error("permission")
+    case NotExists extends Error("not existing")
+  export Error._ // adds Permission and Exists to the enclosing scope
```
l-kent, line 50: this is Scala 2 syntax, Scala 3 is `export Error.*`

docs/development/scala.md
```diff
@@ -0,0 +1,163 @@
...
+  class C(v: Int) extends X(v)     // also compiles due to `v` parameter shadowing
+  class E(val v: Int) extends X(v) // compile error due to missing override modifier 
+  ```
+
+- `case class`es are accessible like tuples, the below is valid code (never write this)
+  ```scala
+  case class X(a: Int, b: Int)
+  val x = X(1,2)
+  x._1 == x.a
+  ```
```
l-kent, line 119: I think this should have more emphasis as something NOT to do

docs/development/scala.md
ailrst: The point generally is that using procedural constructs gets you the X/Y problem that makes one reach for nonsense like `breakable`, which isn't required if you're writing functional style code.  

I don't think there's ever any reason to write a while loop in Scala they always become convoluted procedural fragments of code. Pattern matching is usually better than if/else but it does have its place. And generally I think map should be used instead of for, except for purely side-effecting operations which should be rare if you have immutable data, but they obviously mean the same thing in scala. Except that for is overloaded for do notation.

l-kent: The only thing that actually causes problems is `breakable`, which indeed should be avoided. That part was fine, which is why I didn't highlight it. 

For loops being syntactic sugar for map doesn't actually cause problems and using for can sometimes be more readable than using map. There are plenty of cases where if statements are appropriate and pattern matching is not, so more useful advice may be 'pattern matching is recommended where possible' or something along those lines? There are of course plenty of places where we have mutable data as well - for better or worse the project is not written in anything remotely close to a purely functional style and moving away from that would be far more trouble than it's worth. While loops generally only make sense when involving mutable collections but we have those.

I'm mostly concerned that advising people who work on this project (who generally have little-to-no background in functional programming) to avoid these standard procedural programming constructs that they are familiar with will just cause them to write worse, more confused code that avoids them unnecessarily.

PR #237, opened 2024-08-22
docs: add usage.md for users of BASIL

katrinafyi:
This moves the outdated "Lifting a single example" readme section into a separate docs file, and updates and elaborates it. This is meant to be a self-contained page for end-users and so it duplicates some of the installation steps.

I would appreciate feedback on the content, especially the Boogie portion. I also wonder if some of this content should remain elsewhere, as its useful to BASIL developers as well.


docs/usage.md
```diff
@@ -0,0 +1,190 @@
...
+   aarch64-suse-linux-readelf -s -r -W a.out > a.relf
+   ```
+
+### Using BASIL to generate Boogie
+
+This generates a Boogie file in out.bpl.
+```
+basil --input a.gts --relf a.relf -o out.bpl
+```
+(If using BASIL in-repo, substitute `./mill run` in place of `basil`.
```
ailrst, line 137: "If using basil compiled from source rather than use the nix package" or similar.

docs/usage.md
```diff
@@ -0,0 +1,190 @@
...
+```
+
+The `/useArrayAxioms` flag is necessary for Boogie versions 2.16.8 and greater;
+for earlier versions it can be removed.
+This improves the verification speed, by using Boogie's axiomatic encoding of arrays rather than Z3's. 
+This encoding does not support extensionality, array extensionality can be disabled on newer versions of Boogie with:
+
+```
+boogie example.bpl /proverOpt:O:smt.array.extensional=false
+```
```
ailrst, line 163: But he axiomatic encoding is faster still at the expense of completeness

ailrst (commented):
Looks good thanks for doing this, I can fix the comments below myself tomorrow

docs/usage.md
katrinafyi: I said in repo because it specifically needs the repo checked out to ./mill, but happy with any rewording.

docs/usage.md
```diff
@@ -0,0 +1,190 @@
...
+  github:katrinafyi/pac-nix#basil
+```
+
+ddisasm may also be installed through its provided Docker images or
+APT repository (as of Aug 2024, only supports Ubuntu 20.04).
+You can also, of course, build any of these tools manually.
+
+See also: [tool-installation](development/tool-installation.md)
+and [development: building](development/readme.md#building)
+(primarily if you are interested in buliding BASIL manually).
```
katrinafyi, line 55:
```suggestion
(primarily if you are interested in building BASIL manually).
```

docs/usage.md
ailrst: I don't think its likely anyone will obtain the source other than by checking out the repo

PR #239, opened 2024-09-03
Loads all symbol table row data

ailrst:


src/main/scala/translating/ReadELFLoader.scala
```diff
@@ -6,16 +6,56 @@ import util.ILLoadingConfig
...
-  def visitSyms(ctx: SymsContext, config: ILLoadingConfig): (Set[ExternalFunction], Set[SpecGlobal], Map[BigInt, BigInt], Int) = {
+  def visitSyms(ctx: SymsContext, config: ILLoadingConfig): (List[ELFSymbol], Set[ExternalFunction], Set[SpecGlobal], Map[BigInt, BigInt], Int) = {
     val externalFunctions = ctx.relocationTable.asScala.flatMap(r => visitRelocationTableExtFunc(r)).toSet
     val relocationOffsets = ctx.relocationTable.asScala.flatMap(r => visitRelocationTableOffsets(r)).toMap
-    val globalVariables = ctx.symbolTable.asScala.flatMap(s => visitSymbolTable(s)).toSet
     val mainAddress = ctx.symbolTable.asScala.flatMap(s => getFunctionAddress(s, config.mainProcedureName))
+
+    val symbolTable = ctx.symbolTable.asScala.flatMap(s => visitSymbolTable(s)).toList
+    val globalVariables = (symbolTable.collect {
+      case ELFSymbol(num, value, size, ELFSymType.OBJECT, ELFBind.GLOBAL, ELFVis.DEFAULT, _, name) =>  SpecGlobal(name, size * 8, None, value)
```
l-kent, line 52: Removing the `allowedChars` check means that variables with illegal characters in Boogie are now added to the .bpl output in some cases. In practice this `allowedChars` check just meant that references to external objects, such as the `stack_chk_guard`, were not included in the set of `SpecGlobal`s. A more robust solution would probably be to exclude symbol table entries with ndx `UND` rather than relying on them containing an illegal character such as `@`.

ailrst: Ah yeah removing that was a mistake, ndx UND sounds like a good fix.

PR #240, opened 2024-09-04
Points to analysis union find

sadrabt:
- [ ] extend DSAUnionFindSolver from UnionFindSolver
- [ ] Rename Local tests
- [ ] Document tests 
- [ ] replace tuples with case classes
- [ ] single instance of stack symbolic access for each variable violated by indirect_call and malloc_with_local3
- [ ]  bu(proc) in DSA throws no such elemenet exception for tests correct/indirect_call/clang_O2:GTIRB and correct/malloc_with_local3/clang_O2:GTIRB
- [ ] storeExitJump(exitToEntry(exit), d1, d2) throws cast error for correct/syscall/gcc_O2:GTIRB and correct/syscall/clang_O2:GTIRB
- [ ] resolve merge conflicts

This is DSA as implemented by `points-to-analysis-alternate-merge` in #222 which instead uses a union-find data structure.  There is also added functionality for converting DSG's to a dot graph


l-kent:
Rather than closing the old pull request, it would make more sense to just change the branch the pull request for, that way the existing comments would remain

sadrabt:
> Rather than closing the old pull request, it would make more sense to just change the branch that pull request #222 was for, that way the existing comments would remain.

The pull request only allows for changing the base branch (main), should I force push  this branch to #222's branch instead?

l-kent:
That would be fine

PR #241, opened 2024-09-05
Rewrite IR Interpreter

ailrst:
Rewrite of the interpreter in a FP state-monad style, supporting new features

- breakpoints
- traces
- cycle count limit
- intrinsic functions
- partial memory region support
- indirect call resolution
- nondeterministic gotos handled via lookahead

This builds on the call-statement pr and includes the readelf loading changes. 

- It adds the tests `DifferentialAnalysis` which runs all the `src/test/correct` example programs through the static analysis, and compares the interpreter's final state, and execution trace to the un-transformed program. Many of these still fall over due to missing intrinsics (e.g. malloc), and memory initialisation.
- Adds the test `ConstPropInterpreterValidate` which converts the const-prop results to boolean expressions, and runs the interpreter, asserting that the results hold at every code point.

For more details see https://github.com/UQ-PAC/BASIL/blob/interpreter/docs/development/interpreter.md

Incidentally on my computer the interpreter runs at approximately 345 kHz (basil instructions/second)


src/test/scala/ir/InterpreterTests.scala
```diff
@@ -1,46 +1,69 @@
...
+}
+
+class InterpreterTests extends AnyFunSuite with BeforeAndAfter {
 
+  Logger.setLevel(LogLevel.WARN)
 
+  def getProgram(name: String): IRContext = {
     val loading = ILLoadingConfig(
       inputFile = s"examples/$name/$name.adt",
       relfFile = s"examples/$name/$name.relf",
```
l-kent, line 42: Tests should be updated to use inputs from the tests folder instead of the examples folder (which has mostly been removed)

l-kent:
I haven't had a deep look at this yet but:

- The test "var load store" in InterpreterTests doesn't currently pass.
- There are warnings about incorrect indentation and pattern match exhaustivity and deprecated casting which should be addressed.
- In general we should have filenames correspond to class names where relevant, & put files in folders according to their package.
- Significantly more commenting would make this much easier to understand, especially with so many complex type parameters, and the heavy use of tuples in place can make the code harder to follow.
- There are a bunch of minor style issues like case classes don't need their parameters to be specified as 'val'

src/test/scala/InterpretTestConstProp.scala
```diff
@@ -0,0 +1,119 @@
...
+
+import java.io.IOException
+import java.nio.file.*
+import java.nio.file.attribute.BasicFileAttributes
+import ir.dsl.*
+import util.RunUtils.loadAndTranslate
+
+import scala.collection.mutable
+
+class ConstPropInterpreterValidate extends AnyFunSuite {
```
l-kent, line 22: This doesn't match the name of the file, it should have a consistent name.

l-kent:
I've merged main into this so it is up-to-date now but there are some outstanding issues with DifferentialAnalysis that should be addressed before this is merged into main:

-DifferentialAnalysis fails in many cases, including cases such as `syscall/gcc_O2:BAP` where the interpreter throws an exception. 
-when DifferentialAnalysis prints text starting with "STDOUT" it is never anything useful and the formatting seems to be broken.

These issues existed before the merge.

This still holds too and would be very very good to have:
> - Significantly more commenting would make this much easier to understand, especially with so many complex type parameters, and the heavy use of tuples in place can make the code harder to follow.

There are a few formatting & organisational issues I intend to fix as well, such as fixing where the file structure doesn't match the package structure.

ailrst:
Since this took two months to merge these changes (and the parameter pr) have been included in https://github.com/UQ-PAC/BASIL/pull/256 and it has been updated with the current main. 

I don't intend to fix any of the issues you mention here.

> DifferentialAnalysis fails in many cases, including cases such as syscall/gcc_O2:BAP where the interpreter throws an exception

This is due to these programs having no main procedure, so the test cases are dubious to begin with. 
The rest of the cases that fail due to the interpreter reaching an error stop are due to incompleteness in the interpreter (usual wrt initialisation, syscalls and missing procedures), which require substantial work to address (such linking a libc). This generally requires a more complete linker emulation, which I am not interested in working on right now. Having these tests pass spuriously would be pointless.

> when DifferentialAnalysis prints text starting with "STDOUT" it is never anything useful and the formatting seems to be broken.

The formattting isn't broken. These show what is printed to stdout which is useful to know when the interepted program is printing to stdout. I'm not interested in making a complete implementation of `sprintf` in Scala to make this more useful in the case of format strings right now.

> Significantly more commenting would make this much easier to understand, especially with so many complex type parameters, and the heavy use of tuples in place can make the code harder to follow.

Extensive documentation is provided here https://github.com/UQ-PAC/BASIL/pull/241/files#diff-4926c58a6f33f8324a92d12cf914e957622fb809f0692aee0777b7dad2d02c7a

In future, don't chuck a 2000 line diff of superficial syntax and formatting changes on the end of my pr please.

I'm closing this as the changes are included in the other pr.

PR #243, opened 2024-09-16
Test time analyses csmith

ailrst:
This profiles the default static analysis pipeline against a set of random C programs generated by [csmith](https://github.com/csmith-project/csmith). 

It outputs the results in CSV and gnuplot format to produce plots like:

large

usage:

```shell
cd examples/csmith && python3 gen.py && cd ../..
mill test.testOnly '*TimeStaticAnalysis'
gnuplot src/test/analysisTiming/plot.gp
xdg-open  src/test/analysisTiming/analysisres.svg
```

The generation script and tests all run singlethreaded so this takes quite a while to run unfortunately. Tests could possibly be multithreaded via `Future.sequence` but I don't know of a good way to avoid the JVM running out of memory by loading everything at once.


examples/csmith/gen.py
```diff
@@ -0,0 +1,71 @@
...
+fn_max = 11
+block_min = 4
+block_max = 8 
+seed_min = 0
+num_rand = 50 
+
+# produces programs that we can't handle
+# cmd = ["csmith", "--seed", str(seed), "--max-block-size", str(max_blocksize), "--max-funcs", str(max_funcs), "--output", f"{fname}.c"]
+
+csmith_safe = "--no-arrays --no-bitfields --no-checksum --no-comma-operators --no-longlong --no-int8 --no-uint8 --no-float --no-math64 --no-inline-function --no-safe-math --no-packed-struct --no-pointers --no-structs --no-unions --no-volatile-pointers --no-const-pointers".split(" ")
```
l-kent, line 15: I'm curious, what can't we handle about each of these things that have been disabled?

ailrst: It was just an incompleteness in steensgaard wrt MemoryLoads, this list is being overly conservative I'm not sure precisely what its disabling that makes it not generate that case, I'll dig up the example again.

l-kent: Please share whatever case broke the Steensgaard, and maybe file an issue for it.

examples/csmith/Makefile
```diff
@@ -0,0 +1,32 @@
+
+SOURCES=$(wildcard *.c)
+SOURCENAMES=$(basename $(SOURCES))
+RELFS=$(addsuffix .relf,$(SOURCENAMES))
+ADTS=$(addsuffix .adt,$(SOURCENAMES))
+OUTS=$(addsuffix .out,$(SOURCENAMES))
+CHECKSUMS=$(addsuffix .checksum,$(SOURCENAMES))
+
+CSMITH_HEADER = "/nix/store/6rawnpny818v7wki6zlr3i0f3phmysyb-csmith-2.3.0/include/csmith-2.3.0/"
```
l-kent, line 9: is hardcoding this a good idea/going to be useful to anyone else?

ailrst: No, It should be at least the normal install location /usr/include/csmith

examples/csmith/gen.py
ailrst: https://github.com/UQ-PAC/BASIL/issues/246

l-kent:
We don't want to disable the MRA-related analyses when regions are disabled because they are still used for resolving indirect calls which is separate functionality.

PR #245, opened 2024-09-16
Tests Cleanup

l-kent:
- Removes many redundant tests, including all the no_plt_no_pic variants, which reduces the execution time by nearly half
- Removes most of examples folder, which was largely very out of date, redundant, and not a good target for tests
- Separated tests containing unresolved indirect calls into a new IndirectCallTests suite that replaces the existing IndirectCallsTests, containing many more tests and more detailed checks
- Consistently uses BigInts for addresses since they can be longer than 32 bits (it may make sense to make them Longs instead)
- Removed deprecated lifting scripts and replaces them with a single script, scripts/lift.sh which can be used to lift single examples
- Removes deprecated MemoryRegionAnalysisMiscTest
- Takes steps towards unifying test suite functionality with BASILTest (replacing TestUtil) now the base for many other tests
- Adds new test suites AnalysisSystemTestsBAP and AnalysisSystemTestsGTIRB which just run all the tests in /correct and /incorrect with the static analysis enabled - this is very useful for finding bugs
- Minor cleanup elsewhere in the tests

I'm still not sure what to do with all the histogram stuff in SystemTests - it doesn't really feel like it all belongs there, but I'm not sure where to put it.

I probably should move the cases where GTIRB successfully pre-resolves indirect calls back into the /correct folder, because they're at least tests of the lifter? 

I'm not sure what to do with correct/syscall/clang_O2 - maybe it should be in IndirectCallTests? It regularly breaks things in interesting ways though and the indirect call isn't really the point?


src/test/scala/test_util/BASILTest.scala
```diff
@@ -0,0 +1,136 @@
...
+
+case class TestConfig(boogieFlags: Seq[String] = Seq("/timeLimit:10", "/useArrayAxioms"),
+                      staticAnalysisConfig: Option[StaticAnalysisConfig] = None,
+                      useBAPFrontend: Boolean,
+                      expectVerify: Boolean,
+                      checkExpected: Boolean = false,
+                      logResults: Boolean = false
+                     )
+
+trait BASILTest extends AnyFunSuite {
```
ailrst, line 19: This unnecessarily restricts you to using a particular suite type in order to use these functions. These can just be defined as functions and imported rather than inherited.

src/test/scala/SystemTests.scala
```diff
@@ -329,18 +258,19 @@ def histoToSvg(title: String, imgWidth: Int, imgHeight: Int, bins: List[Int], mi
...
-  val timeValues = res("verifyTime").map(_.toDouble)
-  val histo = histogram(50, Some(800.0, 1000.0))(timeValues.toSeq)
-  println(histoToSvg("test histogram", 500, 300, histo, 800.0, 1000.0))
+    sourceList.tail.foreach { line =>
+      val cols = line.split(",")
+      headers.zip(cols).foreach((h, v) => res(h).append(v))
+    }
+    val timeValues = res("verifyTime").map(_.toDouble)
+    val histo = histogram(50, Some(800.0, 1000.0))(timeValues.toSeq)
+    println(histoToSvg("test histogram", 500, 300, histo, 800.0, 1000.0))
```
ailrst, line 274: IMO SystemTests is a bit of an abuse of the unit test framework, especially the result aggregating stuff I added on. I'd like to be able to run the tests concurrently, and ideally we'd be able to factor out the part that gets all the test cases out of the file heirachy, the part that runs the tests, and the processing of the results to tables, csv etc., to reuse them for other tests. 

It would be nice to just concurrent map over the a list of TestConfig to run all the tests, and produce a result, (e.g. containing the timing information we use here), and then be able to reduce it however we want.

Scalatest can kind of parallelise by cases rather than suite if you use a different test suite parent class, but I didn't have much success with it and the result aggregation of course makes race conditions.

src/test/scala/SystemTests.scala
```diff
@@ -36,48 +26,52 @@ trait SystemTests extends AnyFunSuite {
 
   val testResults: mutable.ArrayBuffer[(String, TestResult)] = mutable.ArrayBuffer()
 
-  def runTests(programs: Array[String], path: String, name: String, conf: TestConfig): Unit = {
+  private val testPath = "./src/test/"
+
```
ailrst, line 30: Can we set   `Logger.setLevel(LogLevel.WARN)`, printing the globals, proc addresses and then the same list of translation stage log messages for every test really isn't useful and makes it harder to see the failing/passing tests.

.github/workflows/run-examples.yml
```diff
@@ -45,7 +42,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: System Tests
-        run: ./mill test.testOnly '*SystemTests*' || true
+        run: ./mill test.testOnly 'SystemTests*' || true
```
ailrst, line 45: Now we expect SystemTests to pass we should remove the `|| true`, its possibly also worth adding the AnalysisSystemTests as a separate stage.

src/test/scala/test_util/BASILTest.scala
```diff
@@ -0,0 +1,136 @@
...
+  def checkExpected(expectedOutPath: String, BPLPath: String): (Boolean, Boolean) = {
+    val hasExpected = File(expectedOutPath).exists
+    var matchesExpected = true
+    if (hasExpected) {
+      if (!BASILTest.compareFiles(expectedOutPath, BPLPath)) {
+        matchesExpected = false
+        info("Warning: Boogie file differs from expected")
+      }
+    } else {
+      info("Note: this test has not previously succeeded")
```
ailrst, line 83: These info messages are the only thing using functionality from the test suite as far ask I can tell? This could probably be moved back to the SystemTests trait.

ailrst (changes requested):
SystemTests/arrays_simple and syscall/clang_O2 are still failing in system-tests, I suggest we add a suite for TDD style tests, functionally the same as system tests but just containing the borderline test cases. 
Otherwise I have no strong objections.

src/test/scala/SystemTests.scala
l-kent: I agree that the result aggregating stuff is a bit of an abuse of the framework, but it isn't straight forward to replace it so I left it for now.

ailrst: Yeah I think its fine to leave for now

.github/workflows/run-examples.yml
l-kent: We don't expect SystemTests to pass yet - arrays_simple doesn't pass yet for reasons unrelated to indirect calls and that has long been the case.

AnalysisSystemTests is not really worth putting in the GitHub action - it's useful to check for crashes but rather redundant in total. Many of the other test suites may be worth including though.

l-kent:
Which are the borderline cases you mean? arrays_simple/gcc because it still fails? Getting it to pass requires a more sophisticated specification system.

src/test/scala/SystemTests.scala
l-kent: I just replaced most of the Logger.info calls with Logger.debug, only keeping ones to print the very basic test progress.

ailrst:
> Which are the borderline cases you mean? arrays_simple/gcc because it still fails? Getting it to pass requires a more sophisticated specification system.

Yeah, if it requires new features to pass we should put it in a separate test suite to test new features against.

l-kent:
Other tests that don't always pass right now are some of the malloc_memcpy_strlen_memset_free tests which tend to cause Z3 to time out (though I think this is slightly inconsistent). Should they go in their own suite, or in the same one as arrays_simple?

ailrst:
I think we could relax the time limit (by 5 - 10 s?) so they pass reliably, since we don't have so many tests timing out now. They should consistently pass in a reasonable amount of time, clang is 8s on my laptop and gcc < 1s.

l-kent:
Relaxing the time limit doesn't help for incorrect/malloc_memcpy_strlen_memset_free/gcc_O2:BAP but it does for correct/malloc_memcpy_strlen_memset_free/clang_O2:BAP.

l-kent:
correct/malloc_memcpy_strlen_memset_free/clang_O2:BAP still times out for the github action too. Maybe we should just put those tests in their own suite?

ailrst:
Yeah that's probably best

l-kent:
I think this is good now. I've moved the malloc_memcpy_strlen_memset_free tests to their own ExtraSpecTests suite

ailrst (approved):

PR #251, opened 2024-09-23
frontend fixes

ailrst:
- handle unsupported opcodes from https://github.com/UQ-PAC/gtirb-semantics/pull/22
- disambiguate functions with local binding (c `static`) which have the same name but different addresses in the boogie output
- handle relf files with empty relocation table
- some missing relf enum values

### DIFF
--- a/src/main/scala/translating/SemanticsLoader.scala
+++ b/src/main/scala/translating/SemanticsLoader.scala
@@ -100,14 +100,21 @@ class SemanticsLoader(parserMap: immutable.Map[String, Array[Array[StmtContext]]
 
         // LittleEndian is an assumption
         if (index.isDefined && value.isDefined) {
-          Some(MemoryAssign(mem, index.get, value.get, Endian.LittleEndian, size, label))
+          Some(MemoryAssign(mem, index.get, value.get, Endian.LittleEndian, size.toInt, label))
         } else {
           None
         }
-
+      case "unsupported_opcode.0" => {
+        val op = args.headOption.flatMap(visitExpr) match {
+          case Some(IntLiteral(s)) => Some("%08x".format(s))
+          case c => c.map(_.toString)
+        }
+        val comment = " unsupported opcode" + op.map(": " + _).getOrElse("")
+        Some(Assert(FalseLiteral, Some(comment)))
+      }
       case _ =>
-        Logger.debug(s"Unidentified function call $function: ${ctx.getText}")
-        None
+        Logger.error(s"Unidentified function call $function: ${ctx.getText}")
+        Some(Assert(FalseLiteral, Some( " unsupported: " + ctx.getText)))
     }
   }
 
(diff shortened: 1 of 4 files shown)


src/main/scala/translating/SemanticsLoader.scala
```diff
@@ -558,6 +569,8 @@ class SemanticsLoader(parserMap: immutable.Map[String, Array[Array[StmtContext]]
       case "__BranchTaken" => None
       case "BTypeNext" => None
       case "BTypeCompatible" => None
+      case "TPIDR_EL0" => Some(Register(name, 64))
```
l-kent, line 572: What opcodes produce this?

src/main/scala/translating/SemanticsLoader.scala
```diff
@@ -579,15 +592,15 @@ class SemanticsLoader(parserMap: immutable.Map[String, Array[Array[StmtContext]]
...
 
   private def visitIdent(ctx: IdentContext): String = {
     ctx.ID.getText.stripPrefix("\"").stripSuffix("\"")
   }
 
-  private def visitInteger(ctx: IntegerContext): Int = {
-    ctx.DEC.getText.toInt
+  private def visitInteger(ctx: IntegerContext): BigInt = {
+    BigInt(ctx.DEC.getText, 10)
   }
```
l-kent, line 604: What's the purpose of this change?

src/main/scala/util/RunUtils.scala
```diff
@@ -221,6 +221,13 @@ object IRTransform {
     Logger.debug(
       s"[!] Removed ${before - ctx.program.procedures.size} functions (${ctx.program.procedures.size} remaining)"
     )
+    val dupProcNames = (ctx.program.procedures.groupBy(_.name).filter((n,p) => p.size > 1)).toList.flatMap(_._2)
+
+    var dupCounter = 0 
+    for (p <- dupProcNames) {
+      dupCounter += 1
+      p.name = p.name + "$" + p.address.map(_.toString).getOrElse(dupCounter.toString)
+    }
```
l-kent, line 230: This is good to have but may potentially cause problems elsewhere because there are some things (like indirect call resolution) that rely on the names of procedures in a somewhat fragile way. That is something that I want to fix in the indirect call resolution though.

src/main/scala/translating/SemanticsLoader.scala
ailrst: The opcode passed through unsupported_op  as an int isn't representible in a signed Int32.

src/main/scala/translating/SemanticsLoader.scala
ailrst: mrs https://katrinafyi.github.io/aslp-web/5/?opcode=0xd53bd040&bytes=40+D0+3B+D5&asm=mrs+x0%2C+TPIDR_EL0&debug=0

src/main/scala/util/RunUtils.scala
ailrst: Yes its not correct to rely on names it should go through symbol table address -> basil procedure address, we should probably be clear about what the correct unique procedure identifier is in the IR though.

l-kent (approved):

PR #253, opened 2024-09-27
docs: note about mac os

katrinafyi:
For the interest of @utting and @yousifpatti,

Add rudimentary notes about lifting on macos, mostly based on experiences of @ncough. 

Although the cross compiler works (reported by Nick), there are still issues which prevent the basil toolchain working.

I am aware of issues with bap and ddisasm through Nix on Mac. This may or may not be fixed by compiling the packages manually.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -58,6 +58,19 @@ See also: [tool-installation](development/tool-installation.md)
 and [development: building](development/readme.md#building)
 (primarily if you are interested in building BASIL manually).
 
+#### Mac OS
+
+The Mac OS platform (both x86 and arm64)
+uses a different binary format (Mach-O instead of ELF),
+so you will need to install and use a cross-compiler targeting aarch64-unknown-linux-gnu such as the one here:
+https://github.com/messense/homebrew-macos-cross-toolchains
+This means the commands for gcc and readelf might be different from those shown below.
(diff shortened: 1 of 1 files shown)


l-kent:
What are the issues with the toolchain, is it just the issues with Nix?

katrinafyi:
The ongoing issues? I know there are problems running the listed Nix packages. I do not know if they will work outside of Nix.

l-kent:
I know people have gotten BAP to work on Mac by compiling from source (as much of a hassle as that is), though I'm not sure whether that's the ASLp build, but I don't know about DDisasm.

ncough:
I have had BAP, BAP + ASLp and DDisasm all working on macOS in the past, compiling from source in each case. I don't recall exactly what had to be done for each, but I can try digging this up if it would be useful to others.

katrinafyi:
Good news, Github lets us run Actions on macos. Unfortunately, the situation for pac-nix is [not good](https://github.com/katrinafyi/pac-nix/actions/runs/11099645815).
I do not think this will be fixed soon. I guess, in the meantime, the workflow has to be compiling all the tools from scratch.

katrinafyi:
abandoning this PR in favour of https://github.com/UQ-PAC/BASIL/pull/288/

PR #256, opened 2024-10-09
Simplification pass

ailrst:
Adds the flag `--simplify` (which implies `--parameter-form`) which, after running analysis performs partial evaluation and simplification of the IR.  

This usually removes the flag registers and pushes the comparisons into the assume statements where they are used, folds constants through and does some canonicalisation and term simplification.

It sits on top of the interpreter https://github.com/UQ-PAC/BASIL/pull/241 for partial evaluation, and includes improvements to the parameter form which I will copy into that PR https://github.com/UQ-PAC/BASIL/pull/255. When https://github.com/UQ-PAC/BASIL/pull/243 is merged we should use the interpreter to validate the transform using the memory state hash the csmith examples produce.  Currently this is just validated against the system tests, all of which pass with the transform applied. 

this implements:

- lift registers to locals and pass everything via procedure parameters (this required adding parameters to the IR)
- a dynamic single assignment transform via variable renaming, adds an index field to local variables for this purpose so we can easily rename them without
- constant + copy propagation over the IR
- multiple term-rewriting based simplification passes
- a pass to convert 64 bit locals where only the low n bits are accessed to variables of n bits

The simplification/partial eval/copyprop rewrites are applied repeatedly with a fixed number of iterations. 

For an example of what simplifications this produces: https://gist.github.com/ailrst/1b29e85a6ad891810dcc83801b1e2c6a


ailrst:
-  added a bitvector width reduction pass that makes the local variables smaller when only the low n bits of them are ever used in the program (implementation relies on DSA form), this is overapproximate but seems to be an improvement. 
- implemented high level condition recovery in the simplification rewrite system, this is a bit fragile to changes in the rewriter and lifter, but is robust enough to handle both ddisasm and bap
    - some other improvements to the rewriter to simplify unneccessary bv-to-bool conversions and canonicalise towards boolean logic in exprs
- added a basic translation validation for the expression rewriter; currently it just dumps every rewrite to a smt2 file on which you can invoke z3 i.e. `mill run -i examples/conds/conds.gts -r examples/conds/conds.relf -o conds-simp.bpl --simplify && z3 -smt2 simps.smt2| grep -E '^sat' | wc -l` should print 0 (instances of 'sat')

```diff
+|    assume bvsgt32(R0_54, 1000bv32);
-|    assume ((NF == VF) && (ZF == 0bv1));
+|    assume bvsgt32(R0_34, 8bv32);
-|    assume (!((NF == VF) && (ZF == 0bv1)));
+|    assume bvsle32(R0_34, 8bv32);
-|    assume (!((CF == 1bv1) && (ZF == 0bv1)));
+|    assume bvule32(R1_43[32:0], R0_194);
-|    assume (!(!((CF == 1bv1) && (ZF == 0bv1))));
+|    assume bvugt32(R1_43[32:0], R0_194);
-|    assume (!(CF == 1bv1));
+|    assume bvult32(R1_34, R0_158);
```

ailrst:
This improves the DSA, and adds a check for the property. It also adds block coalescing and a debugging flag to perform procedure trimming before running the analysis (so I can test a specific function in cntlm). 

This no longer adds all globals from specs to the boogie, because we can't tell if they are global or have been promoted to local. If they are referenced in spec they must be defined in the actual program, so this shouldn't matter. 
Test `basic_lock_security_write` and `basic_sec_policy_write` are failing now, because the Gamma_R0 in the precondition is optimised out of the program (R0 isn't live at the stat of the program, so I suspect its miscompiled, especially given the signature `int main(int secret)` is undefined behaviour.  `basic_lock_read/clang/GTIRB` is also timing out suspiciously.

utting:
It would be nice to have some top-level comments (e.g. on class SimpExpr) to give an overview of how simplification works and what transforms the 'simplifier' argument can do.  Should this be listed as an extra translation phase in docs/readme.md?

ailrst:
I've added some docs here, mainly api docs on simplifyExpr and AbsInt and the simplification pass summary to the markdown.  https://github.com/UQ-PAC/BASIL/pull/318.

utting:
Merged simplification pass into main, before adding the scalafmt changes, to give a clearer history of the merge.

Issue #194, opened 2024-03-04
Fix conditions in irreducible loop analysis


ailrst:
https://github.com/UQ-PAC/bil-to-boogie-translator/issues/191

The header block does not have conditions added to maintain the original control flow in the transformed loop. 

I'm also unclear if there needs to be conditions to prevent loops between the header and a single dependent block when the flow is meant to switch between dependent blocks.

l-kent:
From memory the way this works (though I haven't confirmed so this may be slightly off) is that we want to replace this:
```
block A
goto entry1

block B
goto entry2
```

with this:
```
block A
headerSwitch := 1
goto header

block B
headerSwitch := 2
goto header

block header
goto condition1, condition2

block condition1
assume headerSwitch == 1
goto entry1

block condition2
assume headerSwitch == 2
goto entry2
```

This can obviously just use bools for two blocks, or be extended to support any number of entrances to a loop.

This is not much of a priority though.

utting:
Irreducible loops are already handled, but this would improve precision.

PR #257, opened 2024-10-10
Irreducible loops cleanup

ailrst:
Addresses https://github.com/UQ-PAC/BASIL/issues/194 and https://github.com/UQ-PAC/BASIL/issues/194

Refactors the analysis to be iterative rather than recursive, work over blocks rather than Commands, adds flow conditions and a cli flag to disable the analysis/transform. 

It also factors out the algorithm's state from class members + mutable collections into an immutable case class, which I think makes the high level api a bit clearer, but the algorithm itself is probably more difficult to understand.


src/main/scala/analysis/IrreducibleLoops.scala
```diff
@@ -55,47 +51,138 @@ class Loop(val header: CFGPosition) {
...
-    val funcEntries = cfg.procedures
-    funcEntries.foreach { funcEntry => traverse_loops_dfs(funcEntry, 1) }
-    loops.values.toSet
+  def identify_loops(cfg: Program): State = {
+    cfg.procedures.toSet.map(_.entryBlock).flatten
+      .foldLeft(State())((st, eb) => (traverse_loops_dfs(st, eb, 1)._1))
   }
 
+
+  private def processVisitedNodeOutgoingEdge(istate: State, edge: LoopEdge, DFSPpos: Int): State = {
```
l-kent, line 90: DFSPpos is never used, is there a reason it exists?

src/main/scala/analysis/IrreducibleLoops.scala
```diff
@@ -110,167 +197,145 @@ class LoopDetector(cfg: Program) {
...
-                  if (iloopHeaders(from) != ih) {
-                    thisLoop.reducible = false
-                    thisLoop.reentries += edge
-                  }
-                } else {
-                  // `from` must be outside the loop
-                  thisLoop.reducible = false
-                  thisLoop.reentries += edge
-                }
+  private def traverse_loops_dfs(_istate: State, _b0: Block, _DFSPpos: Int): (State, Option[Block]) = {
```
l-kent, line 200: Nothing actually uses the second output parameter, what is it for & what does it mean?

l-kent:
This seems fine overall, I'll merge it in once those unused parameters are explained

src/main/scala/analysis/IrreducibleLoops.scala
ailrst: I think I just missed it in the refactor, its been moved into State.

src/main/scala/analysis/IrreducibleLoops.scala
ailrst: Do you mean the `Option[Block]`? I believe this is the first loop header following _b0, this is a holdover from the recursive style of the algorithm which traverses all the successors and tags block that is returned as a header. It is retained in the refactor so the psuedo-return  `retval` that is set in the `FinishProcessNode` state in the explicit call stack form matches the return value of the whole function.

l-kent: The only time traverse_loops_dfs gets called, it has the tuple access ._1 at the end which only returns the State. Do we need to have this second output parameter at all? If there is some reason we want to keep it, there should be a comment documenting what it means.

ailrst: I've just removed it as it isn't really necessary.

PR #262, opened 2024-10-29
Memory Regions in Boogie Output

l-kent:
This makes it so that memory regions produced by the analyses are correctly added to the Boogie output to replace `mem`, as long as the analyses do not produce incorrect or inconsistent results.

@yousifpatti I have cleaned up the code significantly - please look over it to make sure I have not deleted anything important as I have removed a lot of unused code in the analyses.

Before merging, I intend to put the regions functionality behind a flag and clean up a few more things.


yousifpatti (approved):
Looks good. I have copied the Strided Interval Code to a separate branch

Issue #276, opened 2024-11-27
Overlapping Accesses Soundness


l-kent:
As discussed, DataPointer needs to be renamed back to DataLocation, and the extra layer of cells represented by Function should be removed as it was based on a confused understanding of the semantics of the analysis.

The unsoundness issue in handling overlapping accesses in the DSA should also be fixed.

PR #278, opened 2024-11-28
Points to analysis overlapping access

sadrabt:
Overlapping accesses functionality for DSA.
Addresses overlapping accesses to Stack and Global Memory Locations through Addresses, Input/Output Parameters and Indirect Accesses

- [x] Revised Memory Locations for DSA 
- [x] Function Memory Locations represented by cells expanded to their size
- [x] Overlapping Accesses Soundness #276


l-kent:
This seems to have fixed the issues (I'm surprised it was that simple) I was having yesterday but I'm going to test it with some more complex cases to make sure there aren't still any underlying problems.

sadrabt:
> This seems to have fixed the issues (I'm surprised it was that simple) I was having yesterday but I'm going to test it with some more complex cases to make sure there aren't still any underlying problems.

Did your issue's program include function pointers?
The subsequential change is that instead of initialising nodes for function memory locations with a cell of size 0 at offset 0, now DSA initialises a node with a cell of the size of the memory location at offset 0. When the function locations are loaded, they would be merged into the differences between offsets of their pointers in the SSA variable node. For example, in jumptable/clang since function cell sizes were 0 the two function memory cells wouldn't be merged unifying their nodes once the. Now that their sizes are 20 they would be merged at offsets 0 and 8 unifying their nodes. 
The other global/stack/heap have their cell sizes initialised to 0 since DSA learns their sizes through accesses but the function locations are never read/written to.

l-kent:
Yes, it seemed to be that incorrect sizes were causing some weird problems.

l-kent:
I've added a test which demonstrates the remaining issue - if two pointers are loaded together into a single 128-bit register, then it is not correct to merge their pointees as if they are contiguous - the cells should always be collapsed.

l-kent:
You've misunderstood the problem, the "overlapping access" test was fine before.

l-kent:
Although I guess it doesn't make a difference right now since it would never be possible for the cells to point to different things

sadrabt:
I agree it doesn't make a  lot of difference with the current approach, it's simpler and definitely sound to grow the size of the pointer cell to equal the size of the access, it will then merge with it's adjacent cells that are also getting accessed which then in turn also unifies their pointees.

Keeping the multiple pointer cells getting dereferenced as separate is trickier. Currently the pointer cells are kept separate and their pointees are merged and also collapsed. 

I'm fine with either specially if we want to be able to get more precision out of overlapping accesses, we will need to consider the pointer cells separate, so it's good to keep them that way and pick up on any issues if they still exist

l-kent:
For now it's we just get clearer results if we do merge & collapse the cells like that I think, but it's good that you've figured out how to make the more precise version work because we'll likely want to build on that in the future. Sorry for the confusion.

l-kent:
I tried to merge this in but I can't merge this in without breaking things so I will leave it for another time.

sadrabt:
will have a look after the workshop

sadrabt:
I'm not sure about why some examples (secret_write and basic_function_call_caller) fail verification now

src/main/scala/analysis/data_structure_analysis/LocalPhase.scala
```diff
@@ -310,52 +305,53 @@ class LocalPhase(proc: Procedure,
...
-                    unsupportedPointerArithmeticOperation(n, index, Node(Some(graph)).cells(0))
-                }
-              case arg: Variable =>
-                //                assert(varToSym(n).contains(arg))
-                visitPointerArithmeticOperation(n, indexCell, arg, byteSize, false)
-                visitPointerArithmeticOperation(n, Node(Some(graph)).cells(0), arg, byteSize, true)
-              case _ =>
-                ???
-            }
-        }
```
l-kent, line 348: The tests fail because you didn't keep the changed structure here which was important for correctly identifying the access indices, but when I've tried to keep this changed structure, then the DataStructureTests start failing and I'm not sure why that is.

src/main/scala/analysis/data_structure_analysis/LocalPhase.scala
```diff
@@ -310,52 +305,53 @@ class LocalPhase(proc: Procedure,
...
+            assert(!indexCell.node.get.flags.merged)
+
             val valueCells = graph.getCells(n, value)
-            val result = valueCells.foldLeft(addressPointee) { (c, slice) =>
+            val valueCell = valueCells.tail.foldLeft(graph.adjust(valueCells.head)) { (c, slice) =>
               graph.mergeCells(graph.adjust(slice), c)
             }
+
+            if global.isDefined then
+              graph.mergeCells(graph.find(global.get), indexCell)
```
sadrabt, line 327: Is this not doing the same exact merges with indexCell that the other structure is doing? I think the only thing that is changes is that first we merge all the possible valueCells with each other and then merging them with the addressPointee instead of getting the address pointee and then merging it with value cells

l-kent: I don't really remember why I figured out that change in order was necessary but it made a clear difference. 

https://github.com/UQ-PAC/BASIL/tree/points_to_analysis_overlapping_access_merge_test T

This branch has been merged to maintain that structure and with that all the memory region tests pass but it makes some of the tests in DataStructureAnalysisTest fail - can you have a look at that?

sadrabt: Could you please take a look at [this branch](https://github.com/UQ-PAC/BASIL/tree/points_to_analysis_overlapping_access_oldorder). I have reordered the merging operations to match the old way and now all the DSAMemoryREgionSystemTestsBAP pass but the gtirb ones throw an error where accessesEndian is empty

https://github.com/UQ-PAC/BASIL/blob/e68d1a81d81bf2da74856933a4b0ac4daafdfd26/src/main/scala/translating/IRToBoogie.scala#L278

l-kent: Ah thanks, great work. I should be able to fix that, I ran into a similar bug before.

sadrabt: There are memory sections getting created with size 0 because the larges accessed size of them are 0

l-kent: I remember why the change in structure was needed - the old version ignored the case where a Literal is written to a memory location. In practice this is only ever 0 I think? But it meant that cells weren't updated properly for those accesses.

l-kent: Yeah, the largestAccessedSize appears to be incorrect in those cases, but I haven't figured out why yet.

l-kent: An example is secret_write/clang:O2:GTIRB - the cell representing x doesn't have its largestAccessedSize updated and remains at 0, even though x is clearly accessed in the program. It may be because the access to x is storing 0 in it?

l-kent: https://github.com/UQ-PAC/BASIL/blob/7fe1979cb03b7c6d44cd212fd38f39caa3a26958/src/main/scala/analysis/data_structure_analysis/LocalPhase.scala#L347-L354

Is it that the largestAccessedSize of a cell only gets updated during a store if this part is reached?

sadrabt: Okay I think the problem was that if a literal is assigned that DSA knows is not an address to a global, it ignored it but now it grows the memory region it was assigned to that size

l-kent: It should grow the cell but it doesn't, I think?

sadrabt: I don't know why it worked for the adt version without that tho

sadrabt: > It should grow the cell but it doesn't, I think?

Yeah this def was the case

l-kent: That seems to have fixed it, great! I don't know why there was a difference either, I'll give it a closer look next week to see if there's anything obvious there.

Issue #249, opened 2024-09-23
Memory Regions in IL


sadrabt:
Replace memory accesses in the IL with correct DSA nodes

PR #279, opened 2024-11-28
Memory regions in Boogie using Data Structure Analysis

l-kent:
Implements #249
- Adds a `--memory-regions dsa` option to insert regions based on the Data Structure Analysis results into the Boogie output
- The existing version of memory region insertion can now be used by using `--memory-regions mra` 
- Fixes issue with invalid characters appearing in DSA .dot file output
- External functions/objects listed in the symbol table are now handled correctly in the DSA, with the pointers being correctly treated as having size 8, renamed to indicate that they are pointing to the external function, and they now point to a new `ExternalLocation` which represents the abstract external function or object
- Adds `DSAMemoryRegionSystemTestsBAP` and `DSAMemoryRegionSystemTestsGTIRB` test suites which run all SystemTests with the  `--memory-regions dsa` option set

The DSA-based memory regions are currently able to handle all existing test cases successfully, though it likely does not handle anything involving overlapping accesses successfully yet since the DSA implementation there is not yet finished. It tracks the memory regions by creating a new Map, `accessIndexToSlice`, which maps from `MemoryLoad`s and `MemoryStore`s to `Slice`s representing the indices of the accesses. This mapping is used as the basis for regions, with distinct cells representing distinct memory regions.


src/main/scala/analysis/RegionInjector.scala
```diff
@@ -145,16 +163,148 @@ class RegionInjector(program: Program, mmm: MemoryModelMap) {
...
+
+  private def transformMemorySections(program: Program): Unit = {
+    // if any addressRange entries point to a mergedRegion: grab the bytes
+
+    val dsg = DSATopDown(program.mainProcedure)
+
+    // need to check if nodes match too
+
+    dsg.globalMapping.foreach { (range, field) =>
+      val cell = dsg.find(field.node.cells(0))
```
sadrabt, line 266: field.node.getCell(field.offset), value of field offset may have changed in bottom up or top town phase

src/main/scala/analysis/RegionInjector.scala
```diff
@@ -145,16 +163,148 @@ class RegionInjector(program: Program, mmm: MemoryModelMap) {
...
+      if (mergedRegions.contains(cell)) {
+        program.initialMemoryLookup(range.start) match {
+          case Some(section) =>
+            val size = (range.end - range.start).toInt
+            val bytes = section.getBytes(range.start, size)
+            // should probably check that region is entirely contained within section but shouldn't happen in practice?
+            val newSection = MemorySection(mergedRegions(cell).name, range.start, size, bytes, section.readOnly, Some(mergedRegions(cell)))
```
sadrabt, line 273: each global may consistent of multiple disjoint cells each representing different regions. The size of memory section should be the largest accessed size of the cell?

src/main/scala/analysis/RegionInjector.scala
l-kent: How do I correctly get the cell that a globalMapping entry corresponds to then?

src/main/scala/analysis/RegionInjector.scala
l-kent: When is the `size` of a `Cell` different to the `largestAccessedSize`?

I did realise that my approach doesn't currently take into account the case where an access should have two cells associated with it (due to overlapping accesses) but it was not really possible to figure that out yet since that hasn't yet been implemented on this branch..

src/main/scala/analysis/RegionInjector.scala
sadrabt: dsg.find(field.node.getCell(field.offset)) should give you the cell. 

The actual offset which field.offset corresponds to in the node returned by dsg.find is slightly tricker. You can call dsg.find(field.node) which returns a field. The corresponding offset in the new node is
 field.offset + dsg.find(field.node).offset

sadrabt: You are correct that dsg.find(field.node.getCell(field.offset))  returns a cell. 

But we also track slices which are internalOffsets into the cells. 

If we have 
cell = dsg.find(field.node.getCell(field.offset))
cellStart= cell.offset
cellEnd = cell.offset + cell.largestAccessedSize

then 
field.offset + dsg.find(field.node).offset
will give you a number between cellStart and cellEnd

l-kent: Ah ok, I don't think I need to care about the internalOffsets here so dsg.find(field.node.getCell(field.offset)) should work?

sadrabt: Yeah it should give you the right cell

sadrabt:
The rest looks good to me

sadrabt:
DSA-Regions extra Constraints #280

PR #282, opened 2024-12-05
cntlm example without duktape

ailrst:
This adds a lifted build of cntlm with the PAC (proxy auto configuration) feature disabled and duktape removed, from the release here https://github.com/UQ-PAC/cntlm/releases/tag/build-2-glibc

This reduces its size by 20% and makes it much more manageable for basil to analyse.


l-kent:
Is there a reason to include the version with duktape in this? We already have that in the repo as-is

ailrst:
Just so that it is directly comparable to the example without duktape. The existing versions in the repo don't have any information attached about how they were compiled or with which sourcecode.

l-kent:
Do we want the existing versions at all? Do we need a version with duktape at all? I don't think we're ever going to be able to do anything useful with it.

ailrst:
I found duktape useful for stress testing the (intraprocedural) simplification-pass analyses as it contained a lot of complex behaviour and variety. I think its useful as a worst-case test case for analyses.

I'm not sure if we need the existing versions, but I think its good to keep them as they are known to lift successfully and there have been some changes to the lifter since. E.g. the gtirb here doesn't load because theres a new intobits pattern (fixed in simplification-pass branch).

l-kent:
If there's been a breaking change to the lifter, that's really something we want to address in a separate pull request so we can get it merged in faster?

ailrst:
Yeah. But it was trivial to fix in the simplification-pass branch and isn't simple to cherrypick out the change that fixes it.

l-kent:
What's the point of committing this now then if the main branch can't actually use it?

ailrst:
- To use to write the fix to main
- To merge it into simplification pass which already works and is downstream of main
- Because the BAP example already works

l-kent:
If the old version will be redundant once this is merged and the parser is updated for the GTIRB changes, it'd be simpler to remove the old version and update the parser all together here?

PR #288, opened 2024-12-13
Reproducible Docker build for test cases in src/test

katrinafyi:
This has been my pet project for a while and my continuous focus for the last week and a bit. It has been through a fair bit of beta-testing to work out some kinks. Hopefully, it is a well-prepared change.

The effect of this PR is to remove all of the committed lifter outputs from src/test/ and replace them with a system to compile and lift them in a reproducible way. This is done with a Docker image which contains fixed versions of the required tools, a number of scripts to tie it together, and hashes recorded within the repository in place of the committed files.

Please see https://github.com/UQ-PAC/BASIL/blob/nix-docker-build/src/test/readme.md for documentation about the Docker build. If you want to try out the changes, that readme should have all the steps needed (let me know if anything is not clear).

**Note!** The SystemTests are not passing. I will need some help to work these out, since the changes to ADT/GTIRB files might be very large since they were last updated in the repository.

There are obvious benefits:
- Clear provenance for lifted examples. Versions of tools are clearly tracked by the Docker image, generated by Nix.
- Consistency across different operating systems. The existing makefile implicitly assumes some version of Ubuntu and a x86_64-linux system.
- Reduced repository size and preventing further increases as the binary files are updated.

We foresaw the following pain points and have worked to reduce them:
- Strong dependency on Docker for CI and running tests. We avoid this by uploading a tarball of the Docker-compiled files and adding a Makefile rule which simply downloads and extracts them. This is fast and suitable for CI as well as users who do not need to change test cases. We hope this will not worsen the Basil onboarding experience.
- Accidentally mixing locally-compiled files and Docker-compiled files when making the hashes. The md5sum-updating rules are written to ensure they are run within Docker. To avoid files being built without Docker, then activating Docker and updating hashes, we track whether each intermediate file was made inside or outside Docker. Updating hashes is only allowed when they are all from Docker.

There are also drawbacks / limitations.
- Lifted files differences are not visible in web view. This may be a benefit as it cleans up diffs. However, it is more troublesome to find the diffs if you want to see them. 
- Size and complexity of Docker image. It is a little bulky to handle and takes up sizable disk space. Hopefully, this burden is lessened by the pre-built tarball. Unused Docker images can also be pruned frequently from your computer.
- Dependency on Docker to start developing test cases. This is true, but I think the trade-off is worthwhile to get consistency. Building using local tools is still possible using the old Makefile infrastructure.
- To ensure identical file hashes, BAP adt and bir files are normalised by replacing variable Tid addresses with deterministic ones, starting from %00000000 and incrementing. Basil might have made assumptions on these IDs.
- The use of some scripts here might have bugs. We make efforts to program defensively and have tested on Linux and MacOS, but some bugs might remain.
- The Makefiles inherit the existing limitation that test cases must be a single C file. This can be extracted to only handle the lifting / hashing part, to make this more flexible and usable with binaries such as cntlm.

Anticipated questions:
- Why not Git LFS? LFS only solves one part of the problem (the distribution of binaries). The versioning of compilers and tools is still unknown. Github also has tight limits on LFS storage and bandwidth.
- Why not Docker by itself instead of Docker via Nix? I think that it would be fairly difficult to make a traditional Dockerfile build have the level of consistency which we want. If the Docker build is not repeatable, it has much the same problems as the existing committed binaries, just in GHCR instead of the git repository. Nix is already used to distribute some of the lifter tools, and there is Nix tooling to easily bundle these into a Docker image which can be built by anyone with Nix.
- Okay, but who knows Nix? Sadly not too many. People should learn! However, for the purposes of building this Docker image, it pulls directly from the packages already in Nix. Updating these packages is already automatic, so all you have to do here is update the docker-flake reference.
- Why not what we already had? See benefits.

Catalogue of changed files:

To make the changes easier to see, I have made a diff which excludes all the generated file changes: https://gist.github.com/katrinafyi/23f9a7de5b99cf1843814ccb72302e18.

In summary,
- Most of the changes are to the Makefiles in src/test, including the .mk files in make/.
- Helper scripts in make/docker-helper.sh and make/bap-normalise.py.
  - Maybe bap-normalise should live in the bap-aslp Nix package. That way, it would also be versioned by the Docker image.
- make/docker-flake.txt pins the Docker flake to a particular pac-nix commit (and hence the packages of that commit). src/test/docker-contents.txt lists all Nix derivations within the Docker image.
- Each test case's compiler directory, e.g. correct/arrays_simple/clang, has a new md5sums file which tracks the hashes of a.out, adt, bir, relf, and gts files.
- src/test/compiled.md5sums collates the md5sums from all subdirectories, for easier checking
- src/test/compiled.url.txt is a URL to an online file holding the pre-compiled files, in a .tar.bz2 file. This is extracted on top of src/test to populate the files.
  - maybe we should hash the tarball, in case our file host ever becomes malicious.
- src/test/readme.md extensively documents this system. docs/... is updated to point to this.

Orthogonal but related changes:
- Makefile fix and improvements to be more robust and customisable (e.g. lifting individual tests or folders). 
- Restructuring DSA and irreducible loops folders to be lifted by the makefile.

**TODO**
- [x] update expected files
- [x] ci tarball creation and download
- [x] docker-hash in one directory, next to big md5sum. name could be docker-contents.txt or similar
- [ ] problem: since makefiles wildcard directories, it is easy for leftover empty directories to cause unknowable errors like "clang no input files". maybe fixed by writing makefile expression to filter for directories containing C files.
- [ ] document above problem


ailrst:
Thanks for doing this, I am keen for this approach to get merged as it should allow us to repeatably test with larger examples, (e.g. from svcomp and cntlm) without blowing out the repository size. 

I will have a look at why the tests might have started failing as this pr shouldn't be changing anything.

katrinafyi:
The status of the tests after disabling hardening:
```python
Run completed in 3 minutes, 46 seconds.
Total number of tests run: 410
Suites: completed 2, aborted 0
Tests: succeeded 353, failed 57, canceled 0, ignored 0, pending 0
*** 57 TESTS FAILED ***
1 targets failed
test.testOnly 57 tests failed:
```

In GTIRB, the only tests failing are gcc_pic ones. Of the GTIRB gcc_pic tests, 20/31 fail.

In BAP, 23/31 gcc_pic are failing. 14 tests with other compiler configurations are also failing:
```bash
$ grep 'BAP,false' full-testresult-BAP.csv | grep -v gcc_pic | cut -d, -f1
functionpointer/clang:BAP
functionpointer/clang_pic:BAP
functionpointer/gcc:BAP
functionpointer/gcc_O2:BAP
indirect_call/clang:BAP
indirect_call/clang_pic:BAP
indirect_call/gcc:BAP
jumptable2/clang:BAP
jumptable2/clang_O2:BAP
jumptable2/clang_pic:BAP
jumptable2/gcc:BAP
jumptable2/gcc_O2:BAP
syscall/clang_O2:BAP
syscall/gcc_O2:BAP
```
[full-testresult-BAP.csv](https://github.com/user-attachments/files/18123041/full-testresult-BAP.csv)
[full-testresult-GTIRB.csv](https://github.com/user-attachments/files/18123042/full-testresult-GTIRB.csv)

l-kent:
```
functionpointer/clang:BAP
functionpointer/clang_pic:BAP
functionpointer/gcc:BAP
functionpointer/gcc_O2:BAP
indirect_call/clang:BAP
indirect_call/clang_pic:BAP
indirect_call/gcc:BAP
jumptable2/clang:BAP
jumptable2/clang_O2:BAP
jumptable2/clang_pic:BAP
jumptable2/gcc:BAP
jumptable2/gcc_O2:BAP
syscall/clang_O2:BAP
syscall/gcc_O2:BAP
```
None of these are tests currently in the repo except syscall/gcc_O2, which is likely explained by the difference in gcc version. The current lifting scripts haven't been set up to exclude specific cases from being lifted for BAP or GTIRB but they probably should be since some test cases are BAP or GTIRB only for various reasons.

src/test/Makefile
```diff
@@ -1,12 +1,47 @@
...
+	set -u; $$DOCKER_CMD hash > docker-contents.txt  # before compessing, make sure docker-contents.txt is up to date.
+	md5sum --quiet -c compiled.md5sum  # before compressing, make sure our files match expected hashes.
+	list=`mktemp`; cut -d' ' -f3 compiled.md5sum > $$list && tar caf $(TARBALL) -T $$list && rm $$list
+	sha1sum $(TARBALL)
+
+.PHONY: extract
+extract:
+	# check existing file, otherwise download fresh copy.
+	{ tail -n1 compiled.url.txt | sha1sum -c - ; } \
+		|| curl "$$(head -n1 compiled.url.txt)" -o $(TARBALL)
```
l-kent, line 41: I get 
```
curl: (3) URL using bad/illegal format or missing URL
```

l-kent:
It would be nice to give the md5sums files their own extension, mostly for github filtering purposes.

src/test/Makefile
```diff
@@ -1,12 +1,47 @@
-TARGETSS := all verify clean cleanall cleanlift recompile json cleanjson cleangts gts
+TARGETSS := all verify repro-stash repro-check md5sum-check md5sum-update clean cleanall cleanlift recompile json cleanjson cleangts gts
 
-SUBTARGETS = $(wildcard correct/*/ incorrect/*/)
-.PHONY : $(TARGETSS) $(SUBTARGETS) correct incorrect
+# subdirectories of src/test. to be entered into by this makefile.
+DIRS := correct incorrect extraspec_correct extraspec_incorrect indirect_calls memory_regions procedure_summaries dsa irreducible_loops
```
l-kent, line 4: dsa and irreducible_loops are not currently set up to be lifted with the same scripts as everything else

src/test/Makefile
katrinafyi: that's surprising. does the compiled.url.txt exist? are you in the src/test directory?

src/test/Makefile
katrinafyi: i rearranged their directory structure so they worked. if this is not appropriate, i can revert those changes.

l-kent: I don't think it's broken those tests as much as just created a bunch of extra files that are not used by those test suites. I can't say whether those new files would pass or not but most would be redundant.

src/test/Makefile
l-kent: I think it's just git automatically & incorrectly adjusting the line endings.

src/test/Makefile
katrinafyi: I think the paths do need updating, because the .adt/etc files have moved into new compiler-specific subdirectories. I'm also not sure which compiler was used to produce the previous files.

l-kent: Yes, but you haven't removed the old files so that hasn't actually broken anything.

I can safely say that the DSA tests will not pass if you just switch out the input files though.

I don't know which compiler was used to produce the previous files either, those aren't my tests, but most of the compiler option-specific versions for the DSA tests and the irreducible loops tests would be redundant.

katrinafyi: Oops, you're right, I forgot to delete the old files. I'll undo those changes for now so they can be given more attention.

l-kent:
> In GTIRB, the only tests failing are gcc_pic ones. Of the GTIRB gcc_pic tests, 20/31 fail.
> 
> In BAP, 23/31 gcc_pic are failing. 

I'm not seeing this with the provided pre-compiled files.

katrinafyi:
Most of the gcc_pic tests were fixed after by adding `-pie` to the gcc_pic compiler flags. I think this non-f flag affects the linker behaviour? But I am not quite sure whether it makes sense to have it there.

After this change, I think the only failures are the BAP ones you pointed out.

https://github.com/UQ-PAC/BASIL/actions/runs/12315042870/job/34372404783

l-kent:
syscall/gcc_O2:BAP is the only test that fails that is supposed to be included, and it does seem to be significantly different.

I haven't looked at the test suites beyond SystemTestsBAP and SystemTestsGTIRB yet though.

l-kent:
If you could share a few of the gcc_pic tests that didn't work then that would be good, I'd like to have a look at those to see why they don't pass.

katrinafyi:
these are the tests which failed in [the previous GHA run](https://github.com/UQ-PAC/BASIL/actions/runs/12314284422/job/34370065134):
<details>

```
test.testOnly 57 tests failed: 
  SystemTestsBAP correct/basic_lock_security_write/gcc_pic:BAP
  SystemTestsBAP correct/basic_arrays_write/gcc_pic:BAP
  SystemTestsBAP correct/secret_write/gcc_pic:BAP
  SystemTestsBAP correct/basic_lock_security_read/gcc_pic:BAP
  SystemTestsBAP correct/basic_function_call_reader/gcc_pic:BAP
  SystemTestsBAP correct/no_interference_update_x/gcc_pic:BAP
  SystemTestsBAP correct/basic_assign_assign/gcc_pic:BAP
  SystemTestsBAP correct/basic_lock_read/gcc_pic:BAP
  SystemTestsBAP correct/basic_sec_policy_read/gcc_pic:BAP
  SystemTestsBAP correct/cjump/gcc_pic:BAP
  SystemTestsBAP correct/basicassign_gamma0/gcc_pic:BAP
  SystemTestsBAP correct/basic_assign_increment/gcc_pic:BAP
  SystemTestsBAP correct/no_interference_update_y/gcc_pic:BAP
  SystemTestsBAP correct/initialisation/gcc_pic:BAP
  SystemTestsBAP correct/basic_function_call_caller/gcc_pic:BAP
  SystemTestsBAP correct/functionpointer/gcc_O2:BAP
  SystemTestsBAP correct/functionpointer/gcc:BAP
  SystemTestsBAP correct/functionpointer/clang:BAP
  SystemTestsBAP correct/functionpointer/gcc_pic:BAP
  SystemTestsBAP correct/functionpointer/clang_pic:BAP
  SystemTestsBAP correct/basic_lock_unlock/gcc_pic:BAP
  SystemTestsBAP correct/syscall/gcc_O2:BAP
  SystemTestsBAP correct/syscall/clang_O2:BAP
  SystemTestsBAP correct/basic_loop_assign/gcc_pic:BAP
  SystemTestsBAP correct/indirect_call/gcc:BAP
  SystemTestsBAP correct/indirect_call/clang:BAP
  SystemTestsBAP correct/indirect_call/gcc_pic:BAP
  SystemTestsBAP correct/indirect_call/clang_pic:BAP
  SystemTestsBAP correct/jumptable2/gcc_O2:BAP
  SystemTestsBAP correct/jumptable2/clang_O2:BAP
  SystemTestsBAP correct/jumptable2/gcc:BAP
  SystemTestsBAP correct/jumptable2/clang:BAP
  SystemTestsBAP correct/jumptable2/gcc_pic:BAP
  SystemTestsBAP correct/jumptable2/clang_pic:BAP
  SystemTestsBAP correct/basic_sec_policy_write/gcc_pic:BAP
  SystemTestsBAP correct/using_gamma_conditional/gcc_pic:BAP
  SystemTestsBAP correct/ifglobal/gcc_pic:BAP
  SystemTestsGTIRB correct/basic_lock_security_write/gcc_pic:GTIRB
  SystemTestsGTIRB correct/basic_arrays_write/gcc_pic:GTIRB
  SystemTestsGTIRB correct/secret_write/gcc_pic:GTIRB
  SystemTestsGTIRB correct/basic_lock_security_read/gcc_pic:GTIRB
  SystemTestsGTIRB correct/basic_function_call_reader/gcc_pic:GTIRB
  SystemTestsGTIRB correct/no_interference_update_x/gcc_pic:GTIRB
  SystemTestsGTIRB correct/basic_assign_assign/gcc_pic:GTIRB
  SystemTestsGTIRB correct/basic_lock_read/gcc_pic:GTIRB
  SystemTestsGTIRB correct/basic_sec_policy_read/gcc_pic:GTIRB
  SystemTestsGTIRB correct/cjump/gcc_pic:GTIRB
  SystemTestsGTIRB correct/basicassign_gamma0/gcc_pic:GTIRB
  SystemTestsGTIRB correct/basic_assign_increment/gcc_pic:GTIRB
  SystemTestsGTIRB correct/no_interference_update_y/gcc_pic:GTIRB
  SystemTestsGTIRB correct/initialisation/gcc_pic:GTIRB
  SystemTestsGTIRB correct/basic_function_call_caller/gcc_pic:GTIRB
  SystemTestsGTIRB correct/basic_lock_unlock/gcc_pic:GTIRB
  SystemTestsGTIRB correct/basic_loop_assign/gcc_pic:GTIRB
  SystemTestsGTIRB correct/basic_sec_policy_write/gcc_pic:GTIRB
  SystemTestsGTIRB correct/using_gamma_conditional/gcc_pic:GTIRB
  SystemTestsGTIRB correct/ifglobal/gcc_pic:GTIRB
  ```
  
  </details>
  
  these are the compiled files at that time:
[u6jsqr.tar.gz](https://github.com/user-attachments/files/18144969/u6jsqr.tar.gz)

**Comment by l-kent** [comment]
Yeah it seems like it was handling relocations in a completely different way without the `-pie` flag so it wouldn't have been possible to handle. I'm not really sure what the difference that caused that is but I'm glad it's not a problem for BASIL at least.

**Comment by katrinafyi** [comment]
I should mention that the reason I kept #289 separate is that in order to fix the CI, I changed the GHA to execute without Docker. It uses a Docker at the moment which I assume has the build environment needed. 

Installing all the dependencies and boogie manually is not so bad, but you might have versions changing out from underneath you (notably z3 or boogie versions) (but Ubuntu LTS updates packages pretty slowly). If this is not something to worry about, I can merge the CI changes in with this PR.

**Comment by ailrst** [comment]
syscall/gcc_O2:BAP was only ever passing vacuously so I think we should just disable it. The optimizer made fork() the main function, which has no implementation as it is a library function. Since this example has no spec it was only verifying the transitivity of the default rely condition mem == old(mem). Generally I err to deleting the syscall example entirely as there is nothing to verify there and there isn't enough code for it to really be testing the frontend. The rest of its cases (in SystemTests) are only really testing that the lifter emits a call to fork() rather than an indirect call.

**Comment by l-kent** [comment]
It wasn't really vacuous, it was useful to test that we don't crash for the weird edge case where the main function is just a stub that bridges to an external call, and that has pretty often caught bugs.

It's no longer verifying with these changes because the lifter setup with docker is not producing equivalent files for this one case - it's now the same as syscall/clang_O2:BAP where there is an unresolved indirect call.

**Comment by l-kent** [comment]
Like the Boogie verification was vacuous, sure, but the test itself has been useful.

**Comment by l-kent** [comment]
> Installing all the dependencies and boogie manually is not so bad, but you might have versions changing out from underneath you (notably z3 or boogie versions) (but Ubuntu LTS updates packages pretty slowly).
> 
> If this is not something to worry about, I can merge the CI changes in with this PR. Otherwise, I can look closer at why it fails inside Docker.

It is possible that versions changing out from underneath us could cause issues - Boogie is tested with Z3 version 4.8.8 (which is about 4 years old) and there have been times where using a different Z3 version has caused unexpected regressions in Boogie. I don't know which versions the current Docker setup for the Github action has though. This is unlikely to cause problems very often given how slowly Ubuntu LTS packages are updated as you say, but it would be good to just settle on particular Boogie + Z3 versions rather than leaving it up to whatever is in the Ubuntu repo?

**Comment by katrinafyi** [comment]
Do you have any guesses as to which gcc version compiled the old syscall/gcc_O2:BAP?

**Comment by l-kent** [comment]
I think it was 11.3.0

**Comment by l-kent** [comment]
The docker-helper.sh script seems to assume podman is being used.

**Comment by katrinafyi** [comment]
>  assume podman 

It does, but it should be overridable. I'll update the docs to mention this.

**Comment by l-kent** [comment]
For 
```
+ exec docker pull --platform linux/amd64 ghcr.io/uq-pac/basil-tools-docker:flake-014a-ef9c2374
```
when running `docker-helper.sh pull` I get 
> Error response from daemon: manifest unknown

**Comment by katrinafyi** [comment]
> flake-014a-ef9c2374

You seem to have \r\n line endings in the src/test/make/docker-flake.txt file

**Comment by l-kent** [comment]
Oh I thought I'd fixed that but I guess not

**Comment by katrinafyi** [comment]
Just checking, do you use git through WSL or through just Windows? WSL, I'd think, wouldn't have this problem.

**Comment by katrinafyi** [comment]
I was planning to do that, I was just curious.

**Comment by l-kent** [comment]
At the very least, all of the following probably should have eol=lf but I'm probably missing something since I'm still having an issue with `make md5sum-check`
```
Makefile
*.sh 
*.mk
*.md5sum
```

Docker works to lift everything, it's just incredibly slow (unsurprisingly). The md5 sums all match.

There's still a bunch of test cases where it needs to be added to the config files not to compile the BAP or GTIRB variants - it's mostly GTIRB variants in the indirect_calls folder I think. 

The .expected files will need to be updated before merging since they're still for the earlier version I think?

**Comment by katrinafyi** [comment]
What happens with md5sum-check?

**Comment by l-kent** [comment]
```
md5sum: 'correct/arrays_simple/clang/a.out'$'\r': No such file or directory
: FAILED open or read/clang/a.out
```
but I can't track down where the `\r` is being introduced into it easily

File: src/test/make/docker-helper.sh [hunk 3]
```diff
@@ -0,0 +1,171 @@
...
+
+elif [[ "$1" == start ]]; then
+  # starts an instance of the docker image.
+  set -x
+  exec $DOCKER run $DOCKER_PLATFORM -v"$GIT_ROOT:$GIT_ROOT" --rm -td --user $DOCKER_USER --name $unique_container $unique_image
+
+elif [[ "$1" == stop ]]; then
+  # starts the instance of the docker image.
+  set -x
+  exec $DOCKER rm -f -t 1 $unique_container
```
l-kent, line 92: This doesn't work with docker, only podman, since docker doesn't have a -t flag for rm

katrinafyi:
> issue with make md5sum-check

I can't encounter this, even after forcing git to use crlf by default. You can try (very carefully) resetting the src/test directory: `rm -rf src/test && git checkout src/test`. Otherwise, does `make extract` work? how about `md5sum -c compiled.md5sum`?

l-kent:
```md5sum -c compiled.md5sum```

worked fine once I fixed the line endings in compiled.md5sum, which is how I knew the md5sums were fine, but the `\r`s were still getting into `make md5sum-check` somewhere. I'll try it again

katrinafyi:
@l-kent For correct/syscall/gcc_O2, can I trouble you to build the test case with `-O2 -v` , then post the log output and binary? I didn't have success with either gcc 11, 12, or 13. It would be good to do this on an OS as similar to the one you think was used before.

katrinafyi:
Diffs in `./mill test` output as of time of this comment: https://gist.github.com/katrinafyi/01181184e153a75a8c0d710ee089c6da

We fail 54 tests compared to 30 failing in main. I'll take a look, maybe it's BAP-related.

PR #293, opened 2024-12-19
Support new lifter unsupported opcode format

ailrst:
Updates frontend to support the format in  https://github.com/UQ-PAC/gtirb-semantics/pull/22

Test code: [gate_server_example.tar.gz](https://github.com/user-attachments/files/18194266/gate_server_example.tar.gz)

Also adds a case to loading `cvt_bool_bv` that wasn't covered previously.

### DIFF
--- a/src/main/scala/translating/GTIRBLoader.scala
+++ b/src/main/scala/translating/GTIRBLoader.scala
@@ -13,7 +13,7 @@ import scala.collection.mutable.ArrayBuffer
 import com.grammatech.gtirb.proto.Module.ByteOrder.LittleEndian
 import util.Logger
 
-class GTIRBLoader(parserMap: immutable.Map[String, Array[Array[StmtContext]]]) {
+class GTIRBLoader(parserMap: immutable.Map[String, List[Either[(String, String), List[StmtContext]]]]) {
 
   private val constMap = mutable.Map[String, IRType]()
   private val varMap = mutable.Map[String, IRType]()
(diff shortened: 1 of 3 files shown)


src/main/scala/util/RunUtils.scala
```diff
@@ -143,16 +163,20 @@ object IRLoading {
...
-            case _ => ""
+            case o => o.toString
           }
           Logger.error(s"""Semantics parse error:\n  line: $line\n$extra""")
           throw e
       }
     }
 
-    val parserMap = semantics.map(_.map((k: String, v: Array[Array[String]]) => (k, v.map(_.map(parse_insn)))))
+    val parserMap : Map[String, List[Either[(String,String), List[StmtContext]]]] =
```
l-kent, line 173: (String, String) should probably be a case class instead of a tuple to make the semantics of it clearer

l-kent:
Would it be cleaner if `Result.value` was a `List[StmtContext]` and so then `parserMap` could be a `Map[String, List[InsnSemantics]]`? That seems like it would streamline some of the conversion steps?

ailrst:
That makes it a lot cleaner, thanks

src/main/scala/translating/GTIRBLoader.scala
```diff
@@ -30,20 +35,36 @@ class GTIRBLoader(parserMap: immutable.Map[String, Array[Array[StmtContext]]]) {
...
+    for (instsem <- instructions) {
       constMap.clear
       varMap.clear
 
-      for ((s, i) <- instruction.zipWithIndex) {
-        
-        val label = blockAddress.map {(a: BigInt) =>
-          val instructionAddress = a + (opcodeSize * instructionCount)
-          instructionAddress.toString + "$" + i
+      val s = instsem match {
```
l-kent, line 42: is this `s` used for anything at all? it seems to just be of type `Unit` so the `val` definition could be removed completely?

l-kent:
This seems fine, but I'm wondering if we want to print something via the logger to let the user know more immediately that there was a lifting issue?

PR #301, opened 2025-01-21
Add block info, interproc solver, analysis prettyprinter

ailrst:
- add loop information to the IR so you can easily implement widening heuristic in the AbstractDomain.join function without proper solver support

```scala
class Block {
  var inLoop: Set[Loop]
  def isLoopHeader () : Bool
  def isLoopParticipant () : Bool
}
```

- Add simple interprocedural solver which takes a fixed-point over the call graph (very naive inefficient implementation but seems OK for definitely-returns analysis in cntlm-noduk). 

This applies an interprocedural analysis to each procedure, abstracts the results to a summary for the procedure-level, and takes this to a fixed point.

To use this define a local analysis `AbstractDomain[LocalDomain]`, and a summary generator:

```scala
trait ProcedureSummaryGenerator[L, LocalDomain] extends GenericAbstractDomain[Procedure, Procedure, L] {
  def localTransferCall(l: LocalDomain, summaryForTarget: L, p: DirectCall) : LocalDomain
  def updateSummary(prevSummary: L, p: Procedure,  resBefore: Map[Block, LocalDomain], resAfter: Map[Block, LocalDomain]) : L
}
```

And pass them to the solver:

```scala
class interprocSummaryFixpointSolver[SummaryAbsVal, LocalAbsVal, 
  A <: AbstractDomain[LocalAbsVal]](localDomain : A, 
    sg: ProcedureSummaryGenerator[SummaryAbsVal, LocalAbsVal]) {
   ...
    solveProgInterproc(p: Program, backwards: Boolean): Map[Procedure, SummaryAbsVal]
}
```

This is a bit confusing possibly because it overrides the local AbstractDomain.transfer for directCalls to inject the summary provided by localTransferCall. 

- Add analysis result printing api to `PrettyPrinter`:

```scala
pp_prog_with_analysis_results(beforeLive, afterLive, ctx.program, x => s"Live vars: ${x.map(_.name).toList.sorted.mkString(", ")}")
```

```
    block www_authenticate_4218176_basil_return [
      // Live vars: R0, R29, R31
      return (R0:bv64, R29:bv64, R31:bv64)
      // Live vars: 
    ]
```

### DIFF
--- a/src/main/scala/ir/transforms/AbsInt.scala
+++ b/src/main/scala/ir/transforms/AbsInt.scala
@@ -15,25 +15,30 @@ import scala.util.{Failure, Success}
 import ExecutionContext.Implicits.global
 
 
-trait AbstractDomain[L] {
-  def join(a: L, b: L, pos: Block): L
-  def widen(a: L, b: L, pos: Block): L = join(a, b, pos) /* not used */
+
+trait GenericAbstractDomain[Location, Code, L] {
+  def join(a: L, b: L, pos: Location): L
+  def widen(a: L, b: L, pos: Location): L = join(a, b, pos) /* not used */
   def narrow(a: L, b: L): L = a
-  def transfer(a: L, b: Command): L
-  def init(b: Block): L = bot
+  def transfer(a: L, b: Code): L
+  def init(b: Location): L = bot
 
-  def isFixed(prev: L, next: L) : Boolean = (prev == next)
+  def isFixed(prev: L, next: L): Boolean = (prev == next)
 
+  def top: L
+  def bot: L
+}
+
+
+trait AbstractDomain[L] extends GenericAbstractDomain[Block, Command, L] {
   def transferBlockFwd(a: L, b: Block): L = {
     transfer(b.statements.foldLeft(a)(transfer), b.jump)
   }
   def transferBlockBwd(a: L, b: Block): L = {
     b.statements.toList.reverse.foldLeft(transfer(a, b.jump))(transfer)
   }
-
-  def top: L
-  def bot: L
 }
+
 
 trait PowerSetDomain[T] extends AbstractDomain[Set[T]] {
   def bot = Set()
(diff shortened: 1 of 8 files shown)


src/main/scala/ir/transforms/AbsInt.scala
```diff
@@ -15,95 +15,101 @@ import scala.util.{Failure, Success}
...
+    def transfer(a: L, b: Command): L = b match {
+      case call: DirectCall => {
+        transferCallSummary(a, procSummary(call.target), call)
       }
-      state = domain.transfer(state, b.jump)
+      case o => d.transfer(a, b)
     }
+}
 
+trait ProcedureSummaryGenerator[L, LocalDomain] extends GenericAbstractDomain[Procedure, Procedure, L] {
```
sadrabt, line 88: should this extend `ProcAbstractDomain[L]` instead?

src/main/scala/ir/transforms/AbsInt.scala
```diff
@@ -15,95 +15,101 @@ import scala.util.{Failure, Success}
...
+  def localTransferCall(l: LocalDomain, summaryForTarget: L, p: DirectCall) : LocalDomain
+  def updateSummary(prevSummary: L, p: Procedure,  resBefore: Map[Block, LocalDomain], resAfter: Map[Block, LocalDomain]) : L
```
sadrabt, line 90: from my understanding these are two transfer functions with simplified signature 
(LocalDomain, SummaryDomain) -> LocalDomain
(LocalDomain, SummaryDomain) -> SummaryDomain

is the updating of summaries necessary because of recursion? would we otherwise be able to do an ordered computation of the summaries using the call graph so that we wouldn't need to update the summaries in the 
`interprocSummaryFixpointSolver`

sadrabt:
Otherwise looks good and doesn't break anything for me 🏖️

src/main/scala/ir/transforms/AbsInt.scala
ailrst: Yep

src/main/scala/ir/transforms/AbsInt.scala
ailrst: The updateSummary method also serves the purpose of going from LocalDomain -> SummaryDomain the signature just takes the whole local analysis results and calculates a summary (although it probably really only needs the entry/exit block labels).  Its signature (as a transfer function) is more like (SummaryDomain, LocalDomain*) -> SummaryDomain.

But yeah you are right if we went in reverse topological order we would need to update the summaries less.

ailrst:
Thanks for your review

Issue #305, opened 2025-01-24
IntraprocSSASolver diverges on loops


ailrst:
When encountering loops this analysis is just going to keep increasing the SSA indexes on every iteration and never reach a fixed point. 

I suspect this analysis really just wants to compute the reaching definitions (as its not doing any transform).

PR #306, opened 2025-01-24
make ssa lazy dependent on mra or vsa

ailrst:
Workaround for https://github.com/UQ-PAC/BASIL/issues/305 that disables SSA,mmm,steensgaard if not needed. SSA is required now only for the flags --resolve-indirect-calls (which uses vsa) and --memory-regions mra. 

In the long term we will instead probably want DSA-based indirect call resolution in a separate analysis pipeline to the generic --analyse.


ailrst:
Replaced by https://github.com/UQ-PAC/BASIL/pull/307

PR #307, opened 2025-01-28
Address #305 with new analysis producing a mapping (position, variable) -> integer

ailrst:
Fixes https://github.com/UQ-PAC/BASIL/issues/305. 

This attempts to perform the same analysis as the "IntraSSA" with one pass. Since it's not really possible to produce a ssa-like renaming without introducing phi-nodes or copies, I'm not sure what the 'ssa' invariant the rest of the analysis pipeline is expecting, but this should at least give some discrimination for straight-line lists of assignments without diverging on loops.

It works by 

1. Give every syntactic variable-definition its own fresh index
2. Use union find to solve for the sets of indexes which, for any pair there exists a use with both definitions reachable.


l-kent:
This seems to work (and actually improves the MRA's results for whatever reason). Should we just deprecate the existing SSASolver too?

src/main/scala/analysis/SSA.scala
```diff
@@ -93,3 +94,126 @@ class IntraprocSSASolver(program: Program, writesTo: Map[Procedure, Set[Register
...
+ *
+ * Pass 1: Map each definition to a number in reverse post order
+ * Pass 2: Use reaching definitions results to unify definition indexes under their set of uses
+ *
+ */
+def getCommonDefinitionVariableRenaming(p: Program, writesTo: Map[Procedure, Set[Register]]) : Map[CFGPosition, (Map[Variable, FlatEl[Int]], Map[Variable, FlatEl[Int]])] = {
+
+  val mappings = p.preOrderIterator.collect {
+    case b: Assign => b -> nextSSACount()
+  }.toMap
```
l-kent, line 124: Is this used?

ailrst: Ah no I meant to delete it

l-kent:
I'm happy to merge this in alongside removing the existing SSASolver if you agree it makes sense to do so?

ailrst:
I'll remove the old SSA, if we need to refer to it it will be in the history.

l-kent:
Do we want to put it inside a class or something since just having this single method is a pretty different structure to the rest of the codebase?

ailrst:
No, I really don't see the point, this has no state and is already namespaced by the package.

l-kent:
On closer examination the 'improvement' in the MRA result seems to be noise created by non-determinism somewhere but I'm not sure if it's from this or if it's in the MRA itself.

l-kent:
I think it's probably from this though because I didn't notice this behaviour before.

l-kent:
Ok it's definitely not this, I'll merge.

PR #309, opened 2025-02-03
BAPLoader produces unique block names

### LINKED ISSUE #238: Ensure blocks have unique names
BAP can sometimes produce multiple blocks in a procedure that have the same name, such as in tests/correct/syscall/gcc_O2. This can cause problems as we do not make sure the blocks have unique names but we rely on the blocks having unique names.

l-kent:
Uses BAP TIDs when parsing block names from BAP to ensure that BASIL blocks are given unique names

Fixes #238

### DIFF
--- a/src/main/scala/translating/BAPLoader.scala
+++ b/src/main/scala/translating/BAPLoader.scala
@@ -1,13 +1,16 @@
 package translating
 
-import Parsers.BAP_ADTParser._
-import bap._
+import Parsers.BAP_ADTParser.*
+import bap.*
 import ir.Endian
 
 import scala.annotation.tailrec
-import scala.jdk.CollectionConverters._
+import scala.collection.mutable
+import scala.jdk.CollectionConverters.*
 
-object BAPLoader {
+class BAPLoader {
+
+  private val tidToLabel = mutable.Map[String, String]()
 
   def visitProject(ctx: ProjectContext): BAPProgram = {
     val memorySections = visitSections(ctx.sections)
@@ -31,7 +34,7 @@ object BAPLoader {
   def visitProgram(ctx: ProgramContext): List[BAPSubroutine] = ctx.subs.sub.asScala.map(visitSub).toList
 
   @tailrec
-  def visitExp(ctx: ExpContext): BAPExpr = ctx match {
+  final def visitExp(ctx: ExpContext): BAPExpr = ctx match {
     case e: ExpParenContext  => visitExp(e.exp)
     case e: LoadContext      => visitLoad(e)
     case e: BinOpContext     => visitBinOp(e)
@@ -109,8 +112,8 @@ object BAPLoader {
 
   def visitIndirectCall(ctx: IndirectCallContext): BAPIndirectCall = {
     val returnTarget = Option(ctx.returnTarget) match {
-      case Some(r: DirectContext) => Some(parseLabel(r.tid.name))
-      case None                   => None
+      case Some(r: DirectContext) => Some(visitTid(r.tid))
+      case None => None
     }
     val line = visitQuoteString(ctx.tid.name)
     val insn = parseFromAttrs(ctx.attrs, "insn").getOrElse("")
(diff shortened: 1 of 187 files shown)


src/main/scala/translating/BAPLoader.scala
```diff
@@ -241,12 +244,32 @@ object BAPLoader {
...
+    if (tidToLabel.contains(tid)) {
+      tidToLabel(tid)
+    } else {
+      // check for conflicts
+      val label = parseLabel(ctx.name)
+      var newLabel = label
+      var i = 1
+      val labels = tidToLabel.values.toSet
+      while (labels.contains(newLabel)) {
+        newLabel = s"$label$$$i"
```
ailrst, line 258: Using `$` as a separator is a pain as editors typically don't recognise it as part of a identifier token making it a pain to select & search for them. Use something else as a separator.

src/main/scala/translating/BAPLoader.scala
```diff
@@ -241,12 +244,32 @@ object BAPLoader {
...
+  }
+
   def visitQuoteString(ctx: QuoteStringContext): String = ctx.getText.stripPrefix("\"").stripSuffix("\"")
 
   def parseAllowed(s: String): String = s.map(c => if allowedChars.contains(c) then c else '$')
 
-  def parseLabel(ctx: QuoteStringContext): String =
-    "l" + parseAllowed(visitQuoteString(ctx).stripPrefix("@").stripPrefix("%"))
+  def parseLabel(ctx: QuoteStringContext): String = {
+    "$" + parseAllowed(visitQuoteString(ctx).stripPrefix("@").stripPrefix("%"))
```
ailrst, line 271: I don't see a reason for this change and it seems to be the only cause for this pr updating every expected file?

ailrst (changes requested):
Is there an example that reproduces the issue this is addressing?

src/main/scala/translating/BAPLoader.scala
l-kent: What should be used? $ is very convenient to use because it cannot be used in C identifiers but can be used in Boogie identifiers. We are already using it extensively in the GTIRB parser for that reason.

src/main/scala/translating/BAPLoader.scala
l-kent: Using a character like 'l' is more likely to lead to conflicts in edge cases

l-kent:
correct/syscall/gcc_O2 still exists and still has this issue in main, it just is not reflected in the Boogie output as the blocks are removed due to the procedure being a stub.

src/main/scala/translating/BAPLoader.scala
ailrst: Including the TID should address that though?

l-kent: Conflicts with non-block identifiers, not other block identifiers

src/main/scala/translating/BAPLoader.scala
ailrst: I would prefer an underscore. `$` makes even less sense in the gtirb parser since its emitting arbitrary UUIDs which have no relation to c identifiers.

l-kent: Underscores do not work for this purpose as they can be used in C identifiers

ailrst: You're worried about a C function name conflicting with a block `blockname_${TID}`?

src/main/scala/translating/BAPLoader.scala
ailrst: Such as procedure identifiers? I don't think we have boogie functions beginning with l and that is easy to avoid as we have control there. The solution I have to this in simplification-pass is to always include procedure addresses in their identifier names which should make them disjoint from the block ids bap is realistically going to emit. I don't think its really worth worrying about.

src/main/scala/translating/BAPLoader.scala
l-kent: The TID is not used in the block name (just to distinguish between distinct blocks that BAP has given the same name) but I would like to avoid any chance of that sort of thing happening. It is an edge case, yes.

ailrst:
Included in https://github.com/UQ-PAC/BASIL/pull/435

PR #310, opened 2025-02-04
Lattice sets and maps

b-paul:
Includes a special set and map representing terms in an abstract powerset lattice or map lattice.


ailrst (approved):
This looks fine

PR #311, opened 2025-02-06
fixes from cntlm with offline lifter

ailrst:
Some fixes to get cntlm lifted with the offline lifter to go through the gtirb frontend

- relax the expectations of the gtirb loader a bit so it doesn't immediately crash on bad rASL and detached blocks
- more assertions in simplifier as I keep adding them when debugging
- fix to repeat bits in cil visitor


sadrabt (approved):
looks good

PR #319, opened 2025-02-12
Simplification pass autofmt

ailrst:
The simplification pass merge seems like a good opportunity to enable autoformatting. 

This updates the scalafmt version to a working one (previously it was crashing on most files in the project), and adds an action which checks whether new code is formatted, and formats the project.

We also set `align.preset = none` ([scalafmt docs](https://scalameta.org/scalafmt/docs/configuration.html#alignpreset)) in .scalafmt.conf to avoid aligning `=>` in matches and reduce diff noise.


utting:
Merged scalafmt pass to reduce spurious code changes and improve readability (on average).

PR #321, opened 2025-02-12
ToScala for converting Basil IR to DSL strings

katrinafyi:
Implements a function for converting arbitrary Basil IR into stringified Scala source code, making use of the DSL expressions. This should help with creating new test cases based on examples found from binaries.

## Usage

This is provided as a .toScala extension method on statements, procedures, and programs. An example can be found in the [ToScalaTest](https://github.com/UQ-PAC/BASIL/blob/4523d578990a330778de3b0f70907c4c0a7c8aaa/src/test/scala/ir/ToScalaTest.scala) file.

The --dump-il has modified to dump the Scala as well, for example
```bash
./mill run --load-directory-bap src/test/correct/function1/gcc --dump-il function
```
will produce a function1-output.scala file that looks something like this:
```python
prog(
  proc("get_two",
    block("lget_two",
      LocalAssign(LocalVar("R30_begin", BitVecType(64), 0), Register("R30", 64), None),
      LocalAssign(Register("R31", 64), BinaryExpr(BVADD, Register("R31", 64), BitVecLiteral(BigInt("18446744073709551600"), 64)), Some("%00000346")),
      MemoryStore(StackMemory("stack", 64, 8), BinaryExpr(BVADD, Register("R31", 64), BitVecLiteral(BigInt("15"), 64)), Extract(8, 0, Register("R0", 64)), Endian.LittleEndian, 8, Some("%0000034e")),
      MemoryStore(StackMemory("stack", 64, 8), BinaryExpr(BVADD, Register("R31", 64), BitVecLiteral(BigInt("8"), 64)), Extract(32, 0, Register("R1", 64)), Endian.LittleEndian, 32, Some("%00000356")),
      MemoryStore(StackMemory("stack", 64, 8), Register("R31", 64), Register("R2", 64), Endian.LittleEndian, 64, Some("%0000035e")),
      MemoryLoad(LocalVar("load18", BitVecType(8), 0), StackMemory("stack", 64, 8), BinaryExpr(BVADD, Register("R31", 64), BitVecLiteral(BigInt("15"), 64)), Endian.LittleEndian, 8, Some("%00000365$0")),
      LocalAssign(Register("R1", 64), ZeroExtend(56, LocalVar("load18", BitVecType(8), 0)), Some("%00000365$1")),
...
```

Please also have a read of the doc comments in the various files:
https://github.com/UQ-PAC/BASIL/blob/2ffab7f889e3dbed3f139a73fcd74eae37a33817/src/main/scala/ir/dsl/ToScala.scala#L9-L37

## Implementation

This implements generation of the ToScala instances by using compile-time introspection. This is nice as it runs all within the Scala compiler, avoiding the need to go via astexplorer.net and Python. The generated code produces identical values to the previous Python-generated Scala code.

See code and doc comments in the Scala file. In the same file, there is also a short tutorial that covers the necessary background material.

https://github.com/UQ-PAC/BASIL/blob/31e35de7bb71136520db38eecfa49a46ad77ebbb/src/main/scala/ir/dsl/ToScalaDeriving.scala#L6-L28

------------

- [x] twine + indent
- [x] assertions for changed constructors
- [x] ~~port python to scala using scalameta directly~~
- [x] using params in ToScala[...]
- [x] use less given and more normal functions
- [x] split blocks
- [ ] perf profiling
- [x] name sanitisation
- [x] perf tweaks to indent method
- [x] **support procedure call and return parameters**
- [x] **think about macro deriving**


src/main/scala/ir/dsl/IRToDSL.scala
```diff
@@ -0,0 +1,267 @@
...
+
+// end generated
+
+
+given ToScala[Return] with
+  def toScala(x: Return): String = "ret"
+given ToScala[DirectCall] with
+  def toScala(x: DirectCall): String =
+    val str = summon[ToScala[String]]
+    s"directCall(${str.toScala(x.target.procName)}\")"
```
katrinafyi, line 167:
```suggestion
    s"directCall(${str.toScala(x.target.procName)})"
```

katrinafyi:
This is ready for review now. The major new feature is splitting, which extracts intermediate expressions in the output into separate functions. This helps avoid "method too large" errors in the Scala compiler. See:
https://github.com/UQ-PAC/BASIL/blob/4523d578990a330778de3b0f70907c4c0a7c8aaa/src/main/scala/ir/dsl/ToScala.scala#L102-L133

For fun, we tried this with cntlm-noduk. Using splitting as well as the `split_toscala_output.py` script, this produces Scala files which can compile in 2 minutes. The DSL expressions then take 15 seconds to evaluate, and running toScala on the loaded DSL takes another 800 ms.

katrinafyi:
Good news, we have excised the python by using mirror-based deriving. This is ready for review now. No more big changes, I promise :)

ailrst:
Trying to get `ir.MemorySection` in `Program.scala` to implement `ToScala` its not currently working because MergedRegionMRA has a circular dependency through a through Set[MergedRegionMRA]. (MergedRegions also happen to be a classes but we can easily make them case classes in a sealed trait). I would be fine with throwing away the merged region field in MemorySection as its just an analysis result we can recompute but that requires manually writing the serialiser for merged region rather than using the macro.

ailrst:
I'm curious if the typeclass approach to serialising is just a neccessity of the macros in this case or you think its a good way to do serialisers in general. It lets you define everything in parallel but if I was implementing it I would just do a big match expression in a function. The twines and indenting generalization are good to have though, we would refactor the other printers to use them if we had time.

katrinafyi:
I believe _not_, I think you can initiate the macros in any place.

But the trouble with the big match expression or basic function overloading is it needs to handle a range of non-hierarchical and generic types like scalar types and collections. It's hard to handle these cases because you can't add the constraint that you need ToScala[T] to get ToScala[List[T]], for example. 

I think this is most conveniently expressed by type classes. Type classes also give us extension methods which I think are kinda nice to use (whether extension methods are good is a separate discussion).

ailrst (approved):

PR #326, opened 2025-02-17
Return Some[] from irrefutable unapply methods

katrinafyi:
this changes the return type of irrefutable unapply methods to be Some[...] instead of Option[...], aiding the compiler's deduction of pattern-match exhaustivity.

this avoids confusing warnings where there is a case

```scala
case LocalVar(_, _, _) => ...
```

but the compiler still says

    it would fail on pattern case: _: ir.LocalVar

\- presumably accounting for the situation where the unapply method returns None.

with this change, we are also able to remove one "unreachable code" assertion in GTIRBToIR.scala, as the compiler correctly deduces it is unreachable.

in fact, if you remove that explicit unreachable case without applying this PR, the scala compiler suggests this change with -explain:
```scala
[warn] -- [E029] Pattern Match Exhaustivity Warning: /home/progs/basil/src/main/scala/translating/GTIRBToIR.scala:377:34 
[warn] 377 |      val jumpCopy = currentBlock.jump match {
[warn]     |                     ^^^^^^^^^^^^^^^^^
[warn]     |match may not be exhaustive.
[warn]     |
[warn]     |It would fail on pattern case: _: ir.Unreachable, _: ir.Return, _: ir.GoTo
[warn]     |---------------------------------------------------------------------------
[warn]     | Explanation (enabled by `-explain`)
[warn]     |- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
[warn]     | There are several ways to make the match exhaustive:
[warn]     |  - Add missing cases as shown in the warning
[warn]     |  - If an extractor always return Some(...), write Some[X] for its return type
[warn]     |  - Add a case _ => ... at the end to match all remaining cases
[warn]      ---------------------------------------------------------------------------

```

### DIFF
--- a/src/main/scala/ir/Expr.scala
+++ b/src/main/scala/ir/Expr.scala
@@ -375,7 +375,7 @@ case class LocalVar(varName: String, override val irType: IRType, val index: Int
 }
 
 object LocalVar {
-  def unapply(l: LocalVar): Option[(String, IRType, Int)] = Some((l.name, l.irType, l.index))
+  def unapply(l: LocalVar): Some[(String, IRType, Int)] = Some((l.name, l.irType, l.index))
 
 }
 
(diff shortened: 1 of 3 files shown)


ailrst (approved):
This is huge, thanks for fixing this.

PR #332, opened 2025-02-20
cloning: preserve labels and addresses

ailrst:
still to do: 

- [x] preserve localvar indices
    - its only not supporting indices on formal parameters, which should never have an ssa index anyway
- [ ] can we update irtodsl to serialise MemorySection using the macro?

### DIFF
--- a/src/main/scala/ir/dsl/IRToDSL.scala
+++ b/src/main/scala/ir/dsl/IRToDSL.scala
@@ -1,5 +1,6 @@
 package ir.dsl
 
+import scala.collection.immutable.SortedMap
 import ir.*
 
 /**
@@ -18,9 +19,9 @@ object IRToDSL {
   def localVarToTuple(x: LocalVar) = (x.name, x.irType)
 
   def convertJump(x: Jump): EventuallyJump = x match {
-    case Unreachable(label) => unreachable
-    case Return(label, out) => ret(out.map(keyToString).toArray: _*)
-    case GoTo(targs, label) => goto(targs.map(_.label).toArray: _*)
+    case Unreachable(label) => EventuallyUnreachable(label)
+    case Return(label, out) => EventuallyReturn(out.map(keyToString).toArray, label)
+    case GoTo(targs, label) => EventuallyGoto(targs.map(t => DelayNameResolve(t.label)).toList, label)
   }
 
   def cloneStatement(x: NonCallStatement): NonCallStatement = x match {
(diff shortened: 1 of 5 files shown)


src/main/scala/ir/dsl/IRToDSL.scala
```diff
@@ -18,9 +19,9 @@ object IRToDSL {
   def localVarToTuple(x: LocalVar) = (x.name, x.irType)
 
   def convertJump(x: Jump): EventuallyJump = x match {
-    case Unreachable(label) => unreachable
-    case Return(label, out) => ret(out.map(keyToString).toArray: _*)
-    case GoTo(targs, label) => goto(targs.map(_.label).toArray: _*)
+    case Unreachable(label) => EventuallyUnreachable(label)
+    case Return(label, out) => EventuallyReturn(out.map(keyToString).toArray, label)
+    case GoTo(targs, label) => EventuallyGoto(targs.map(t => DelayNameResolve(t.label)).toList, label)
```
katrinafyi, line 24: I chose to use the DSL functions because I wanted to have a single point to construct EventuallyThings. This would help maintain invariants such as sortedness. I also had a thought to (eventually, maybe) make the Eventually classes private.

src/main/scala/translating/PrettyPrinter.scala
```diff
@@ -288,6 +289,7 @@ class BasilIRPrettyPrinter(
   }
 
   override def vproc(p: Procedure): PPProg[Procedure] = {
+    p.sortBlocksRPO()
```
katrinafyi, line 292: I don't really like mutating the IR within the pretty-printer. Maybe this can be maintained as an invariant, but this could be difficult because lots of things can affect the RPO.

src/test/scala/ir/IRTest.scala
```diff
@@ -366,4 +366,9 @@ class IRTest extends AnyFunSuite {
     assert(prog(recursiveproc) != null)
     assert(recursiveproc.addToProg(emptyprog) != null)
   }
+
+  test("sort test")  {
+
+  }
```
katrinafyi, line 372: fixme

katrinafyi (changes requested):
thanks for the help! i think the IR changes with RPO should be a separate PR

src/main/scala/ir/dsl/IRToDSL.scala
ailrst: Default params and variadic functions don't play nice and I was trying to make the functions stay nice (but incomplete program representation) for manually writing them. It would be fine if we overloaded them though so we have a complete variant of them

src/test/scala/ir/IRToDSLTest.scala
```diff
@@ -104,4 +115,86 @@ class IRToDSLTest extends AnyFunSuite {
...
+    val cloned = IRToDSL.convertProgram(prog).resolve
+
+    for (orig <- prog) {
+      for (clone <- cloned) {
+        assert(orig ne cloned, "No references to control-flow objects shared between IR")
+      }
+    }
+
+    prog.sortProceduresRPO()
+    cloned.sortProceduresRPO()
```
ailrst, line 142: This test is still sorting the callgraph so the order of procedures agrees, changing procedure order is probably a bug in the cloning.

src/main/scala/ir/dsl/IRToDSL.scala
katrinafyi: Yeah I'm happy leaving this as-is for now.

PR #337, opened 2025-02-26
Atomic instructions support

l-kent:
Support for atomic instructions such as `casal` which were added in ARMv8.1a. Rely and guarantee predicates are not applied or checked inside an atomic section, only outside the atomic section.

sync_read and sync_write examples have been added that demonstrate this new functionality.

Work still needs to be done to merge this into main, as the relevant sections of code have changed dramatically.


ailrst:
Hopefully,

```bash
$ git cherry-pick f8d7e80edba52478d28e86b0c86229d8113592ad 
$ git cherry-pick 7740b7b03f09f9773555c87a5d63f396e67b12e7 
$ # changes to scalafmt config ^
$ scalafmt src
$ git add -u && git commit -m 'format'
$ # merge main into head
```

Should prevent a substantial amount of the merge conflicts. 
The code you've touched in this PR shouldn't have too many changes
in main otherwise.

ailrst:
I've merged main in in this branch if that helps; https://github.com/UQ-PAC/BASIL/compare/atomics-merge?expand=1

src/main/scala/ir/Statement.scala
```diff
@@ -126,13 +135,13 @@ class Assume(
   override def acceptVisit(visitor: Visitor): Statement = visitor.visitAssume(this)
 }
 
-object Assume:
-  def unapply(a: Assume): Some[(Expr, Option[String], Option[String], Boolean)] =
+object Assume {
+  def unapply(a: Assume): Option[(Expr, Option[String], Option[String], Boolean)] =
```
ailrst, line 139: This was probably reverted in the merge incorrectly

src/main/scala/ir/transforms/Simp.scala
```diff
@@ -489,8 +490,9 @@ def coalesceBlocks(p: Program): Boolean = {
       } else if (
         b.nextBlocks.size == 1 && b.nextBlocks.head.statements.nonEmpty && b.statements.nonEmpty
         && b.nextBlocks.head.prevBlocks.size == 1
-        && b.statements.lastOption.map(s => !(s.isInstanceOf[Call])).getOrElse(true)
+        && b.statements.lastOption.forall(s => !s.isInstanceOf[Call])
         && !(b.parent.entryBlock.contains(b) || b.parent.returnBlock.contains(b))
+        && b.nextBlocks.forall(_.atomicSection.isEmpty)
```
ailrst, line 495: This might move the content of this block `b` which ends an atomic section into a successor outside the atomic section. 

something like : `b.atomicSection.isEmpty && b.nexBlocks.forall(_.atomicSection.isEmpty) || b.atomicSection.forall(atomic => b.nextBlocks.forall(atomic.blocks.contains))`


The previous case line 481 could also needs to check its not across an atomic boundary rather than that `b` isnt in an atomic section.

src/main/scala/ir/Program.scala
```diff
@@ -483,6 +483,7 @@ class Block private (
   private var _jump: Jump,
   private val _incomingJumps: mutable.HashSet[GoTo]
 ) extends HasParent[Procedure] {
+  var atomicSection: Option[AtomicSection] = None
```
ailrst, line 486: Should we also remove the block from the atomic section when it is removed from the procedure (possibly on deParent?) I wouldn't expect not doing this to cause issues as we only ever really care about 'is this block in an atomic section'.

src/main/scala/translating/GTIRBToIR.scala
```diff
@@ -321,19 +320,24 @@ class GTIRBToIR(
   private def convertLabel(procedure: Procedure, label: ByteString, blockCount: Int): String = {
     procedure.name + "__" + blockCount + "__" + byteStringToString(label)
       .replace("=", "")
-      .replace("-", "__")
-      .replace("/", "__")
+      .replace("-", "~")
+      .replace("/", "\'")
```
ailrst, line 324: Reversion from merge

ailrst (commented):
Can you please describe the structure the loader is establishing for the atomic sections somewhere (possibly with an invariant check for the well-formedness of the atomic sections). My understanding is that the atomic start and atomic end just have to fully enclose some blocks, and the blocks each get a reference to the section they are in. (and that the atomic start/end statements aren't supposed to be semantically meaningful; just comments?)

What is the motivation for implementing the block re-arranging twice; once in each frontend; rather than as an early transform of the IR? Both frontends seem to initially emit a very similar structure that just bounds the atomic section with a special nop statements.

src/main/scala/translating/GTIRBToIR.scala
l-kent: Maintaining uniqueness with these characters is important for practical reasons so that they can be distinguished in the Boogie output. Replacing both with the same character is what I had originally tried but it caused problems.

l-kent: An underscore is already a possible character

src/main/scala/ir/transforms/Simp.scala
l-kent: I've just set it so that if the block on either side is atomic then we don't combine blocks.

ailrst: That's fine

src/main/scala/ir/Program.scala
l-kent: I think I would prefer to leave this to whatever is manipulating the blocks, to make it easier to account for any bugs that may arise. I have added a check for malformed atomic sections.

l-kent:
> What is the motivation for implementing the block re-arranging twice; once in each frontend; rather than as an early transform of the IR? Both frontends seem to initially emit a very similar structure that just bounds the atomic section with a special nop statements.

The structures produced by the two lifters are different enough that this is not possible in any reasonable way. It makes sense to account for these lifter-specific variations in the part of the code that is specific to each lifter.

l-kent:
> My understanding is that the atomic start and atomic end just have to fully enclose some blocks, and the blocks each get a reference to the section they are in. 

Yes - we just care about whether a block is inside an atomic section, and where the start and end of an atomic section is. 

> (and that the atomic start/end statements aren't supposed to be semantically meaningful; just comments?)

They are semantically meaningful for the purpose of inserting rely & guarantee conditions into the Boogie output. Because atomic  sections cannot be interfered with by other threads, we have to place all rely/guarantee conditions on the outside of an atomic section, with none allowed inside. I will make this clear in documentation.

ailrst:
> They are semantically meaningful for the purpose of inserting rely & guarantee conditions into the Boogie output. Because atomic sections cannot be interfered with by other threads, we have to place all rely/guarantee conditions on the outside of an atomic section, with none allowed inside. I will make this clear in documentation.

By not semantically meaningful I was more referring to that we use the `Block.atomicSection` to know this rather than the atomicstart/atomic end statement.

ailrst (approved):

PR #338, opened 2025-02-26
output initial memory in program.toScala

ailrst:
This implements `InitialMemory.toScala` and makes the generated program include it. An interface like `ToScala.deriveWithExclusions` that allows eliminating record fields rather than sum alternatives would help here, but I ended up making the output more specialised to reduce its verbosity so that part of the implementation is really the least complex.


src/main/scala/ir/dsl/ToScalaDeriving.scala
```diff
@@ -23,7 +23,7 @@ import scala.compiletime.{summonInline, erasedValue, constValue, error}
  *       case A
  *       case B
  *     }
- *     given ToScala[EAAA] =
+ *     given ToScala[EAAA] with
```
katrinafyi, line 26: my apologies. we should add a mention that this is the old given syntax

src/main/scala/ir/dsl/ToScala.scala
```diff
@@ -85,12 +85,17 @@ given [T](using ToScala[T]): ToScalaLines[Seq[T]] with
...
+  extension (x: Some[T])
     def toScala: String = x match
-      case None => "None"
       case Some(x) => s"Some(${x.toScala})"
 
+given ToScalaString[None.type] with
+  extension (x: None.type)
+    def toScala: String = "None"
+
+given [T](using ToScala[T]) : ToScala[Option[T]] = ToScala.derived
```
katrinafyi, line 97: does None.toScala not work without this change? ig it is relying on several dimensions of variance

src/main/scala/ir/dsl/ProgramToScala.scala
```diff
@@ -46,10 +46,37 @@ def procedureToScalaWith(blockToScala: Block => Twine)(x: Procedure): Twine =
...
+          )
+        ),
+        ")"
+      )
+
 def programToScalaWith(procedureToScala: Procedure => Twine)(x: Program): Twine =
   val main = x.mainProcedure
   val others = x.procedures.to(LazyList).filter(_ ne main)
-  indentNested("prog(", (main #:: others).map(procedureToScala), ")")
+  val mem = indentNested("Seq(",  x.initialMemory.values.map(_.toScalaLines), ")")
```
katrinafyi, line 78: this can reuse ToScala[Seq] or ToScala[Iterable]. I can take a look tmrw

katrinafyi (commented):
thanks! some questions and some touch-ups which I can do tomorrow

src/main/scala/ir/dsl/ToScala.scala
ailrst: It doesn't work because it tries to summon a `Some[T]` and can't choose a `T` because its unconstrained. 

```
[error] -- [E008] Not Found Error: /home/am/Documents/programming/2024/bil-to-boogie-translator/src/main/scala/ir/dsl/ProgramToScala.scala:52:15 
[error] 52 |  val x = None.toScala
[error]    |          ^^^^^^^^^^^^
[error]    |value toScala is not a member of object None.
[error]    |An extension method was tried, but could not be fully constructed:
[error]    |
[error]    |    ir.dsl.given_ToScalaString_Option[T](
[error]    |      /* ambiguous: both object given_ToScalaLines_Block in package ir.dsl and object given_ToScalaLines_Procedure in package ir.dsl match type ir.dsl.ToScala[T] */
[error]    |        summon[ir.dsl.ToScala[T]]
[error]    |    ).toScala()
[error] one error found
```

katrinafyi: Ug yeah right.

ailrst:
I'd like to make it optionally ignore initial memory because it can be very large but that conflicts with toScala being a typeclass.

katrinafyi:
Ig you can add a custom extension method to programs like toScalaWithMemory and have the default omit the memory. Not beautiful but it works

katrinafyi:
@ailrst can you review the changes i tacked onto your pr? particularly the mixins and documentation

src/main/scala/ir/dsl/ProgramToScala.scala
```diff
@@ -101,19 +194,23 @@ given ToScalaLines[Program] with
  */
 object ToScalaWithSplitting {
 
+  // NOTE: a new instance must be created for every use of toScalaLines, since
+  // the class has mutable state.
+  private def instance = new ToScalaWithSplitting {}
+
```
ailrst, line 200: Why is this a member and not explicitly constructed in each call to toScalaLines

katrinafyi: DRY ig. changed now

PR #342, opened 2025-02-27
Use PrettyPrinter for prog/proc toString

katrinafyi:
Honestly, idk if this change is what we want because it makes the toString very large. this would be
troublesome when displaying procedures as elements of a collection, for example.

This reverts commit ff30fb7d3e45bb3f190ffa99a06d51e2acc0db29.

### DIFF
--- a/src/main/scala/ir/Program.scala
+++ b/src/main/scala/ir/Program.scala
@@ -118,7 +118,8 @@ class Program(
   }
 
   override def toString(): String = {
-    serialiseIL(this)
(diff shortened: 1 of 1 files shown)


ailrst:
I'd like to go through and make this change for all the basil IR structures.

katrinafyi:
Feel free to take control of this pr.

Do you have thoughts on the effect on lists of UR structures?

ailrst:
This change isn't a priority so we can close and address it later if you want

PR #343, opened 2025-02-27
Liftmake script

ailrst:
This is a small script to compile and lift examples by wrapping make. It is compatible with the reproducible build environment by pulling compiler and lifter binary paths from environment variables.

### DIFF
--- a/scripts/liftmake.sh
+++ b/scripts/liftmake.sh
@@ -1,0 +1,58 @@
+#!/bin/bash
+
+CFILE_NAME=""
+
+while getopts "b:c:" o; do
+case "${o}" in
+    b)
+        BIN_NAME=${OPTARG}
+        ;;
+    c)
+        CFILE_NAME="$CFILE_NAME ${OPTARG}"
+        ;;
+    j)
+        JOBS=${OPTARG}
+        ;;
+    *) 
+        if ${OPTARG} == "--" ; then
+        	break;
+        fi
+    esac
+done
+shift $((OPTIND-1))
+
+if [[ -z "$CFILE_NAME" ]] ; then 
+    echo "Usage: -c cfile.c [-b binaryname] [ -- make args]"
+    echo "Set variables CC, CFLAGS, "
+    exit 1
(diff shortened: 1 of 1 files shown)


scripts/liftmake.sh
```diff
@@ -0,0 +1,62 @@
...
+while getopts "b:c:" o; do
+case "${o}" in
+    b)
+        BIN_NAME=${OPTARG}
+        ;;
+    c)
+        CFILE_NAME="$CFILE_NAME ${OPTARG}"
+        ;;
+    j)
+        JOBS=${OPTARG}
```
katrinafyi, line 14: unused? or is this automatically picked up by make

katrinafyi (commented):
lgtm but does it need to be a script? a makefile + assertions on variables being set could be okay. but idm either way

scripts/liftmake.sh
```diff
@@ -0,0 +1,59 @@
...
+        if ${OPTARG} == "--" ; then
+        	break;
+        fi
+    esac
+done
+shift $((OPTIND-1))
+
+if [[ -z "$CFILE_NAME" ]] ; then 
+    echo "Usage: -c cfile.c [-b binaryname] [ -- make args]"
+    echo "Optinoally variables CC, CFLAGS, DDISASM, READELF, GTIRBSEM, BAP to change the the binary path to the lifter tools"
```
katrinafyi, line 23: typo

scripts/liftmake.sh
```diff
@@ -0,0 +1,59 @@
...
+    exit 1
+fi
+
+BIN_NAME=${BIN_NAME:=$(echo "$CFILE_NAME" | sed -s 's/.c$//' | tr -d ' ')}
+export BIN_NAME
+export CFILE_NAME
+
+make $@ -f - << 'EOF'
```
katrinafyi, line 31: the `$@` should be double quoted

katrinafyi (approved):
lgtm after fixes

PR #345, opened 2025-02-28
Procedure call inlining

ailrst:
This adds a function to perform procedure inlining that preserves DSA form. It adds a pass  under --simplify that inlines the plt launchpads from gtirb with a very basic heuristic. 

Its tested inasfar as `--validate-simplfy` succeeds on running cntlm, so it preserves DSA form. It needs a stronger test to make sure dataflow isn't being lost though. 

[example diff](https://www.jstoolset.com/diff/2a2167f417ddc158)

### DIFF
--- a/src/main/scala/ir/transforms/Inline.scala
+++ b/src/main/scala/ir/transforms/Inline.scala
@@ -1,0 +1,163 @@
+package ir.transforms
+import ir.*
+import util.Logger
+import ir.cilvisitor.*
+import ir.dsl.IRToDSL.*
+import ir.dsl.*
+
+import scala.collection.mutable
+
+object Counter {
+  var id = 0
+  def next() = {
+    id += 1
+    id
+  }
+}
+
+def memoise[K, V](f: K => V): K => V = {
+  val rn = mutable.Map[K, V]()
+
+  def fun(arg: K): V = {
+    if (!rn.contains(arg)) {
+      rn(arg) = f(arg)
+    }
+
+    rn(arg)
+  }
+  fun
+}
+
+def renameBlock(s: String): String = {
+  s + "_" + (Counter.next())
+}
+
+class VarRenamer(proc: Procedure) extends CILVisitor {
+
+  def doRename(v: Variable): Variable = v match {
+    case l: LocalVar if l.name.endsWith("_in") => {
+      val name = l.name.stripSuffix("_in")
+      proc.getFreshSSAVar(name, l.getType)
+    }
+    case l: LocalVar if l.index != 0 =>
+      proc.getFreshSSAVar(l.varName, l.getType)
+    case _ => v
+  }
+  val memoed = memoise(doRename)
+
+  def rename(v: Variable) = {
+    memoed(v)
+  }
+
+  override def vlvar(v: Variable) = {
+    ChangeTo(rename(v))
+  }
+
+  override def vrvar(v: Variable) = {
+    ChangeTo(rename(v))
+  }
+
+}
+
+def convertJumpRenaming(blockName: String => String, varName: CILVisitor, x: Jump): EventuallyJump = x match {
+  case GoTo(targs, label) => EventuallyGoto(targs.map(t => DelayNameResolve(blockName(t.label))).toList, label)
+  case Unreachable(label) => EventuallyUnreachable(label)
+  case Return(label, out) =>
+    EventuallyReturn(
+      out.toList.map { case (v: Variable, e: Expr) =>
+        ((v.name, visit_expr(varName, e)))
+      }.toArray,
+      label
+    )
+}
+
+def keyToString[T](varRenamer: CILVisitor)(x: (Variable, Expr)): (String, Expr) =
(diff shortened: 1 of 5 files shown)


src/main/scala/ir/transforms/Inline.scala
```diff
@@ -0,0 +1,162 @@
...
+    }
+    case l: LocalVar if l.index != 0 =>
+      proc.getFreshSSAVar(l.varName, l.getType)
+    case _ => v
+  }
+  val memoed = memoise(doRename)
+
+  def rename(v: Variable) = {
+    memoed(v)
+  }
```
katrinafyi, line 50: one of these variables can be omitted

src/main/scala/ir/transforms/Inline.scala
```diff
@@ -0,0 +1,162 @@
...
+  require(c.target.entryBlock.isDefined)
+  require(c.target.returnBlock.isDefined)
+
+  val proc = c.parent.parent
+  val block = c.parent
+  val target = c.target
+  // rename ssa variables to maintain DSA in the new procedure
+  // rename block labels to avoid creating conflicts if inlining the same function multiple times
+  val renamer = VarRenamer(proc)
+  val blockRenamer: String => String = memoise(renameBlock)
```
katrinafyi, line 116: renamer can be varRenamer, to better differentiate from blockRenamer

src/main/scala/ir/transforms/Inline.scala
```diff
@@ -0,0 +1,162 @@
...
+  block.replaceJump(GoTo(entryTempBlock))
+
+  // replace return in target with a jump to the aftercall block
+  val (returnTemp, resolveReturnBlock) = eventuallyReturnBlock.copy(j = unreachable).makeResolver
+  proc.addBlock(returnTemp)
+
+  // resolve internal call blocks
+  val resolvers = internalBlocks.map(_.makeResolver)
+  resolvers.foreach { case (block, _) => proc.addBlock(block) }
+  resolvers.foreach { case (_, resolve) => resolve(prog, proc) }
```
katrinafyi, line 138: i _think_ you might be able to use EventuallyBlock.resolve here instead of going through makeResolver
```suggestion
  internalBlocks.foreach(b => proc.addBlock(b.resolve(prog, proc)))
```

src/main/scala/ir/transforms/InlinePLTLaunchpad.scala
```diff
@@ -0,0 +1,23 @@
...
+
+def inlinePLTLaunchpad(prog: Program) = {
+  for (p <- prog.procedures) {
+
+    val candidate =
+      (p.blocks.size <= 4)
+        && p.calls.size == 1
+        && p.calls.forall(_.isExternal.contains(true))
+        && p.procName.startsWith("FUN")
+        && !p.calls.contains(p)
```
katrinafyi, line 12: does this needs more checks? e.g., that the only statement within the procedure is an external call?

src/main/scala/ir/transforms/Inline.scala
```diff
@@ -0,0 +1,162 @@
...
+
+}
+
+def convertJumpRenaming(blockName: String => String, varName: CILVisitor, x: Jump): EventuallyJump = x match {
+  case GoTo(targs, label) => EventuallyGoto(targs.map(t => DelayNameResolve(blockName(t.label))).toList, label)
+  case Unreachable(label) => EventuallyUnreachable(label)
+  case Return(label, out) =>
+    EventuallyReturn(
+      out.toList.map { case y: (Variable, Expr) =>
+        ((y._1.name, visit_expr(varName, y._2)))
```
katrinafyi, line 68: this can use a tuple pattern inside the case

katrinafyi (approved):
lgtm. seems standard and useful, tests would be nice to have. small nit picks

src/main/scala/ir/transforms/Inline.scala
ailrst: I doubt this will work if the procedure contains a loop the resolve on a goto will fail the target isn't already a member of the procedure its resolving against.

src/main/scala/ir/transforms/InlinePLTLaunchpad.scala
ailrst: Its deliberately coarse because there's extra garbage (e.g. setting R17) in the procedure

src/main/scala/ir/transforms/Inline.scala
ailrst: The closure needs to be explicitly constructed at class scope but i want the function to look like a normal method definition with a type signature so we do this.

src/main/scala/ir/transforms/Inline.scala
ailrst: Scala lsp auto rename doesn't work so I can't be bothered

ailrst:
It will get tested when the simplification pass diff tests get merged in a separate pr

PR #349, opened 2025-03-05
Port unit tests to munit

katrinafyi:
Munit has nicer documentation? And more obvious behaviour?


katrinafyi:
munit's output is a bit nicer, but it does not print any test case names until the entire suite is done. this is probably a big detractor.

image

katrinafyi:
i'm leaving this for now. while the diff output is nicer, its other quirks are frustrating enough to dissuade me. munit also provides more minimal features than scalatest. i think we can slog it out with scalatest for a bit longer, maybe with customising the Prettifier.

ailrst:
Maybe we'd be better off moving to make

PR #351, opened 2025-03-06
create more generic interpreter test for value analysis

ailrst:
This factors out the `InterpretConstProp` test into a more reasonable separate generic test for checking value analyses against the interpreter. 

You have to provide 

1. An analysis result `block -> variable -> absval`
2. A function `valueInAbstractVal: abstract:absval -> concrete:expr -> expr` which encodes the soundness check as a basil IR expression that evaluates to true when the `concrete` is an element of `abstract`. 

Its a bit iffy on choosing which variables to generate assertions for, if you test uninitialised variables it will crash the interpreter, but its fairly normal for uninitialised variables to appear in abstract domains.

### DIFF
--- a/src/test/scala/InterpretTestConstProp.scala
+++ b/src/test/scala/InterpretTestConstProp.scala
@@ -31,9 +31,17 @@ import util.RunUtils.loadAndTranslate
 
 import scala.collection.mutable
 
-class ConstPropInterpreterValidate extends AnyFunSuite {
+class ConstPropInterpreterValidate extends AnyFunSuite with TestValueDomainWithInterpreter[FlatElement[BitVecLiteral]] {
 
   Logger.setLevel(LogLevel.ERROR)
+
+  def valueInAbstractValue(absval: FlatElement[BitVecLiteral], concrete: Expr) = {
+    absval match {
+      case Top => TrueLiteral
+      case Bottom => TrueLiteral /* deliberately don't check */
+      case FlatEl(value) => BinaryExpr(BVEQ, value, concrete)
+    }
+  }
 
   def testInterpretConstProp(testName: String, examplePath: String) = {
     val loading = ILLoadingConfig(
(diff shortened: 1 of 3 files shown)


src/test/scala/InterpretTestConstProp.scala
```diff
@@ -31,10 +31,18 @@ import util.RunUtils.loadAndTranslate
...
 
-class ConstPropInterpreterValidate extends AnyFunSuite {
+class ConstPropInterpreterValidate extends AnyFunSuite with TestValueDomainWithInterpreter[FlatElement[BitVecLiteral]] {
 
   Logger.setLevel(LogLevel.ERROR)
 
+  def valueInAbstractValue(absval: FlatElement[BitVecLiteral], concrete: Expr) = {
+    absval match {
+      case Top => TrueLiteral
+      case Bottom => TrueLiteral /* deliberately don't check */
```
katrinafyi, line 41: sorry, why not check? it was FalseLiteral in the old ConstPropInterpreterValidate and that's what i'd expect

src/test/scala/util/TestValueAnalysisWithInterpreter.scala
```diff
@@ -0,0 +1,81 @@
...
+    // run the interpreter evaluating the analysis result at each command with a breakpoint
+    val interpretResult = interpretWithBreakPoints(ictx, breaks.toList, NormalInterpreter, InterpreterState())
+
+    assert(interpretResult._1.nextCmd == Stopped())
+    val breakres: List[(BreakPoint, _, List[(String, Expr, Expr)])] = interpretResult._2
+    breakres.flatMap { case (bp, _, l) =>
+      l.map { case (name, test, evaled) =>
+        CheckResult(name, bp, test, evaled)
+      }
+    }.toList
```
katrinafyi, line 65: idk the interpreter, so how does this work? it looks like the breakpoint results are handled all at once after interpreting, instead of the breakpoints interrupting during interpretation, is that right?

what is the inner list inside `breakres`? is it each time a particular breakpoint is encountered?

src/test/scala/util/TestValueAnalysisWithInterpreter.scala
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  def assertCorrectResult(breakres: List[CheckResult]) = {
+    assert(breakres.nonEmpty)
+    assert(breakres.nonEmpty)
```
katrinafyi, line 70:
```suggestion
    assert(breakres.nonEmpty)
```

katrinafyi (approved):
no major objections

src/test/scala/InterpretTestConstProp.scala
ailrst: Yeah the test isn't passing now and I don't want to debug the analysis because its not actually used :shrug:. Looking at the failing case it shouldn't have been mapped to bot.

src/test/scala/util/TestValueAnalysisWithInterpreter.scala
ailrst: > what is the inner list inside breakres? is it each time a particular breakpoint is encountered? 

Yes it should be

> idk the interpreter, so how does this work? it looks like the breakpoint results are handled all at once after interpreting, instead of the breakpoints interrupting during interpretation, is that right?

Yes its configurable in `BreakPointAction` as to whether it stops, where we just want to evaluate the tests and process the result after.

ailrst: The return type of the breakpoint is pretty horrible but I haven't got around to refactoring it

PR #354, opened 2025-03-10
DSL compiler for simple While language

ailrst:
This PR adds functions to the DSL so that code can be written in a simple while language and compiled to BASIL IR basic blocks, this adds:

`If (cond) Then (... blocks ...) Else (... blocks ...)`

`While (cond) Do (... blocks ...) `

`For (init, cond, incr) Do (... blocks ...)`

It also adds infix operators to build basil IR expressions. 

```scala
val p = prog(
  proc(
    "is_prime",
    Seq("n" -> bv64),
    Seq("ans" -> bv1),
    blocks(
      If(n <= 1.bv64)
        Then (ret("ans" -> (0.bv1)))
        Else (For(i := (2.bv64), i < n, i := i + (1.bv64))
          Do (If(n % i === (0.bv64)) Then (ret("ans" -> (0.bv1))))),
      ret("ans" -> (1.bv1))
    )
  )
)
```


katrinafyi:
Big changes! 

Really fast preliminary comments:
-  I think the new higher DSL should be in a separate file.
- also, it would be nice to organise the internal case classes separately from the user-facing functions as much as possible. for example, maybe they can be inside an object and marked private.
- are there guards against "invalid syntax". like putting two else clauses on a single if statement?

ailrst:
Reviewing this can wait until you're at work of course but, I agree with the above. 

> are there guards against "invalid syntax". like putting two else clauses on a single if statement?

Yes, though in many places the errors can be quite confusing if you don't know the implementation (esp. due to overloading, using implicit conversions seems to work better). For this case it should be fairly reasonable, `If` is a case class that only defines the method `Then` which returns a new case class which only defines the method `Else`, so you'll get a not found error if you write something else.

src/main/scala/ir/dsl/WhileDSL.scala
```diff
@@ -0,0 +1,199 @@
...
+/**
+  * Sequence a list of lists of blocks, where each is its own diamond, 
+  * so that each list element is executes in sequential order.
+  */
+def blocks(blocks: List[EventuallyBlock]*): List[EventuallyBlock] = {
+  blocks.toList match {
+    case Nil => List()
+    case h :: Nil => h
+    case h :: tail => tail.foldLeft(h)((acc, nextb) => sequence(acc, nextb))
+  }
```
katrinafyi, line 125: would fold be able to handle the nil and singleton cases?

src/main/scala/ir/dsl/InfixConstructors.scala
```diff
@@ -0,0 +1,180 @@
...
+def bv8 = BitVecType(8)
+def bv1 = BitVecType(1)
+
+extension (i: Expr)
+  infix def ===(j: Expr): Expr = i.getType match {
+    case IntType => BinaryExpr(IntEQ, i, j)
+    case b: BitVecType => BinaryExpr(BVEQ, i, j)
+    case BoolType => BinaryExpr(BoolEQ, i, j)
+    case m: MapType => ???
+  }
```
katrinafyi, line 60: since the DSL is meant to be user-facing, I think it would be helpful to have descriptive errors wherever possible. so a proper error message instead of ???.

src/main/scala/ir/dsl/InfixConstructors.scala
```diff
@@ -0,0 +1,180 @@
...
+  infix def !==(j: Expr): Expr = i.getType match {
+    case IntType => BinaryExpr(IntNEQ, i, j)
+    case b: BitVecType => BinaryExpr(BVNEQ, i, j)
+    case BoolType => BinaryExpr(BoolNEQ, i, j)
+    case m: MapType => ???
+  }
+  infix def +(j: Expr): Expr = i.getType match {
+    case IntType => BinaryExpr(IntADD, i, j)
+    case b: BitVecType => BinaryExpr(BVADD, i, j)
+    case BoolType => BinaryExpr(BoolOR, i, j)
```
katrinafyi, line 70: is `+` on booleans intentional? there is already `||`

src/main/scala/ir/dsl/InfixConstructors.scala
```diff
@@ -0,0 +1,180 @@
...
+    case b: BitVecType => BinaryExpr(BVAND, i, j)
+    case BoolType => BinaryExpr(BoolAND, i, j)
+    case m: MapType => ???
+  }
+  infix def ||(j: Expr): Expr = i.getType match {
+    case IntType => ???
+    case b: BitVecType => BinaryExpr(BVOR, i, j)
+    case BoolType => BinaryExpr(BoolOR, i, j)
+    case m: MapType => ???
+  }
```
katrinafyi, line 102: to match programming language convention, it would be nicer for the bitvec operations to use `|` and `&`.

build.sc
```diff
@@ -8,6 +8,7 @@ import contrib.scalapblib._
 
 object basil extends RootModule with ScalaModule with antlr.AntlrModule with ScalaPBModule {
   def scalaVersion = "3.3.4"
+  def ammoniteVersion = "3.0.2"
```
katrinafyi, line 11: is this used?

src/main/scala/ir/dsl/WhileDSL.scala
```diff
@@ -0,0 +1,199 @@
...
+  var count = 1
+  def next() = {
+    count += 1
+    count
+  }
+
+  def nlabel(name: String = "id"): String = {
+    name + "_" + next()
+  }
+}
```
katrinafyi, line 33: consider using thread-safe counter in utils. it's _probably_ not a big issue here, but it would be nice to prevent counter creep

src/main/scala/ir/dsl/WhileDSL.scala
```diff
@@ -0,0 +1,199 @@
...
+ * passed to [[sequence]].
+ */
+private def mkWhile(cond: Expr, body: List[EventuallyBlock]): List[EventuallyBlock] = {
+  val loopExit = Counter.nlabel("while_exit")
+  val loopBackedge = Counter.nlabel("while_backedge")
+  val loopEntry = Counter.nlabel("while_entry")
+  val loopBody = Counter.nlabel("while_body")
+  List(block(loopEntry, goto(loopBody, loopExit)))
+    ++
+      sequence(sequence(List(block(loopBody, Assume(cond))), body), List(block(loopBackedge, goto(loopEntry))))
```
katrinafyi, line 71: blah. use blocks()?

src/main/scala/ir/dsl/WhileDSL.scala
```diff
@@ -0,0 +1,199 @@
...
+ * the second list.
+ */
+def sequence(first: List[EventuallyBlock], rest: List[EventuallyBlock]): List[EventuallyBlock] = {
+  require(first.nonEmpty)
+  require(rest.nonEmpty)
+  require(first.last.j.isInstanceOf[EventuallyUnreachable])
+  val last = first.last
+  last.j = goto(rest.head.label)
+  first ++ rest
+}
```
katrinafyi, line 106: i don't think this is very diamond-like. i also don't like the name sequence because it has a specific meaning for monads.

in the past, i have used "link" for a function like this.

katrinafyi (approved):
no major objections, some questions.  

broad comments:
- there seem to be more than a few instances of a big union type like "EventuallyCall | NonCallStatement | EventuallyStatement | EventuallyJump". is this not expressable any other way?
- lots of overloads. are these all needed?
- overloads are also mixed with implicit conversions. it would be nice to standardise one way or the other.

src/main/scala/ir/dsl/InfixConstructors.scala
ailrst: Yeah its pretty obvious what + means for booleans

src/main/scala/ir/dsl/InfixConstructors.scala
ailrst: Its not actually possible to construct an expression with MapType so this is unreachable

src/main/scala/ir/dsl/WhileDSL.scala
ailrst: I don't think so, the contract of sequence doesn't allow empty lists but I guess you can move the pattern match inside the fold.

src/main/scala/ir/dsl/WhileDSL.scala
ailrst: Sequence comes from sequential composition

build.sc
ailrst: No its to get the repl to work but it doesn't like our code anyway

src/main/scala/ir/dsl/WhileDSL.scala
katrinafyi: I know. in any case, this is not blocking

src/main/scala/ir/dsl/InfixConstructors.scala
katrinafyi: I disagree, but not blocking.

src/main/scala/ir/dsl/InfixConstructors.scala
katrinafyi: I understand. thanks for fixing the ??? elsewhere

src/main/scala/ir/dsl/WhileDSL.scala
```diff
@@ -0,0 +1,196 @@
...
+  * Sequence a list of lists of blocks, where each is its own section
+  * with designated entry and exit, 
+  * so that each list element is executes in sequential order.
+  */
+def blocks(blocks: List[EventuallyBlock]*): List[EventuallyBlock] = {
+  blocks.toList match {
+    case Nil => List()
+    case h :: Nil => h
+    case h :: tail => tail.foldLeft(h)(sequence)
+  }
```
katrinafyi, line 122:
```suggestion
  require(blocks.nonEmpty)
  blocks.tail.foldLeft(blocks.head)(sequence).toList
```
idk

katrinafyi (approved):

src/main/scala/ir/eval/BitVectorEval.scala
```diff
@@ -29,6 +29,14 @@ object BitVectorEval {
     if isNegative(b) then b.value - BigInt(2).pow(b.size)
     else b.value
 
+  /** Converts a signed integer value to its corresponding Bitvector, 
+   *  assuming it falls within the representable range of the bitvector
+   *  with size [[size]].
+   *
+    */
+  def signedInt2BV(size: Int, i: BigInt): BitVecLiteral =
+    if (i > 0) then BitVecLiteral(i, size) else smt_bvneg(BitVecLiteral(-i, size))
```
katrinafyi, line 38: if you pass in `i = INT_MIN`, this might fail to construct a `BitVecLiteral(-INT_MIN, ...)` because twos complement idk

PR #356, opened 2025-03-10
Interpreter: use Vector for trace instead of List

ailrst:
This was concatenating to the end of a list making the tracing interpreter quadratic (ty @katrinafyi for spotting this). 

This makes the runtime go from ~30s to 4s for the examples here https://github.com/UQ-PAC/BASIL/pull/322.

Commenting out the disabled `Logger.debug` calls seems to save ~1% of runtime, since these aren't actually evaluated I can only assume its the overhead of 2-3 method calls and branching.

### DIFF
--- a/src/main/scala/ir/eval/InterpretTrace.scala
+++ b/src/main/scala/ir/eval/InterpretTrace.scala
@@ -64,7 +64,7 @@ def interpretWithTrace[I](
   innerInterpreter: Effects[I, InterpreterError],
   innerInitialState: I
 ): (I, Trace) = {
-  InterpFuns.interpretProg(tracingInterpreter(innerInterpreter))(p, (innerInitialState, Trace(Vector())))
+  InterpFuns.interpretProg(tracingInterpreter(innerInterpreter))(p, (innerInitialState, Trace.empty))
 }
 
 def interpretWithTrace[I](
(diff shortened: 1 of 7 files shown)


src/main/scala/ir/eval/InterpretTrace.scala
```diff
@@ -63,7 +64,7 @@ def interpretWithTrace[I](
   innerInterpreter: Effects[I, InterpreterError],
   innerInitialState: I
 ): (I, Trace) = {
-  InterpFuns.interpretProg(tracingInterpreter(innerInterpreter))(p, (innerInitialState, Trace(List())))
+  InterpFuns.interpretProg(tracingInterpreter(innerInterpreter))(p, (innerInitialState, Trace(Vector())))
```
katrinafyi, line 67: this and other places can use `Trace.empty`. or, a default value should be added to `Trace` to allow `Trace()`.

src/main/scala/util/PerformanceTimer.scala
```diff
@@ -2,6 +2,30 @@ package util
...
+    require(!inside)
+    inside = true
+    entered = System.currentTimeMillis()
+  }
+
+  def exit() = {
+    require(inside)
+    inside = false
+    total += (System.currentTimeMillis() - entered)
+  }
```
katrinafyi, line 20: you can have a method which takes a `=> T` parameter. this would allow you to write code like
```scala
region.within {
  // code to be timed
}
```

```suggestion
  def within[T](body: => T): T = {
    enter()
    val result = body
    exit()
    result
  }

  def enter() = {
    require(!inside)
    inside = true
    entered = System.currentTimeMillis()
  }

  def exit() = {
    require(inside)
    inside = false
    total += (System.currentTimeMillis() - entered)
  }
```

katrinafyi (approved):
thank you for the changes. this is a cons list moment

katrinafyi (approved):
lgtm, still some instances of `Trace(Vector())`

PR #362, opened 2025-03-11
Fix extraspec tests

ailrst:


l-kent (approved):
Should `inlinePLTLaunchpad` then be removed from the simplify pass? Otherwise seems fine

PR #365, opened 2025-03-12
fix: deprecation warnings caused by .toArray

katrinafyi:
The .toArray method converts an iterable to an Array. However, this is a minimal Java-esque array and does not support any of Scala's interfaces. Where an interface like Iterable or Seq was needed, this would require a coercion from Array to ArraySeq, which the compiler warns about because it incurs an additional copy.

This fixes the deprecation warning by using converting directly to ArraySeq, avoiding the implicit coercion and unnecessary copy.

**Also,** enables -deprecation in the mill build config. This will show deprecation warnings in the CI job.

Under the umbrella of: https://github.com/UQ-PAC/BASIL/issues/339.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -9,7 +9,7 @@ import contrib.scalapblib._
 object basil extends RootModule with ScalaModule with antlr.AntlrModule with ScalaPBModule {
   def scalaVersion = "3.3.4"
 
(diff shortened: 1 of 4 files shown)


ailrst:
Thanks

PR #372, opened 2025-03-14
tests: redirect test stdout/err to info() messages 

katrinafyi:
Since test suits have been parallelised, there is this problem where logging stdout/stderr is printed to console and it is not clear which test suite an output message relates to. This is because the standard streams are shared by all threads.

This change will redirect stdout/err within test cases to info() messages.
This helps organise the test output, especially when running in
parallel. With this change, stdout/stderr are redirected through
scalatest's info() method, allowing the output to be associated and
printed with the name of the test case.

The output now looks something like this. You can see the stdout/err lines are prefixed with `+` and indented underneath the test case name.
image

This change does have some disadvantages.
- Both stdout/stderr are redirected to info() with no distinction between them, making it impossible to pipe only stdout/stderr into, e.g., grep.
- The mechanism for redirecting the standard streams is very stateful. This is unpleasing and might complicate things if we would like more parallelism in the future.
- There's a lot of Java code. This could have been written in Scala, but it's very tied to Java-specific APIs so it felt more fitting to use Java.

### DIFF
--- a/src/main/scala/util/Logging.scala
+++ b/src/main/scala/util/Logging.scala
@@ -14,9 +14,12 @@ enum LogLevel(val id: Int):
 class GenericLogger(
   val name: String,
   val defaultLevel: LogLevel = LogLevel.INFO,
-  output: => PrintStream = Console.out,
+  defaultOutput: => PrintStream = Console.out,
   var ANSIColour: Boolean = true
 ) {
+
+  private var _output = () => defaultOutput
+  def output = _output()
 
   val children: HashSet[GenericLogger] = HashSet()
 
@@ -38,7 +41,7 @@ class GenericLogger(
   def disableColour(setChildren: Boolean = false) = setColour(false, setChildren)
   def enableColour(setChildren: Boolean = false): Unit = setColour(true, setChildren)
 
-  def deriveLogger(sname: String, stream: PrintStream): GenericLogger = {
+  def deriveLogger(sname: String, stream: => PrintStream): GenericLogger = {
     val l = GenericLogger(name + "." + sname, level, stream, ANSIColour)
     children.add(l)
     l
@@ -50,7 +53,7 @@ class GenericLogger(
 
   def deriveLogger(name: String): GenericLogger = deriveLogger(name, output)
 
-  // def setOutput(stream: PrintStream) = output = stream
+  def setOutput(stream: => PrintStream) = _output = () => stream
 
   def writeToFile(file: File, content: => String) = {
     if (level.id < LogLevel.OFF.id) {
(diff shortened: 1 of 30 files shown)


src/main/scala/util/Logging.scala
```diff
@@ -50,7 +50,7 @@ class GenericLogger(
 
   def deriveLogger(name: String): GenericLogger = deriveLogger(name, output)
 
-  def setOutput(stream: PrintStream) = output = stream
+  // def setOutput(stream: PrintStream) = output = stream
```
ailrst, line 53: Why did you comment this out

src/main/scala/util/Logging.scala
```diff
@@ -50,7 +50,7 @@ class GenericLogger(
...
 
   def writeToFile(file: File, content: => String) = {
     if (level.id < LogLevel.OFF.id) {
```
ailrst, line 56: The log line below doesn't seem to get captured for some reason, as well as StaticAnalysisLogger and others:

```diff
diff --git a/src/test/scala/SystemTests.scala b/src/test/scala/SystemTests.scala
index 679911064..e807c11d7 100644
--- a/src/test/scala/SystemTests.scala
+++ b/src/test/scala/SystemTests.scala
@@ -46,8 +46,7 @@ trait SystemTests extends AnyFunSuite, test_util.CaptureOutput, BASILTest, TestC
       s"$name,$passed,$verified,$shouldVerify,$hasExpected,$timedOut,$matchesExpected,$translateTime,$verifyTime"
   }
 
-  Logger.setLevel(LogLevel.WARN)
-  DebugDumpIRLogger.setLevel(LogLevel.OFF)
+  Logger.setLevel(LogLevel.DEBUG, true)
 
   object TestResult {
     val csvHeader =
```

`mill test.testOnly AnalysisSystemTests*`

image

src/main/scala/util/Logging.scala
katrinafyi: i had changed output to a by-name parameter and those can't be reassigned. setOutput was never used anyway.

we could, instead, bring back setOutput and use it to do the redirecting within the logger. however, the current code is nice because it redirects all stdout/stderr.

src/main/scala/util/Logging.scala
katrinafyi: Thanks for spotting this! I have realised this is because the printstream needs perfect forwarding through the deriveLogger call

katrinafyi:
Both preceding comments should be addressed by https://github.com/UQ-PAC/BASIL/pull/372/commits/e2eeccc08771449f54012ba7a909ef3b57db8f45

PR #374, opened 2025-03-17
Fix dsa deprecated

sadrabt:


ailrst:
There's still one deprecation warning 

```
[warn] -- Deprecation Warning: /home/am/Documents/programming/2024/bil-to-boogie-translator/src/main/scala/ir/transforms/DynamicSingleAssignment.scala:82:30 
[warn] 82 |            renames = renames + (lhs -> count(lhs))
[warn]    |                      ^^^^^^^^^
[warn]    |method + in trait MapOps is deprecated since 2.13.0: Consider requiring an immutable Map or fall back to Map.concat.
```

ailrst (approved):

PR #377, opened 2025-03-18
Memory transform

sadrabt:
Transforms memory accesses to register assignments using the results of DSA when --memory-transform flag is set
(requires --simplify and --dsa norm)

- creates a global graph (single global region node) and uses it to propagate global constraints among graphs of all procedures 
- replaces memory stores to stack only regions with local assignments
- replaces memory stores to global only regions with memory assignments
- Adds invariant to check unresolved expressions are only used in memory accesses that are unreachable
- transforms memory accesses by splitting the memory to disjoint regions using Data structure analysis results
- determines which values escape a procedure using DSA flags

### DIFF
--- a/src/main/scala/ir/Statement.scala
+++ b/src/main/scala/ir/Statement.scala
@@ -54,12 +54,7 @@ object MemoryAssign {
   def unapply(l: MemoryAssign): Some[(Variable, Expr, Option[String])] = Some(l.lhs, l.rhs, l.label)
 }
 
-class LocalAssign(
-  var lhs: Variable,
-  var rhs: Expr,
-  override val label: Option[String] = None,
-  val reducible: Boolean = true
-) extends SingleAssign {
+class LocalAssign(var lhs: Variable, var rhs: Expr, override val label: Option[String] = None) extends SingleAssign {
   override def modifies: Set[Global] = lhs match {
     case r: Register => Set(r)
     case _ => Set()
--- a/src/main/scala/ir/Statement.scala
+++ b/src/main/scala/ir/Statement.scala
@@ -41,14 +41,17 @@ sealed trait SingleAssign extends Assign {
   override def assignees = Set(lhs)
 }
 
-class MemoryAssign(var lhs: Register, var rhs: Expr, override val label: Option[String] = None) extends SingleAssign {
-  override def modifies: Set[Global] = Set(lhs)
+class MemoryAssign(var lhs: Variable, var rhs: Expr, override val label: Option[String] = None) extends SingleAssign {
+  override def modifies: Set[Global] = lhs match
+    case r: Register => Set(r)
+    case _ => Set()
+
   override def toString: String = s"$labelStr$lhs := $rhs"
   override def acceptVisit(visitor: Visitor): Statement = visitor.visitMemoryAssign(this)
 }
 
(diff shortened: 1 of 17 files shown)


src/main/scala/ir/Statement.scala
```diff
@@ -41,7 +41,22 @@ sealed trait SingleAssign extends Assign {
   override def assignees = Set(lhs)
 }
 
-class LocalAssign(var lhs: Variable, var rhs: Expr, override val label: Option[String] = None) extends SingleAssign {
+class MemoryAssign(var lhs: Register, var rhs: Expr, override val label: Option[String] = None) extends SingleAssign {
```
ailrst, line 44: Can we make lhs a Variable as we likely want locally-scoped observable scalar memory operations.

src/test/memory_transform/clasloc/gcc/clasloc.adt
```diff
@@ -0,0 +1,530 @@
+Project(Attrs([Attr("filename","\"clasloc.out\""),
```
ailrst, line 1: If we are only checking in results from one lifter can we use the gtirb results rather than BAP

ailrst (commented):
The variable names in the output aren't going to be valid boogie identifiers, they should probably be renamed in the transform

src/main/scala/ir/cilvisitor/CILVisitor.scala
```diff
@@ -161,6 +161,10 @@ class CILVisitorImpl(val v: CILVisitor) {
         m.mem = visit_mem(m.mem)
         m.lhs = visit_lvar(m.lhs)
         m
+      case m: MemoryAssign =>
+        m.rhs = visit_expr(m.rhs)
+        m.lhs = visit_lvar(m.lhs).asInstanceOf[Register]
```
ailrst, line 166: Can remove this case when changing lhs to Variable

sadrabt:
do i need to update ResolveSpec in IRtoBoogie to have the spec include accesses to those globals

src/main/scala/ir/transforms/MemoryTransform.scala
```diff
@@ -0,0 +1,177 @@
...
+            assert(indices.map(_.getPointee).toSet.size == 1, s"$proc, ${indices.map(_.getPointee).size}, $load")
+            val index = indices.head
+            val flag = joinFlags(indices)
+            val value = index.getPointee
+            if isGlobal(flag) then
+              ChangeTo(
+                List(
+                  LocalAssign(
+                    load.lhs,
+                    Register(s"Global_${index.interval.move(i => i - index.node.bases(Global))}", load.size),
```
ailrst, line 104: Can you factor out the name generation to a single function? I also think this is not producing unique enough names, for one example I get

```
memory mem : map bv8[bv64];
var Global_0_0 : bv32;
var Global_0_0 : bv64;
var Global_0_0 : bv128;
```

ailrst:
StackIdentification should be disabled by --memory-transform probably

https://github.com/UQ-PAC/BASIL/blob/7a851c72644bdd605979e577b3ab09a4591287cd/src/main/scala/util/RunUtils.scala#L321-L324

ailrst (approved):

PR #381, opened 2025-03-21
micro-optimisations: cached hashcode and reduce hashmap lookups

katrinafyi:
This implements two separate micro-optimisations to speed up the simplification pass, using cntlm-noduk as a benchmark and guided by profiling.

- First, we cache the hashCode() result of case classes the first time they are computed. Given the prevalence of hashmaps, this is a very frequent operation and hashCode of strings and bigints is slow (enough) that it matters.
- Second, we make some very minor changes to the simp CopyProp to reduce the number of hashmap searches needed.  This is again a hotspot and each individual `c(l)` performs a lookup of the hashmap.

I have benchmarked this against 72338dd25eb984bd8dcec83e773327e1fe0d0573 and we see a 36\% speedup on this cntlm-noduk example. (The size of this example might exaggerate the performance improvements.)

```python
Benchmark 1: java -jar original.jar --load-directory-gtirb examples/cntlm-noduk/ --simplify
  Time (mean ± σ):     68.636 s ±  2.387 s    [User: 106.720 s, System: 1.255 s]
  Range (min … max):   64.802 s … 72.607 s    10 runs
 
Benchmark 2: java -jar new.jar --load-directory-gtirb examples/cntlm-noduk/ --simplify
  Time (mean ± σ):     50.367 s ±  0.967 s    [User: 91.243 s, System: 1.328 s]
  Range (min … max):   48.822 s … 51.631 s    10 runs
 
Summary
  java -jar new.jar --load-directory-gtirb examples/cntlm-noduk/ --simplify ran
    1.36 ± 0.05 times faster than java -jar original.jar --load-directory-gtirb examples/cntlm-noduk/ --simplify
```

To show the impact of only the hashcode optimisation, we benchmarked that separately as well.
```python
Benchmark 1: java -jar cachedhashcode.jar --load-directory-gtirb examples/cntlm-noduk/ --simplify
  Time (mean ± σ):     59.590 s ±  2.890 s    [User: 96.942 s, System: 1.403 s]
  Range (min … max):   56.284 s … 65.880 s    10 runs
```

These benchmarks use the hyperfine tool and can be run with:
```bash
git checkout main
./mill assembly
cp out/assembly.dest/out.jar original.jar

git checkout microopt
./mill assembly
cp out/assembly.dest/out.jar asdf.jar
hyperfine 'java -jar original.jar --load-directory-gtirb examples/cntlm-noduk/ --simplify' 'java -jar asdf.jar --load-directory-gtirb examples/cntlm-noduk/ --simplify'
```
This will take at least 20 minutes because hyperfine repeats each version at least 10 times.

### DIFF
--- a/src/main/scala/util/CachedHashCode.scala
+++ b/src/main/scala/util/CachedHashCode.scala
@@ -1,0 +1,22 @@
+package util
+
+/**
+ * A mixin for **case classes** to cache the hashCode the first time it is
+ * computed. This can yield speedups for case classes which contain complex
+ * structures (e.g., string list) and/or values which are frequently searched
+ * in a hashed data structure.
+ *
+ * WARNING: this must /not/ be used for classes which have mutable fields.
+ *          doing so would lead to inconsistencies between equals and hashcode,
+ *          causing unpredictable (read: incorrect) behaviour in hashmap/set.
+ */
+trait CachedHashCode {
+  this: Product =>
+
+  // this is the default implementation of hashCode for case classes.
+  // we need to call it explicitly because Scala will not generate a hashCode
+  // if we have provided one ourself.
+  private lazy val hash = scala.runtime.ScalaRunTime._hashCode(this)
+
+  override def hashCode = hash
+}
--- a/src/main/scala/ir/Expr.scala
+++ b/src/main/scala/ir/Expr.scala
@@ -1,5 +1,6 @@
 package ir
 import boogie._
+import util.CachedHashCode
 import scala.collection.mutable
 
 sealed trait Expr {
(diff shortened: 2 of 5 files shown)


src/main/scala/util/CachedHashCode.scala
```diff
@@ -0,0 +1,22 @@
...
+trait CachedHashCode {
+  this: Product =>
+
+  // this is the default implementation of hashCode for case classes.
+  // we need to call it explicitly because Scala will not generate a hashCode
+  // if we have provided one ourself.
+  private lazy val hash = scala.runtime.ScalaRunTime._hashCode(this)
+
+  override def hashCode = hash
+}
```
ailrst, line 22: I'd like to be sure this has the same behaviour as before, can you find documentation on runtime hashcode or a test that this is still hashing the field values of the case classes, otherwise LGTM.

ailrst (approved):

Issue #378, opened 2025-03-19
Assertion failure in `Simplify :: DynamicSingleAssignment`


aaronbembenek:
Hi there,

I tried running the attached program with BASIL (commit 4cad6a5) options `--analyse --memory-regions dsa --simplify`. I get this error:

```
[INFO]   [!] Loading Program
[INFO]   [!] Removing external function calls
[INFO]   [!] Replacing Indirect Calls
[INFO]   [!] Generating Procedure Summaries
[INFO]   [!] Running Symbolic Access Analysis
[INFO]   [!] Running DSA Analysis
[INFO]   [!] Finished indirect call resolution after 1 iterations
[INFO]   [!] Running Simplify
[INFO]   [!] Simplify :: DynamicSingleAssignment
Exception in thread "main" java.lang.AssertionError: assertion failed: should always have at least one processed predecessor List(Block errorFn_1848__0__b8tsihT4Q6a_SWPo4w8HoA with LocalVar(R0, bv64) := LocalVar(R0_in, bv64)
LocalVar(R10, bv64) := LocalVar(R10_in, bv64)
LocalVar(R11, bv64) := LocalVar(R11_in, bv64)
LocalVar(R12, bv64) := LocalVar(R12_in, bv64)
LocalVar(R13, bv64) := LocalVar(R13_in, bv64)
LocalVar(R14, bv64) := LocalVar(R14_in, bv64)
LocalVar(R15, bv64) := LocalVar(R15_in, bv64)
LocalVar(R16, bv64) := LocalVar(R16_in, bv64)
LocalVar(R17, bv64) := LocalVar(R17_in, bv64)
LocalVar(R18, bv64) := LocalVar(R18_in, bv64)
LocalVar(R1, bv64) := LocalVar(R1_in, bv64)
LocalVar(R29, bv64) := LocalVar(R29_in, bv64)
LocalVar(R2, bv64) := LocalVar(R2_in, bv64)
LocalVar(R30, bv64) := LocalVar(R30_in, bv64)
LocalVar(R31, bv64) := LocalVar(R31_in, bv64)
LocalVar(R3, bv64) := LocalVar(R3_in, bv64)
LocalVar(R4, bv64) := LocalVar(R4_in, bv64)
LocalVar(R5, bv64) := LocalVar(R5_in, bv64)
LocalVar(R6, bv64) := LocalVar(R6_in, bv64)
LocalVar(R7, bv64) := LocalVar(R7_in, bv64)
[...]

PR #384, opened 2025-03-26
Fix dsa assertion

ailrst:
Addresses https://github.com/UQ-PAC/BASIL/issues/378

The offending function was `errorFn_1848`, which was a single empty block jumping to itself, after param anslysis we got the following:

```
proc errorFn_1848(R0_in:bv64, R10_in:bv64, R11_in:bv64, R12_in:bv64, R13_in:bv64, R14_in:bv64, R15_in:bv64, R16_in:bv64, R17_in:bv64, R18_in:bv64, R1_in:bv64, R29_in:bv64, R2_in:bv64, R30_in:bv64, R31_in:bv64, R3_in:bv64, R4_in:bv64, R5_in:bv64, R6_in:bv64, R7_in:bv64, R8_in:bv64, R9_in:bv64) -> ()
{
  name = "errorFn";
  address = 0x738;
  entry_block = "errorFn_1848__0__b8tsihT4Q6a_SWPo4w8HoA";
  blocks = [
    block errorFn_1848__0__b8tsihT4Q6a_SWPo4w8HoA {address = 0x738} [
      var R0: bv64 := R0_in:bv64;
      var R10: bv64 := R10_in:bv64;
      var R11: bv64 := R11_in:bv64;
      var R12: bv64 := R12_in:bv64;
      var R13: bv64 := R13_in:bv64;
      var R14: bv64 := R14_in:bv64;
      var R15: bv64 := R15_in:bv64;
      var R16: bv64 := R16_in:bv64;
      var R17: bv64 := R17_in:bv64;
      var R18: bv64 := R18_in:bv64;
      var R1: bv64 := R1_in:bv64;
      var R29: bv64 := R29_in:bv64;
      var R2: bv64 := R2_in:bv64;
      var R30: bv64 := R30_in:bv64;
      var R31: bv64 := R31_in:bv64;
      var R3: bv64 := R3_in:bv64;
      var R4: bv64 := R4_in:bv64;
      var R5: bv64 := R5_in:bv64;
      var R6: bv64 := R6_in:bv64;
      var R7: bv64 := R7_in:bv64;
      var R8: bv64 := R8_in:bv64;
      var R9: bv64 := R9_in:bv64;
      var R30_begin: bv64 := R30:bv64;
      goto(errorFn_1848__0__b8tsihT4Q6a_SWPo4w8HoA)
    ]
  ];
};
```

These generated register var <- param assignments don't make sense because they are part of this self-loop.

We fix this by

1. Ensure the procedure entry node does not participate in a loop (has no incoming edge) with an early transform (this alone is enough to fix the assertion because the param analysis was creating nonsense, but the DSA analysis was also not really behaving correctly with respect to that nonsense:
2. Make DSA also increment SSA count if a variable use is encountered which is not a parameter and has not been renamed, add both parameters and these renames to the DSA state so that it performs the correct renaming for block self-loops containing variables not defined earlier in the program. Examples of this can be seen in SingleAssignmentTest.scala.

### DIFF
--- a/src/main/scala/util/RunUtils.scala
+++ b/src/main/scala/util/RunUtils.scala
@@ -259,6 +259,8 @@ object IRTransform {
       }
     }
 
+    ctx.program.procedures.foreach(ir.transforms.makeProcEntryNonLoop)
+
     // useful for ReplaceReturns
     // (pushes single block with `Unreachable` into its predecessor)
     while (transforms.coalesceBlocks(ctx.program)) {}
@@ -291,6 +293,7 @@ object IRTransform {
     assert(invariant.singleCallBlockEnd(ctx.program))
     assert(invariant.cfgCorrect(ctx.program))
     assert(invariant.blocksUniqueToEachProcedure(ctx.program))
+    assert(invariant.procEntryNoIncoming(ctx.program))
     ctx
   }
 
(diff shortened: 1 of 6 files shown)


src/main/scala/util/RunUtils.scala
```diff
@@ -740,6 +743,13 @@ object RunUtils {
     DebugDumpIRLogger.writeToFile(File("il-before-dsa.il"), pp_prog(program))
 
     transforms.OnePassDSA().applyTransform(program)
+    if (DebugDumpIRLogger.getLevel().id < LogLevel.OFF.id) {
```
katrinafyi, line 746: slightly surprising

katrinafyi (approved):

PR #396, opened 2025-04-09
gtirb loader: support ite.0 expressions 

katrinafyi:
implements support for ite.0 expressions by converting them to the existing
TempIf structure. then, there is already code to convert TempIf to blocks
and gotos.

adds a new test case containing an sdiv instruction. aslp emits ite to
correctly represent the semantics of returning zero for overflowing
divisions.

the TempIf class causes unending suffering, because it
masquerades as a NOP and interacts poorly with every other piece of
automation in basil. it would be really good to rewrite the gtirb
frontend... sometime.

honestly, maybe this should have been implemented in aslp-side.......

it is also awful to test the frontend because of how interlinked everything is.
all the loading functions expect a monolithic config object, then dispatch to any
number of helper functions which also expect more arguments.
often, these functions make choices to suppress errors and only report the errors
to a logger...

before (here, we have made the loader emit the ite as an "ite.temp" uninterpreted function for visibility)
```python
  block("main_1920__0__xdU61Ad4R3aE_hVJ17n5eQ$__1",
    Assume(UnaryExpr(BoolNOT, BinaryExpr(BVEQ, Register("R1", 64), BitVecLiteral(BigInt("0"), 64))), None, None, true),
    LocalAssign(Register("R0", 64), Extract(64, 0, UninterpretedFunction("ite.temp", Seq(
      BinaryExpr(BoolAND, BinaryExpr(BVEQ, Register("R0", 64), BitVecLiteral(BigInt("9223372036854775808"), 64)), BinaryExpr(BVEQ, Register("R1", 64), BitVecLiteral(BigInt("18446744073709551615"), 64))),
      BitVecLiteral(BigInt("9223372036854775808"), 128),
      SignExtend(64, BinaryExpr(BVSDIV, Register("R0", 64), Register("R1", 64)))
    ), BitVecType(128))), Some("1924_0")),
    ret
  )
```

after:
```python
  block("main_1920__0__xdU61Ad4R3aE_hVJ17n5eQ$__1",
    Assume(UnaryExpr(BoolNOT, BinaryExpr(BVEQ, Register("R1", 64), BitVecLiteral(BigInt("0"), 64))), None, None, true),
    goto("main_1920__0__xdU61Ad4R3aE_hVJ17n5eQ$__2", "main_1920__0__xdU61Ad4R3aE_hVJ17n5eQ$__3")
  ),
  block("main_1920__0__xdU61Ad4R3aE_hVJ17n5eQ$__2",
    Assume(BinaryExpr(BoolAND, BinaryExpr(BVEQ, Register("R0", 64), BitVecLiteral(BigInt("9223372036854775808"), 64)), BinaryExpr(BVEQ, Register("R1", 64), BitVecLiteral(BigInt("18446744073709551615"), 64))), None, None, true),
    LocalAssign(LocalVar("ite_result_xdU61Ad4R3aE_hVJ17n5eQ==_1", BitVecType(128), 0), BitVecLiteral(BigInt("9223372036854775808"), 128), None),
    goto("main_1920__0__xdU61Ad4R3aE_hVJ17n5eQ$__4")
  ),
  block("main_1920__0__xdU61Ad4R3aE_hVJ17n5eQ$__3",
    Assume(UnaryExpr(BoolNOT, BinaryExpr(BoolAND, BinaryExpr(BVEQ, Register("R0", 64), BitVecLiteral(BigInt("9223372036854775808"), 64)), BinaryExpr(BVEQ, Register("R1", 64), BitVecLiteral(BigInt("18446744073709551615"), 64)))), None, None, true),
    LocalAssign(LocalVar("ite_result_xdU61Ad4R3aE_hVJ17n5eQ==_1", BitVecType(128), 0), SignExtend(64, BinaryExpr(BVSDIV, Register("R0", 64), Register("R1", 64))), None),
    goto("main_1920__0__xdU61Ad4R3aE_hVJ17n5eQ$__4")
  ),
  block("main_1920__0__xdU61Ad4R3aE_hVJ17n5eQ$__4",
    LocalAssign(Register("R0", 64), Extract(64, 0, LocalVar("ite_result_xdU61Ad4R3aE_hVJ17n5eQ==_1", BitVecType(128), 0)), Some("1924_0")),
    ret
  )
```


ailrst:
It's borken

```
$ mill run -i ../csmith-tests/csmith2/gcc/csmith2.gts -r ../csmith-tests/csmith2/gcc/csmith2.relf --dump-il c
[INFO]   [!] Loading Program
Exception in thread "main" scala.MatchError: List(Register(V0, bv128)[64:0], Register(V1, bv128)[64:0], true, Register(FPCR, bv32)) (of class scala.collection.immutable.$colon$colon)
	at translating.ConvertITEToTempIf.vexpr(GTIRBToIR.scala:84)
	at ir.cilvisitor.CILVisitorImpl.visit_expr(CILVisitor.scala:140)
	at ir.cilvisitor.CILVisitorImpl.continue$3(CILVisitor.scala:165)
	at ir.cilvisitor.CILVisitorImpl.visit_stmt$$anonfun$1(CILVisitor.scala:176)
	at scala.collection.immutable.List.map(List.scala:247)
	at ir.cilvisitor.CILVisitor$package$.doVisitList(CILVisitor.scala:41)
	at ir.cilvisitor.CILVisitorImpl.visit_stmt(CILVisitor.scala:176)
	at ir.cilvisitor.CILVisitor$package$.visit_stmt(CILVisitor.scala:238)
	at translating.ConvertITEToTempIf.convertStatements$$anonfun$1(GTIRBToIR.scala:100)
	at scala.collection.StrictOptimizedIterableOps.flatMap(StrictOptimizedIterableOps.scala:118)
	at scala.collection.StrictOptimizedIterableOps.flatMap$(StrictOptimizedIterableOps.scala:105)
	at scala.collection.mutable.ArrayBuffer.flatMap(ArrayBuffer.scala:42)
	at translating.ConvertITEToTempIf.convertStatements(GTIRBToIR.scala:100)
	at translating.GTIRBToIR.convertITEToTempIF(GTIRBToIR.scala:300)
	at translating.GTIRBToIR.createIR$$anonfun$4$$anonfun$1(GTIRBToIR.scala:239)
	at scala.collection.immutable.HashSet.foreach(HashSet.scala:951)
	at translating.GTIRBToIR.createIR$$anonfun$4(GTIRBToIR.scala:233)
	at scala.runtime.function.JProcedure1.apply(JProcedure1.java:15)
	at scala.runtime.function.JProcedure1.apply(JProcedure1.java:10)
	at scala.collection.IterableOnceOps.foreach(IterableOnce.scala:619)
	at scala.collection.IterableOnceOps.foreach$(IterableOnce.scala:617)
	at scala.collection.AbstractIterable.foreach(Iterable.scala:935)
	at scala.collection.IterableOps$WithFilter.foreach(Iterable.scala:905)
	at translating.GTIRBToIR.createIR(GTIRBToIR.scala:230)
	at util.IRLoading$.loadGTIRB(RunUtils.scala:211)
	at util.IRLoading$.load(RunUtils.scala:131)
	at util.RunUtils$.$anonfun$19(RunUtils.scala:831)
	at scala.Option.getOrElse(Option.scala:201)
	at util.RunUtils$.loadAndTranslate(RunUtils.scala:831)
	at util.RunUtils$.run(RunUtils.scala:694)
	at Main$.main(Main.scala:273)
	at Main.main(Main.scala
)
```
[gcc.zip](https://github.com/user-attachments/files/19662898/gcc.zip)

src/main/scala/translating/GTIRBToIR.scala
```diff
@@ -38,10 +39,68 @@ import util.Logger
...
+) extends NOP(label) {
+  override def toString = s"TEMPIF($cond, $thenStmts, $elseStmts)"
+}
+
+class FindUninterpretedFunctions extends CILVisitor {
+  var uninterp = List[UninterpretedFunction]()
+
+  override def vexpr(e: Expr) =
+    e match
+      case u: UninterpretedFunction => uninterp = u :: uninterp
```
katrinafyi, line 54: needs predicate guard

src/main/scala/translating/GTIRBToIR.scala
```diff
@@ -38,10 +39,68 @@ import util.Logger
...
+    iteResults = iteResults.empty
+    ChangeDoChildrenPost(
+      List(s),
+      ss => {
+        val x = iteResults.values ++: ss
+        iteResults = iteResults.empty
+        x
+      }
+    )
+  }
```
katrinafyi, line 102: move recursive function inside here

src/test/correct/sdiv/clang_O2/sdiv.adt
```diff
@@ -0,0 +1,485 @@
+Project(Attrs([Attr("filename","\"a.out\""),
```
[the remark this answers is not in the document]
katrinafyi: is there a way to choose the lifter in the current makefile system?

l-kent:
Looking at the result this produces I'm not sure that the division implementation in ASLp makes sense at present? Why does it sign extend the result of the sdiv operation to 128 bits and then extract the bottom 64 bits from it? This extension seems pointless and that surely can't be the intended behaviour?

ailrst:
Its extending just to ensure there is enough precision to represent the result of the ite, likely a bit too conservatively. The division is represented by the entire block

```

**Comment by l-kent** [comment]
I'm not sure why sdiv needs to be implemented like this either, the check that the ite is making seems to be redundant?

The SMT-LIB bvsdiv operator is defined as follows:
```
    (bvsdiv s t) abbreviates
      (let ((?msb_s ((_ extract |m-1| |m-1|) s))
            (?msb_t ((_ extract |m-1| |m-1|) t)))
        (ite (and (= ?msb_s #b0) (= ?msb_t #b0))
             (bvudiv s t)
        (ite (and (= ?msb_s #b1) (= ?msb_t #b0))
             (bvneg (bvudiv (bvneg s) t))
        (ite (and (= ?msb_s #b0) (= ?msb_t #b1))
             (bvneg (bvudiv s (bvneg t)))
             (bvudiv (bvneg s) (bvneg t))))))
```
For the case where `s == INT_MIN && t == -1`, we get `(bvudiv (bvneg s) (bvneg t)))` as the most significant bit is 1 for both, so, the result is INT_MIN which is all this ite does:
```
bvneg INT_MIN == INT_MIN
bvneg -1 == 1
bvudiv INT_MIN 1 = INT_MIN
```

l-kent:
> Its extending just to ensure there is enough precision to represent the result of the ite, possibly a bit too conservatively.

It doesn't do this though. It extends the result of the 64-bit bvsdiv operation to 128-bits. If extra precision is needed then the bvsdiv operands need to be extended, not the result.

ailrst:
ite_result needs to have a type that can store the join of `bvsdiv64(R8, R1)` and `9223372036854775808`, the extend is just there to match the type of the overflow result branch.

l-kent:
Yes, there is no reason for 9223372036854775808bv128 to be 128 bits because it fits as 64 bits? As can clearly be seen by the earlier check if R8 == 9223372036854775808bv64 (which is INT_MIN).

ailrst:
I know thats the case for this case, but its possible asli can't prove this internally when the ite is generated.

ailrst:
I'm not totally certain of the motivation, I think this encoding helped some of the alive2 translation validation examples go through

l-kent:
The ite just seems to be checking for this single case, that the dividend == INT_MIN and the divisor == -1, which bvsdiv already should produce the correct result of INT_MIN for, so I'm unsure how the extension could ever be necessary.

Do you know where those examples are?

l-kent:
The relevant lines in ASLp are the following:

https://github.com/UQ-PAC/aslp/blob/fd6aba50df5fe4b4d9d4954404d91a490b7f9937/libASL/transforms.ml#L377

https://github.com/UQ-PAC/aslp/blob/fd6aba50df5fe4b4d9d4954404d91a490b7f9937/libASL/transforms.ml#L380

In this case, `wx` is 64 and `width abs` is 128, and I'm not sure how there would be a case where `wx > width abs`, so `res` should always just be a zero extended `intmin` which should make this whole thing always redundant?

l-kent:
https://github.com/UQ-PAC/aslp/blob/fd6aba50df5fe4b4d9d4954404d91a490b7f9937/tests/override.asl#L35-L38

Oh is this the issue? The sdiv implementation that's being used in ASLp isn't complaint with SMT-LIB because it doesn't handle the overflow? So the overflow handling is happening in the wrong place and getting pushed into the output for BASIL to deal with?

ncough:
To clarify the ASLp side of things, division in ASL is specified over arbitrary precision integers followed by a cast back to a bitvector of a desired size, as is the case for a variety of other arithmetic instructions. ASLp lowers these integer operations into semantically equivalent bitvector operations, generally assuming a two's complement representation, and then tries to cleanup the effects of the cast later.

To soundly lower division over two integer arguments, known to be representable as 64-bit bitvectors, the only complex case is `INT_MIN_64/-1 = -INT_MIN_64` which is not representable as a 64-bit bitvector. Representing `-INT_MIN_64` requires at least one additional `0` bit, though this gets padded out to 128-bits at present. Therefore, the `ite` is picking between a 128-bit `-INT_MIN_64` constant or a 64-bit `sdiv_bits` operation that is known not to overflow. To ensure everything type-checks, the `sdiv_bits` must be extended to match the chosen representation of `-INT_MIN_64`.

Obviously, none of the above matters given the result is immediately sliced down to a 64-bit value. But the integer elimination pass does not know this and must perform a precise lowering in anticipating of an arbitrary use of the result. A subsequent pass attempts to reason over which bits of an intermediate calculation are actually required, though division wasn't considered, likely due to it being a pain.

Adjusting this cleanup pass to support division (and reverting from the `ite` encoding of `div` to one that just sign extends) should get a representation of division purely as a single `sdiv_int`, which should be semantically equivalent to `bvsdiv`. This should be corrected by https://github.com/UQ-PAC/aslp/pull/134.

l-kent:
Thanks for that!

Do we still want to add this functionality given we no longer directly need it? If we do want the ASLp to produce an ITE expression, it may be simpler to just treat it as an expression in BASIL (this would be possible) but I'm not sure if that would cause too much trouble for the analyses or not. It wouldn't be a problem for translation to Boogie.

katrinafyi:
My hard work 😔😔😔

katrinafyi:
> we no longer directly need it?

I suppose not. I would like to rewrite the gtirb loader (eventually....)

l-kent:
What are the issues with the GTIRBLoader at present?

>the TempIf class causes unending suffering, because it masquerades as a NOP and interacts poorly with every other piece of automation in basil.

TempIf has not been intended to interact with anything outside of GTIRBLoader and GTIRBToIR, which is why no one has done any work to make things like CILVisitor support it. Extending CILVisitor to properly support it would be fairly trivial if necessary though, and would make more sense than trying to work around this present lack of support.

> often, these functions make choices to suppress errors and only report the errors to a logger...

Personally, I would prefer the loader to crash if it encounters something unexpected, so we can easily be aware of issues, but I've been overruled on that in the past.

katrinafyi:
I guess the main things which are unpleasant are in GTIRBToIR, it's just a lot of mutable maps and large conditionals to handle lots of cases. Idk it would make me uneasy.

Re TempIf, I wasn't at all suggesting that we change Basil's visitors to support it. I mean to say that the existence of TempIf is fundamentally problematic and flies in the face of sound functional programming principles. It's unthinkable that you would have a _subclass_ of an algebraic data type. It's basically subverting the type system. I'd like to avoid needing TempIf at all, but that would be a fairly big change.

Re error reporting, it's nice and obvious to throw exceptions, but unfortunately you can't resume execution if you want to disregard exceptions and continue. I think there could be a middle ground between crashing and only logging errors. When making this PR, I would've liked having functions return a `(result, List[Errors])` tuple. But again, this is just what would've been most convenient for me personally - more thought needed maybe.

PR #400, opened 2025-04-11
Support for rely-guarantee generation

j-tobler:
This PR introduces "interference domains" for abstracting interferences between threads, and provides a function for generating sound and compatible rely-guarantee conditions for concurrent programs using these domains.

This work is still in development. For the moment, we don't work on top of memory-region analysis, we ignore memory reads and writes, and we don't distinguish between global and local variables. We also don't provide a good way to define each thread (during testing, we just assume a thread is some given Procedure), and there are still a number of questions about the implementation of the interference domains that need to be resolved before deployment. The contribution of this PR is the core logic of the rely-guarantee generator and of a number of interference domains.


ailrst:
I had a go at the merge, there's possibly some regressions there that are my fault. I'm not sure what the status is with the failing tests here, is some of it unfinished work from @b-paul ? Some may just need tagging as not-working. 

Its unfortunate we didn't get to put in a pr for those changes much earlier as now this pr has two substantial additions, and falls across both the simplification-pass merge and the autoformatter changes. But it is what it is

b-paul:
Neither of the failing tests appear to be related to any incomplete work I have yet to merge

ailrst:
I'll wait to merge https://github.com/UQ-PAC/BASIL/pull/409 so its easier to review this PR in isolation

src/main/scala/analysis/rely_guarantee_generation/InterferenceDomains.scala
```diff
@@ -0,0 +1,381 @@
...
+  * 
+  * Type parameters:
+  * T: The type of lattice element representing a set of state transitions.
+  * S: The type of lattice element representing a set of states.
+  * 
+  * @param stateLattice: A lattice providing operations over S.
+  * @param stateTransfer: A transfer function for deriving reachable states.
+  */
+abstract class InterferenceDomain[T, S](
+  val stateLattice: InterferenceCompatibleLattice[S],
```
ailrst, line 30: Use `trait` rather than abstract class

src/main/scala/analysis/rely_guarantee_generation/InterferenceDomains.scala
```diff
@@ -0,0 +1,381 @@
...
+  // an overapproximation of the transitive closure of t
+  def close(t: T): T
+  // representation of t as a predicate over primed and unprimed variables
+  def toString(t: T): String
+}
+
+/* Fixme: This may result in naming clashes, however we don't always have a
+Procedure at hand to call getFreshSSAVar(...). Ideally, we might pass around
+some kind of variable factory. */
+def get_temp(ty: IRType): LocalVar = LocalVar("rg_tmp_var", ty, 0)
```
ailrst, line 51: Use `util.Counter` to get a global threadsafe counter to produce a new name on each call.

src/main/scala/util/RunUtils.scala
```diff
@@ -373,6 +373,24 @@ object IRTransform {
...
+    val stateTransfer = SignedIntervalDomain().transfer
+    val intDom = ConditionalWritesDomain[StateLatticeElement](stateLattice, stateTransfer)
+    val relyGuarantees =
+      RelyGuaranteeGenerator[InterferenceLatticeElement, StateLatticeElement](intDom).generate(threads)
+    for ((p, (rely, guar)) <- relyGuarantees) {
+      println("--- " + p.procName + " " + "-" * 50 + "\n")
+      println("Rely:")
+      println(intDom.toString(rely) + "\n")
+      println("Guarantee:")
+      println(intDom.toString(guar) + "\n")
```
ailrst, line 391: Use `Logger.info` instead of println

PR #404, opened 2025-04-17
runtime evaluation of Scala code

katrinafyi:
This adds some methods to support the evaluation of Scala code at runtime, with the main goal of evaluating Basil DSL strings such as those produced by ToScala.

This is a somewhat hacky way to close the loop of Basil IR representations and allows dumping and re-loading of Basil IR to/from disk.

This is implemented by adding a dependency on the scala3-compiler package, so we can access its script engine through the Java ScriptEngine interface:
https://docs.oracle.com/javase/8/docs/api/javax/script/ScriptEngine.html

The naming of this feature as "Eval" could be confusing next to the existing SMT eval methods.


ailrst (approved):
Adding scala compiler as a dependency is unfortunate, but it must be done. This is definitely not a great way to implement this but we've gone without this functionality for so long I think its definitely worthwhile.

PR #407, opened 2025-04-21
DSA Clean Up

sadrabt:
Some improvements to DSA readability, performance and precision. 

The precision improvements come from using the globals from the symbol table instead of using less precise heuristics for determining globals

Most of the performance improvements are gained by removing duplicate invariant checks used for debugging. 
Also dsa only creates a bounded number of nodes. It no longer creates new nodes to process merges. This allows it to no longer require compressing graph by cloning it which reduce time spent cloning. 

The dsa flags are also refactored to have three options 
Prereq, Standard and Checks,

Prereq only performs the prepass of sva and constraint generation
Standard performs dsa with fewer checks
Checks performs dsawith all the invariant checks (localCorrectness)

This pull request also includes some additional test cases which checks DSA maintains stack or globals for certain cntlm procedures where precision is fragile


l-kent:
> The precision improvements come from using the globals from the symbol table instead of using less precise heuristics for determining globals

Does this change anything for when there are accessed memory sections that aren't referenced in the symbol table? Can it still handle that type of case?

What's the reason for why one of the tests uses a new -o3 version of cntlm? 

MemoryTransfromTests should be renamed MemoryTransformTests

sadrabt:
If the addres of  memory section is within Interval [base, base + size) of a symbol table global/function entry then it can handle it as a global otherwise it throws an error. There is an exception with when a constant is used to read memory which is not in the globals/functions entries but is a symbol table entry. I think these are stack guards. the checks allow for this but only if the value is read not if it is written to

l-kent:
How does this work for src\test\indirect_calls\jumptable\clang? That has jump table addresses which are loaded from that aren't in the symbol table?

sadrabt:
I wanted to see which ones were working with O3 optimization, and use O3 to write a more robust Interval analysis that uses loop intervals to bound collapses (since there is less things put on the stack and hence may be easier to determine bound prior to dsa). But  irreducible loops has issues with O3 which break the SVA (there are dynamic assignment variables which aren't defined). So i'm happy to remove O3 from this pr

sadrabt:
for jumptable/clang I can relate back all the constants to the symbol table. could you be more specific

l-kent:
For instance, jumptable/clang loads from the address 0x10DC0 which contains a function pointer. This is listed in one of the relocation sections but not the symbol table.

A more difficult case is jumptable3/clang, which loads from 0x85C (or 0x85C + an offset), but this isn't listed as a symbol table object. These sort of cases do exist, not all global memory accesses are to memory that's listed in the symbol table.

sadrabt:
I handle relocation table and external functions as well so 10DC0 is handled, I'll look into jumptable3 now

sadrabt:
the globals won't collapse for jumptable3 but the node representing the index in the memory access with 0x85c + R0_in will. I think this is okay since we don't know what the parameter is and hence in terms of the DSA the node is incomplete. If we know about the parameter (through specification or by cloning the main procedure into another procedure) dsa will able unify the formal and actual (or specified) parameters and recognize the index is referring to a global. The alternative is to treat all constants as globals, this may be okay for the smaller examples but leads to rapid collapse in the globals in CNTLM. It will make result of DSA unusable for things like memory transform pass

l-kent:
With more sophisticated analysis it's possible to put bounds on what R0_in can be at that point, but I'm not expecting that to happen yet of course. Thanks for having a look at it.

I'm going to play around with this briefly to see if I can find a case where there are issues with relying on the symbol table & relocation sections.

l-kent:
After playing around with the compiler I'm not sure if there are other obvious cases other than the switch statement jump table case but it is something we should keep in mind at least - I can't rule out there being some cases that this approach won't work for so if we run into those this may require further work in the future.

src/main/scala/analysis/data_structure_analysis/IntervalDSA.scala
```diff
@@ -428,180 +354,119 @@ class IntervalGraph(
...
+      .map((base, set) => (base, stableNode.bases.getOrElse(base, Set.empty) ++ set.map(_ + delta.getOrElse(0))))
+      .toMap
+
+    val selfPointers = disconnectSelfPointers(stableNode)
+
+    unify(nodeToBeMoved, stableNode, delta.getOrElse(0))
+    nodeToBeMoved.cells.foreach(c =>
+      val (node, offset) = findNode(stableNode)
+      val cell = node.add(c.interval.move(i => i + offset + delta.getOrElse(0)))
+      if c.hasPointee then cell.setPointee(c.getPointee)
```
ailrst, line 463: Maybe the copying of pointees into the representative node should be part of a separate unify method? Its always a direct consequence of unification as far as I understand?

ailrst: I'm a bit unclear why there's two separate loops over `nodeToBeMoved` to copy the bases, then the intervals into the destination node and why they occur either side of the actual 'unify' call.

ailrst (commented):
This is much cleaned up and it looks pretty good, some things I'm not clear on and could be cleaned up further. The performance improvements are also really cool, I'm not sure if the deleted checks are just not useful at all or worth also putting behind a flag but its up to you.

ailrst:
It might make sense to rename the Interval class here to something alluding to its DSA-specific needs as we've just merged another `Interval` domain (in a different package path but still) which sometimes conflicts. Its up to you

src/main/scala/analysis/data_structure_analysis/IntervalDSA.scala
```diff
@@ -428,180 +354,119 @@ class IntervalGraph(
...
-        case (_, true) => Some(cell1.setPointee(cell2.getPointee))
-        case (true, _) => Some(cell2.setPointee(cell1.getPointee))
-        case (_, _) => None // Logger.warn(s"neither $cell1, or $cell2 had a pointee")
-    else Some(cell1.getPointee)
+  def isInSCC(cell: IntervalCell): Seq[IntervalCell] = {
+    var seen = Seq[IntervalCell]()
+    var cur = cell
+    while cur.hasPointee && !seen.contains(cell) && !seen.contains(cur) do
+      cur = get(cur.getPointee)
+      seen = seen.appended(cur)
```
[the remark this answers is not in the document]
sadrabt: True.  I want to collect the SCCs after each merge since new merges may create new ones, therefore, I think I won't be able to for example calculate all the sccs in a graph and cache that info.  . also I'm only interested in if a particular cell (the ones being currently merged are in sccs). I'm open if you have ideas on how this can be improved

src/main/scala/analysis/data_structure_analysis/IntervalDSA.scala
```diff
@@ -428,180 +354,119 @@ class IntervalGraph(
...
+    nodeToBeMoved.cells.foreach(c =>
+      val (node, offset) = findNode(stableNode)
+      val cell = node.add(c.interval.move(i => i + offset + delta.getOrElse(0)))
+      if c.hasPointee then cell.setPointee(c.getPointee)
     )
 
-    assert(resultNode.nonOverlappingProperty)
+    selfPointers.foreach((pointer, pointee) => find(pointer).setPointee(find(pointee)))
+    connectSCC(scc1)
+    connectSCC(scc2)
```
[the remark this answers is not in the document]
sadrabt: yes after unify cells from node 2 will have a corresponding cell in node 1. But still need to maintain the scc edges which involved cells from node 2 which were disconnected in order to bound merges

src/main/scala/analysis/data_structure_analysis/IntervalDSA.scala
sadrabt: the bases aren't used for merging/unification of cells (a possible exception may be that during BU/TD phases the bases are used to determine which cells to merge but the merging itself doesn't utilize the bases). The bases only indicate where a region starts in a node. A region may start at multiple places due to name conflicts (most common example, is a malloc wrapper getting called multiple times, the heap region produced by different calls to the wrapper would have the same name and hence could end up with a node where two different heap regions represented by the same name are merged at different offsets.

The first loop only copies base start points from nodetobemoved + delta into the stable node. However, these bases aren't enough to carry all the information about a nodes and its cells. For example they don't encode what fields (cells) a region may have and what are the outgoing edged from those cells. 

The second loop takes the cell intervals (a cell is an node, interval pair with possibly and outgoing edge to another cell) from the nodeToBeMoved shift them by delta and add them to the new node. Additionally, if the moved cell had an outgoing edge in nodeToBeMoved it sets the outgoing edge in stableNode to that (or unifies it with an already existing one). 

The first loop can occur at any point. It's extra bookkeeping that allows DSA to work backwards from a node/cell and determine the corresponding regions/fields. 

The second loop needs to occur after unify since it relies on Pointees being uptodate. Any edges from stableNode to itself are disconnected. But Edges from nodeToBeMoved to itself remain intact. but after unify these edges correspond to the updated version of their target cell (using find). Therefore any edges from nodeToBeMoved to itself now points to a cell in stableNode.

sadrabt: I moved stuff around a bit in order to make it more clear what relies on unify and what doesn't

src/main/scala/analysis/data_structure_analysis/IntervalDSA.scala
```diff
@@ -1235,7 +1146,73 @@ object IntervalDSA {
...
+  flag
+}
+
+def estimateStackSize(program: Program): Unit = {
+  program.procedures.foreach(proc =>
+    val size = proc.collectFirst {
+      case LocalAssign(_, BinaryExpr(BVADD, Register("R31", 64), arg2: BitVecLiteral), _) if isNegative(arg2) =>
+        bv2SignedInt(arg2).toInt * -1
+    }
+    proc.stackSize = size
```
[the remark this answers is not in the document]
sadrabt: stopped using this and removed it

Issue #370, opened 2025-03-13
Thread-unsafety in tests


katrinafyi:
There are test suites which fail sporadically. I do not think the test cases are significant here. I do not know if these have the same cause.

Locally
```java
DSAMemoryRegionSystemTestsBAP:

- incorrect/nestedifglobal/clang:BAP *** FAILED *** (101 milliseconds)
  java.util.NoSuchElementException: key not found: %000003ad$0: LocalVar(load20, bv32) := MemoryLoad(Memory(mem, 64, 8), bvadd64(Register(R8, bv64), 52bv64), LittleEndian, 32)
  at scala.collection.MapOps.default(Map.scala:289)
  at scala.collection.MapOps.default$(Map.scala:288)
  at scala.collection.AbstractMap.default(Map.scala:420)
  at scala.collection.mutable.HashMap.apply(HashMap.scala:440)
  at analysis.RegionInjectorDSA.injectRegions$$anonfun$4$$anonfun$1(RegionInjector.scala:231)
  at scala.runtime.function.JProcedure1.apply(JProcedure1.java:15)
  at scala.runtime.function.JProcedure1.apply(JProcedure1.java:10)
  at scala.collection.IterableOnceOps.foreach(IterableOnce.scala:619)
  at scala.collection.IterableOnceOps.foreach$(IterableOnce.scala:617)
  at util.intrusive_list.IntrusiveList.foreach(IntrusiveList.scala:30)
  at analysis.RegionInjectorDSA.injectRegions$$anonfun$4(RegionInjector.scala:214)
  at scala.runtime.function.JProcedure1.apply(JProcedure1.java:15)
  at scala.runtime.function.JProcedure1.apply(JProcedure1.java:10)
  at scala.collection.IterableOnceOps.foreach(IterableOnce.scala:619)
  at scala.collection.IterableOnceOps.foreach$(IterableOnce.scala:617)
  at scala.collection.AbstractIterator.foreach(Iterator.scala:1303)
  at analysis.RegionInjectorDSA.injectRegions(RegionInjector.scala:213)
  at util.RunUtils$.staticAnalysis(RunUtils.scala:1063)
  at util.RunUtils$.$anonfun$20(RunUtils.scala:854)
  at scala.Option.map(Option.scala:242)
  at util.RunUtils$.loadAndTranslate(RunUtils.scala:853)
  at test_util.BASILTest.runBASIL(BASILTest.scala:56)
[...]

Issue #360, opened 2025-03-11
Non-determinism in stronglyConnectedComponents


l-kent:
stronglyConnectedComponents in IRCursor.scala is non-deterministic. 

This can be seen when running the procedure_summaries/procedure_summary3/gcc_O2:BAP test case in ProcedureSummaryTests.

stronglyConnectedComponents can produce any of the following on different runs of this test case:
- `ListBuffer(HashSet(g), HashSet(f), HashSet(main))`
- `ListBuffer(HashSet(g), HashSet(f))`
- `ListBuffer(HashSet(g))`
where `g`, `f` and `main` are the Procedures with those names.

In practice, this cause the test case to fail non-deterministically.

ailrst:
Has been fixed

l-kent:
If this has been fixed, shouldn't the procedure_summary3 test be re-enabled?

PR #408, opened 2025-04-22
mark booltobv1 function inline

ailrst:
Prevent z3 from finding an incorrect interpretation of booltobv1 function

### DIFF
--- a/src/main/scala/boogie/BExpr.scala
+++ b/src/main/scala/boogie/BExpr.scala
@@ -575,6 +575,7 @@ case class BInBounds(base: BExpr, len: BExpr, endian: Endian, i: BExpr) extends 
 }
 
 case class BoolToBV1Op(arg: BExpr) extends FunctionOp {
+  def attribute: BAttribute = BAttribute("inline", None)
   val fnName: String = "bool2bv1"
 }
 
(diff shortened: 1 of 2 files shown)


l-kent:
What was Z3 doing exactly?

ailrst:
Without inlining boogie emits an uninterpreted function and axiom, which relies on incomplete quantifier reasoning allowing z3 to pick an incorrect interpretation of the function if the axiom isn't triggered

```smt2
(declare-fun bool2bv1 (Bool) (_ BitVec 1))
(assert (forall ((arg Bool) ) (! (= (bool2bv1 arg) (ite arg #b1 #b0))
 :qid |basiloutbpl.55:29|
 :skolemid |1|
 :pattern ( (bool2bv1 arg))
)))
```

ailrst:
I don't have an easy repro example on hand unfortunately, but this is a sensible change to avoid incompleteness in the solver regardless. This test failure is a spurious nondeterministic failure https://github.com/UQ-PAC/BASIL/issues/370

l-kent:
That's an interesting issue, inlining makes sense to solve it. This should be fine except that one test is failing but I have no idea how this could have affected it? Is that a rare non-determinism issue, possibly related to #360?

ailrst:
Yeah this isn't related to the failing test its a periodic issue in the old dsa which has been troublesome to track down since its very hard to reproduce

ncough (approved):
Looks good to me. Failure is clearly unrelated.

PR #409, opened 2025-04-22
Function summaries

b-paul:
Introduces a new predicate type, along with abstract domains which can encode their states as predicates. This is used to create new precondition and postcondition generation methods. The three that have currently been implemented are:
- a reverse gamma domain, which generates requirements that branch conditions are low.
- a wp-like domain which approximates doing some sort of weakest precondition logic on the program.
- a combination of forward analyses to generate (interprocedurally) slightly path-sensitive postconditions reasoning about the gamma of output variables.


src/main/scala/util/RunUtils.scala
```diff
@@ -17,7 +17,7 @@ import scala.collection.mutable.ListBuffer
 import scala.collection.mutable.ArrayBuffer
 import scala.jdk.CollectionConverters.*
 import analysis.solvers.*
-import analysis.*
+import analysis.{Interval as _, *}
```
b-paul, line 20: There is now a DSA interval which clashed with the NumericalDomains Interval. This temp fix just removes the NumericalDomains Interval import, but perhaps something better should be done to fix this issue

src/main/scala/analysis/LatticeCollections.scala
```diff
@@ -13,9 +15,23 @@ trait InternalLattice[T <: InternalLattice[T]] {
...
+ *
+ * The `term` parameter can be any term of the type L, it just needs to exist to be able to call the top and bottom methods.
+ */
+class InternalLatticeLattice[L <: InternalLattice[L]](term: L) extends Lattice[L] {
+  def lub(x: L, y: L): L = x.join(y)
+  override def glb(x: L, y: L): L = x.meet(y)
+
+  val bottom: L = term.bottom
+  override def top: L = term.top
+}
```
ailrst, line 29: Factoring this out into a separate class you need to inherit from may be overkill, could just have `InternalLattice` extend `Lattice` directly

src/main/scala/analysis/LatticeCollections.scala
```diff
@@ -94,6 +110,14 @@ enum LatticeSet[T] extends InternalLattice[LatticeSet[T]] {
       case DiffSet(_) => None
     }
   }
+
+  /**
+   * Returns whether this set is Top or Top minus some elements
+   */
+  def topped: Boolean = this match {
```
ailrst, line 117: I don't think this is used anywhere?

ailrst (approved):
LGTM

ailrst:
Please squash when merging

PR #413, opened 2025-04-24
Condition propagation with reaching defs to check for clobbers

ailrst:
This transform takes every assume statement and copies the definitions into the expression, provided there is exactly one definition in the program (in DSA form). This enables the branch simplification to work on all branches in the cntlm-noduk example.

The `--validate-simplify` flag now additionally checks that this copy does not violate the DSA. This should not theoretically be possible as the dsa should ensure all definitions dominate uses and if we push a use forwards, as long as we aren't moving across a loop nesting (which requires there be multiple definitions involved) this should remain true of the target we are propagating to. 

Running this on the supported [csmith examples](https://github.com/UQ-PAC/basil-tests/tree/main/src/test/csmith) should still produce identical hashes with and without `--simplify`.


src/main/scala/ir/transforms/Simp.scala
```diff
@@ -529,12 +563,99 @@ def collectUses(p: Procedure): Map[Variable, Set[Command]] = {
...
+
+  def allDefinitions(p: Procedure) = {
+    p.collect { case a: Assign =>
+      a.assignees.map(l => l -> a)
+    }.flatten
+      .groupBy(_._1)
+      .map { case (v, ass) =>
+        v -> ass.map(_._2).toSet
+      }
+  }
```
katrinafyi, line 591: i cant read this. maybe the function can have a return type annotation

src/main/scala/ir/transforms/Simp.scala
```diff
@@ -529,12 +563,99 @@ def collectUses(p: Procedure): Map[Variable, Set[Command]] = {
...
+      def propOK(e: Expr) = {
+        // all dependent variables satisfy dsa property for the use we want to propage to
+        // DSA property: our reaching definitions are all definitions of the variable
+        e.variables.map(v => v -> reachingDefs(pos).get(v)).forall {
+          case (v, d) => {
+            (inparams.contains(v) && defs.get(v).forall(_.isEmpty))
+            || (d.isDefined && d.get.toSet == defs.get(v).toSet.flatten)
+          }
+        }
+      }
```
katrinafyi, line 631: the inner function can be moved outside the if statements?

src/main/scala/ir/transforms/Simp.scala
```diff
@@ -529,12 +563,99 @@ def collectUses(p: Procedure): Map[Variable, Set[Command]] = {
...
+    }
+  }
 
   override def vstmt(s: Statement) = s match {
-    case a @ Assume(_, b, c, d) =>
-      inlineCond(a) match {
-        case Some(cond) => ChangeTo(List(Assume(cond, b, c, d)))
+    case a @ Assume(body, b, c, d) if a.body.variables.exists(goodSubst) =>
+      val deps = mutable.Set[Variable]()
+      val depBlocks = mutable.Set[Assign]()
```
katrinafyi, line 654: unused?

src/main/scala/ir/transforms/Simp.scala
```diff
@@ -641,44 +775,51 @@ def coalesceBlocksCrossBranchDependency(p: Program): Boolean = {
...
+  var didAny = false
+
+  val blocks = proc.blocks.toList
+  for (b <- blocks.sortBy(_.rpoOrder)) {
+    if (
+      b.prevBlocks.size == 1 && b.prevBlocks.head.statements.nonEmpty && b.statements.nonEmpty
+      && b.prevBlocks.head.nextBlocks.size == 1
+      && b.prevBlocks.head.statements.lastOption.forall(s => !s.isInstanceOf[Call])
+      && !(b.parent.entryBlock.contains(b) || b.parent.returnBlock.contains(b))
+      && b.atomicSection.isEmpty && b.prevBlocks.forall(_.atomicSection.isEmpty)
```
katrinafyi, line 788: a lot of these conditions are repeated between the cases, so they could be moved into variables

katrinafyi (approved):
transform seems okay. comments optional.

PR #415, opened 2025-04-28
re-enable procedure_summary3

ailrst:
Working now with fixed SCC function,  this comment remains in the test definition so I'm not sure if its a valid test,

```
  // TODO currently procedure_summary3 verifies despite incorrect procedure summary analysis
  // this is due to BASIL's currently limited handling of non-returning calls
``


l-kent:
That no longer seems to be an issue either, due to the simplification pass's basic handling of non-returning calls and the recent overhaul of procedure summary generation. The test now reliably produces correct procedure summaries.

l-kent (approved):

PR #421, opened 2025-05-08
Update Mill to 12.10

l-kent:
This updates Mill to 12.10, including updating the various tests to use `System.getenv("MILL_WORKSPACE_ROOT")` when referencing the project root directory, which is now required. The tests have also been cleaned up a little bit.

z3process has also been updated to be platform independent


l-kent:
It seems that System.getenv("MILL_WORKSPACE_ROOT") isn't working on Linux in the github action - I'm not really clear on why yet as the Mill documentation doesn't say anything about it being platform specific.

katrinafyi:
I have a patch for check-test-tagging if you want it. This script has caused so much trouble ;-;

If mill has some options to reduce the fluff it outputs, that could help too.

```diff
From e542970b79522ca2f2e835e354e444c62721d56d Mon Sep 17 00:00:00 2001
From: rina <k@rina.fyi>
Date: Thu, 8 May 2025 15:42:25 +1000
Subject: [PATCH] check-test-tagging: update for mill 12 and add error message

---
 .github/workflows/check-test-tagging.sh | 7 +++++--
 1 file changed, 5 insertions(+), 2 deletions(-)

diff --git a/.github/workflows/check-test-tagging.sh b/.github/workflows/check-test-tagging.sh
index fb6d6f46f..9393a5762 100755
--- a/.github/workflows/check-test-tagging.sh
+++ b/.github/workflows/check-test-tagging.sh
@@ -5,7 +5,7 @@ set -o pipefail
 test_dir=src/test
 
 echo '::group::All test suites:'
-tests="$(./mill test.testOnly -- -t '' -oW | tr -d ':' | sort)"
+tests="$(./mill test.testOnly -- -t '' -oW | grep ':$' | tr -d ':' | cut -d' ' -f2 | sort)"
 echo "$tests"
 echo '::endgroup::'
 echo
@@ -20,7 +20,10 @@ find_test_case() {
   t="$1"
   shift
   # NOTE: \> matches the end of a word
-  exec grep 'class\s\+'"$t"'\>' --before-context=2 -R $test_dir "$@"
+  if ! grep 'class\s\+'"$t"'\>' --before-context=2 -R $test_dir "$@"; then
+    echo "find_test_case failed to find '$t'" >&2
+    return 1
+  fi
 }
 
 ok=true
-- 
2.48.1
```

l-kent:
Thanks, that made it display the error properly

katrinafyi:
It looks like mill only sets  MILL_WORKSPACE_ROOT when it's going through its own test command. The CI instead runs the tests through the scalatest.sh script which doesn't go through `./mill test. ...` and hence doesn't have the environment variable. 

I've made some changes to remove scalatest.sh and go straight through mill. This looks like it works now. https://github.com/UQ-PAC/BASIL/compare/mill-update..mill-update-scalatest-fix

Also, if you're looking at the green GH action for windows, it's misleading - for inscrutable powershell reasons, it's actually a no-op.

l-kent:
The changes you made to the Windows GitHub Action have made it so that it works now - why did we have that in there if it wasn't actually working before?

EvalTest does not work for me & also fails in UnitTestsWindows. This issue does not seem to be specific to the new version of Mill.

https://github.com/UQ-PAC/BASIL/blob/eebe343ea591dcc814595288e804e5c3023f36b4/src/main/scala/util/EvalScala.scala#L56-L62

The issue is that `manager.getEngineByName("scala")` fails, returning a null pointer, seemingly because it can't find what it needs on the classpath. I can find mention of this being a longstanding issue with this particular approach with both Mill and sbt, so I'm unsure why we haven't run into those - maybe they are particular to Windows. There are known workarounds for sbt but not Mill.

I tried an alternate approach to create the engine, which throws a slightly more helpful exception (instead of just a NullPointerException when engine.eval() is called)
```
val engine = manager.getEngineFactories.asScala.filter(f => f.getEngineName == "Scala REPL").head.getScriptEngine
```
produces:
```
dotty.tools.dotc.MissingCoreLibraryException: Could not find package scala from compiler core libraries.
Make sure the compiler core libraries are on the classpath. 
```

katrinafyi:
Re windows, I didn't notice the problem until a couple of days ago. One doesn't really anticipate the scenario where a run is green but doesn't run anything. After that, it was low priority to fix because we don't have windows computers and the test only served to catch one particular issue. In any case, it is a happy coincidence that this change also fixed the windows test.

Thanks for looking into the script engine issues. Just checking, when you say EvalTest doesn't work for you, is that on Windows as well?

l-kent:
Yes, I get the same problems as with UnitTestsWindows.

l-kent:
The question is just what do we want to do with EvalTest in UnitTestsWindows. It is going to keep failing unless someone finds a fix but this seems fairly unlikely. The code being tested is otherwise unused, but it's also clear that we can't rely on it for anything given this issue.

I would also like people to confirm that the Mill upgrade works on their machines.

katrinafyi:
The new mill seems to work on my computer. 

For EvalScalaTest, I'm happy to mark the tests as expected fail on Windows then wave this PR through.

l-kent:
I've had a substantial look at the underlying issue and haven't gotten anywhere, good luck if you decide to try to figure it out. How is it possible to mark a test as only failing on Windows?

katrinafyi:
Thanks :")

To mark a test as failing, you would use [TestCustomisation](https://github.com/UQ-PAC/BASIL/blob/main/docs/development/testing.md#testcustomisation-for-dynamically-generated-tests) combined with some way of detecting the OS.

katrinafyi:
Sorry about the -PS3 kerfuffle, I should've written about that in the commit because I ran into that problem too 😔

l-kent:
Mill 12.11 was released over the weekend and I've just had a look - seems like some things have been deprecated but they haven't yet updated the documentation to match so I'm not going to bother yet.

l-kent:
I think the best option short-term for test parallelism is bringing back scalatest.sh so I'll do that, it just needs some minor fixes. The issue with  `MILL_WORKSPACE_ROOT` should be fixed. scalatest.sh doesn't work on Windows but it never was working on Windows so I'm going to disable the Windows action for now.

Mill now has test parallelism options but none of them work well for us without further effort.

`def testParallelism = true` seems to be completely broken by our dynamic tests, as it spawns multiple redundant instances of test classes which all get in the way of each other.

`def testForkGrouping = discoveredTestClasses().grouped(1).toSeq` is much closer to what we want, as it runs each test class in a separate JVM in parallel (with more fine-grained control over the grouping being possible of course) but it has the parallelism problem where all the tests print out their results and progress simultaneously, it doesn't handle that nicely like the ScalaTestRunner does, sadly. There is probably a way to improve this but it would require more work.

l-kent:
Hmmm that doesn't seem to work, I have no idea why we get `mill-server/ exitCode file not found` causing everything to fail. We also now get the classloader issue with EvalTestScala even on Linux because we have to run the ScalaTest Runner through runMainLocal

katrinafyi:
Yeah idk. I think that using ./mill even with no parallelism might be the way to go to get this merged sooner.

The scalatest.sh script was always somewhat suspicious, and the test parallelism is just a nice-to-have. We can fix it in a later PR (but not too late, since slower tests are bothersome).

katrinafyi (approved):

PR #428, opened 2025-05-21
Split global intervals for dsa

sadrabt:
updates DSA flags to allow for specifying phases with --dsa (pre|local|bu|td)

Argument for performing extra checks now has it's own flag and can be enabled with --dsa-checks

Adds functionality to split the global region into many regions with predetermined bounds based on the symbol table (using --split-globals)
This has the effect of bounding a global collapse to a single symbol table variable instead of losing all object/field sensitivity for the whole global region.

For example, for the following C program
```
static int array[10] = {0, 1, 2, 3, 4, 5, 6, 7, 8, 9};
int x;
int *y;

int main(int r) {
    y = &x;
    return array[r];
}
```
the analysis would produce the following DSG for the procedure main:
image
As expected the global region is collapsed since it is accessed via statically indeterminable r (R0_in). This causes a collapse for all of the globals. In contrast with split enabled, the symbol table is used to find bounded global regions for x, y and array resulting in the following dsg:
image
As it can be seen the global region corresponding to the array is collapsed however, the regions corresponding to x and y are preserved (top corner). 

Further --dsa-assert can be used to update the program each time a memory operation accesses a bounded global with collapsed offset. The analysis will insert an assertion checking the value being dereferenced by the memory operation falls within the bounds of the global. For the above program, the following assert is inserted before the access to the array:
` assert (bvsge64(bvadd64(bvshl64(zero_extend2_62(sign_extend32_32(load0_1)[62:0]), 2bv64), 131120bv64), 131120bv64) && bvslt64(bvadd64(bvshl64(zero_extend2_62(sign_extend32_32(load0_1)[62:0]), 2bv64), 131120bv64), 131160bv64));
`
which requires the indeterminable offset (`bvadd64(bvshl64(zero_extend2_62(sign_extend32_32(load0_1)[62:0]), 2bv64)`) to fall within the range of the global region. 

The above program will only verify with a specification provided which bounds R0, eg:
```
Subroutine: main
Requires: 0bv64 <= R0
Requires: R0 < 10bv64
```

Adds functionality to create and maintain eq classes between cells of a node (for when two cells of the same node are verified) instead of collapsing the node (using --eq-classes)

Updates the relf reader to consider symbol table entries with LOCAL flag to be considered global variables (e.g variables declared static in the C source code)

The SVA now separates all constant values either into a global or constant based on symbol table information


ailrst:
Can you add an item in `dos/SUMMARY.md` for dsa page please so that it appears in the index?

src/main/scala/analysis/data_structure_analysis/SymbolicValueAnalysis.scala
```diff
@@ -295,53 +293,121 @@ object SymValues {
...
+    case UnaryExpr(op, arg) => exprToConstants(arg)
+    case AssocExpr(op, args) => args.flatMap(exprToConstants).toSet
+    case BinaryExpr(op, arg1, arg2) => exprToConstants(arg1) ++ exprToConstants(arg2)
+    case FApplyExpr(name, params, returnType, uninterpreted) => params.flatMap(exprToConstants).toSet
+    case variable: Variable => Set()
+    case LambdaExpr(binds, body) => exprToConstants(body)
+    case QuantifierExpr(kind, body, triggers) => exprToConstants(body)
+    case OldExpr(body) => exprToConstants(body)
+  }
+}
```
ailrst, line 405: Could possibly be implemented more simply with a visitor:


```scala
def getConstants(e: Expr) = {
    class GetConstants extends CILVisitor {
      var constants = List[Literal]()
  
      override def vexpr(e: Expr) = e match {
          case l: Literal => 
              contants = l :: contants
              SkipChildren()
          case _ => DoChildren()
    }
    val v = GetConstants()
    visit_expr(v, e)
    v.constants
}
```

docs/src/development/dsa.md
```diff
@@ -0,0 +1,166 @@
+
+# Data Structure Analysis
```
ailrst, line 2: Could possibly more clearly point to the recommended/ideal set of flags, but its fine as is.

docs/src/development/dsa.md
```diff
@@ -0,0 +1,166 @@
...
+which is points-to graph with additional information (see [[Lat05]](https://llvm.org/pubs/2005-05-04-LattnerPHDThesis.pdf) for more details).
+DSA is flow-insensitive and unification based. This allows for partial processing
+of memory constraints concerning a memory object which can later be completed
+as more information about the context (calling/global context) is discovered.
+This partial processing is done in phases.
+
+### Symbolic Value Analysis
+The pre-pass for dsa is made up of two components.
+1. `SymbolicValueAnalysis` which is an
+   intraprocedural value analysis where DSA (Dynamic Single Assign) variable corresponds to a value that is represented using a
```
ailrst, line 23: "where a DSA (Dynamic Single Assignment) variable..."

src/main/scala/Main.scala
```diff
@@ -223,12 +223,19 @@ object Main {
       doc = "Disable producing irreducible loops when --analyse is passed (does nothing without --analyse)"
     )
     noIrreducibleLoops: Flag,
+    @arg(name = "dsa", doc = "Perform Data Structure Analysis (requires --simplify flag) (pre|local|bu|td)")
+    dsaType: Option[String],
+    @arg(name = "dsa-checks", doc = "Perform additional dsa checks (requires --dsa (local|bu|td)")
+    dsaChecks: Flag,
+    @arg(name = "split-globals", doc = "split the globals for dsa (requires --dsa (pre|local|bu|td)")
```
ailrst, line 230: Probably a good idea to prefix `split-globals` and `eq-cells` with `dsa` as well, but I think we generally are going to need a more robust system for dependent flags at some point.

src/main/scala/ir/transforms/MemoryTransform.scala
```diff
@@ -15,59 +15,6 @@ class MemoryTransform(dsa: Map[Procedure, IntervalGraph], globals: Map[IntervalN
 //    dsa.map((proc, graph) => (proc, IntervalDSA.getPointers(graph)))
   val interProcCells: Map[IntervalCell, Set[IntervalCell]] = Map.empty // computeRelations()
```
ailrst, line 16: It looks like this value is unused entirely now?

ailrst: Was this the implementation that checked whether a region was only pointed to by once cell so could therefore be scalarised? I forget how that was resolved now

ailrst (approved):
The docs look good on this and its a nice improvement

I've commented on some superficial stuff, you can decide how to address those. I haven't really looked at the core implementation as the correctness of the approach is a bit beyond the scope of a pr review.

I don't have any major issues with merging this. 

l-kent If you still have review comments they can be addressed in a future issues/pr, we need to get this merged so Sadra can move on.

src/main/scala/ir/transforms/MemoryTransform.scala
sadrabt: yes but this is not necessary hence it's no longer being used just forgot to remove this

PR #429, opened 2025-05-22
parallelise tests using testForkGrouping and post-processing the output

katrinafyi:
this uses testForkGrouping and solves the problem of interespersed output by re-introducing scalatest.sh (i'm sorry)
and collating test suite output based on the `[001]` prefix in front of each line.

the output looks something like this. to make errors easier to find, any red text is re-printed after all the tests are run.
image

this pr changes the linux github actions to use the scalatest.sh. i don't have much confidence it'll work on windows.

**pros**:
- no AnalysisTest2,3,4
- allows configuration of parallelism level with the `-j` flag (though not exposed to command line)

**cons:**
- script probably wont work on windows
- probably wont work on macos either (uses GNU sort's --stable option)
- running `./mill test` without the script will have interspersed output

### DIFF
--- a/.github/workflows/run-examples.yml
+++ b/.github/workflows/run-examples.yml
@@ -79,7 +79,7 @@ jobs:
       - run: dotnet tool install --global boogie --version '3.4.3'
 
       - run: ./mill compile
-      - run: ./mill test.test -oGD -T 1000000 -n test_util.tags.StandardSystemTest
+      - run: ./scripts/scalatest.sh -oGD -T 1000000 -n test_util.tags.StandardSystemTest
 
       - uses: actions/upload-artifact@v4
         with:
@@ -152,7 +152,7 @@ jobs:
       - run: dotnet tool install --global boogie --version '3.4.3'
 
       - run: ./mill test.compile
-      - run: ./mill test.test -oGD -T 1000000 -n test_util.tags.UnitTest
+      - run: ./scripts/scalatest.sh -oGD -T 1000000 -n test_util.tags.UnitTest
 
         # every test with package prefix:
         # sbt "show test:definedTests"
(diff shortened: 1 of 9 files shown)


l-kent:
UnitTestsWindows should have `-j 1` so the output isn't mangled by all the tests outputting in parallel. I don't think that should meaningfully slow it down.

```
run: ./mill -j 1 test.test -oGD -T 1000000 -n test_util.tags.UnitTest
```
The documentation should be updated to note this issue too.

katrinafyi:
I'm withdrawing my support for this PR.

If anyone thinks it's a good idea and should be merged, speak now!

l-kent:
What's the issue?

katrinafyi:
I think it's not a good solution. The disadvantages are fairly intrusive and have the potential to cause confusion. 

Are you speaking in support of the PR?

l-kent:
I think it would be fine with the changes I suggested. The parallelism issues only exist if you run multiple test classes in one command (not a common part of my workflow, so it's only really an issue for the GitHub action), can easily be mitigated with `-j 1` which is equivalent to the status quo, with the only meaningful new downside being that the pass/fail numbers are now split per class and that's pretty minor.

katrinafyi:
Thanks for your comments, I'll touch up this PR. For documentation, do you mean a comment in the yml file or somewhere else as well?

In my workflow, I do usually run the unit test tag. I'll see if I can disable test parallelism by default unless some variable is set.

l-kent:
For documentation, I mean the [documentation](https://github.com/UQ-PAC/BASIL/blob/main/docs/development/testing.md) for running tests should address this issue and recommend using either `-j 1` or scala_test.sh (on Linux only) when running multiple test classes at once.

> I'll see if I can disable test parallelism by default unless some variable is set.

This would be an ideal solution if possible

katrinafyi:
Thanks! I'm mindful of the increased runtime though, I think testForkGrouping spawns a new JVM for every test so there's a fair bit of overhead. Here, the combined AnalysisSystemTests takes 10 minutes longer than it did before. I'm not super happy about this so I'll try a few more things.

l-kent:
It may be worth playing around with testForkGrouping to manually group the tests into only a few VMs to minimise the overhead? I'm not sure how to make that interact well with the tag filtering though

katrinafyi:
Yeah that'd be ideal but I think it's be hard to implement and maintain. If I had more time, I'd try to make the testParallelism option work, since it's basically what we had before.

Tweaking the number of forming groups doesn't seem to have had an effect, but this can be merged for now if you're okay with it.

PR #430, opened 2025-05-23
assume&prove registers are preserved across calls

ailrst:
--simplify now adds an assertion requiring
        1. callee-saved registers r19-r28 and FP (R29) are preserved
           (required by ABI)
        2. Return-address, LR (R30) are preserved
           (assuming well-behaved stack and return structure)
and removes the corresponding registers from the out-param list.


src/main/scala/ir/transforms/CalleePreservedParam.scala
```diff
@@ -0,0 +1,62 @@
...
+
+  val preserved = (19 to 29).map(i => s"R$i").toSet
+
+  def isPreservedParam(v: LocalVar) = {
+    // abi assumption
+    v.varName match {
+      case "R29_in" | "R30_in" | "R31_in" | "R29" | "R30" | "R31" | "R29_out" | "R30_out" | "R31_out" => true
+      case o if preserved.contains(o.stripSuffix("_in").stripSuffix("_out")) => true
+      case _ => false
+    }
```
katrinafyi, line 20:
```suggestion
    v.varName match {
      case "R29_in" | "R30_in" | "R31_in" | "R29" | "R30" | "R31" | "R29_out" | "R30_out" | "R31_out" => true
      case s"${reg}_in" => preserved.contains(reg)
      case s"${reg}_out" => preserved.contains(reg)
      case reg => preserved.contains(reg)
    }
```

we can use format strings as a pattern to avoid the stripSuffix chains.

*also*, why is there the distinction between some being in the preserved set and some being specified as literals?

src/main/scala/ir/transforms/CalleePreservedParam.scala
```diff
@@ -0,0 +1,62 @@
...
+    assert(invariant.correctCalls(p))
+  }
+
+  class MakePreserved extends CILVisitor {
+
+    override def vproc(p: Procedure) = {
+
+      for (param <- p.formalOutParam.filter(isPreservedParam)) {
+
+        val input = p.formalInParam.find(_.name.startsWith(param.varName.stripSuffix("_out"))).get
```
katrinafyi, line 35: i can't read this. what is it doing?

also, i'm suspicious of using startsWith. for instance, R10 starts with R1, but this is obviously not what you're looking for here (i know that, in this case, all the register numbers are 2 digits, but it still makes me uneasy).

src/main/scala/ir/transforms/CalleePreservedParam.scala
```diff
@@ -0,0 +1,62 @@
+package ir
+package transforms
+
+import cilvisitor.*
+
+object CalleePreservedParam {
+
+  /**
+   * Asusming single-return and parameter form 
+   */
```
katrinafyi, line 10: would it be useful to assert that the procedures are in this form?

src/main/scala/ir/transforms/CalleePreservedParam.scala
```diff
@@ -0,0 +1,62 @@
...
+    }
+  }
+
+  def transform(p: Program) = {
+    val v = MakePreserved()
+    visit_prog(v, p)
+    assert(invariant.correctCalls(p))
+  }
+
+  class MakePreserved extends CILVisitor {
```
katrinafyi, line 29: optionally, this is just a loop over all procedures - it doesn't need to be a visitor.

src/main/scala/ir/transforms/CalleePreservedParam.scala
```diff
@@ -0,0 +1,62 @@
...
+    override def vproc(p: Procedure) = {
+
+      for (param <- p.formalOutParam.filter(isPreservedParam)) {
+
+        val input = p.formalInParam.find(_.name.startsWith(param.varName.stripSuffix("_out"))).get
+
+        p.formalOutParam.remove(param)
+
+        for (call <- p.incomingCalls()) {
+          call.outParams = call.outParams.removed(param)
```
katrinafyi, line 40: i like using `-`, but totally optional

katrinafyi (approved):

PR #431, opened 2025-05-23
add Mill modules for BNFC compilation

katrinafyi:
For my next task, I'll look at implementing parsing of the IR format produced by Basil (e.g., in the .il output files). This grammar is defined as a BNFC grammar https://bnfc.digitalgrammars.com/, and this needs to be processed through the bnfc tool and the usual Antlr tool. See: https://bnfc.readthedocs.io/en/latest/user_guide.html#antlrv4

This PR lays the groundwork to fetch a BNFC grammar from a URL, fetch `bnfc` from the official releases, and process it all into a Java module which can be imported into the main Basil module with `def moduleDeps = Seq(bnfc)`

I'm sending this PR out now because it's a pretty complex change to the build system, and it would be good to get some eyes on it now instead of being bundled with a much bigger PR.

bonuses:
- add gitattributes to mark mill files as scala.
- removed unused mill-antlr import from ivy
- please squash these changes, i've excised these changes from the rest of the ongoing work.


ailrst:
Now I think about it, it would probably make more sense to check-in the grammar here and curl it in the language server build

katrinafyi:
I agree....

my hard work 😭

katrinafyi:
It can work without internet by trying an external local copy first. bnfc searches PATH and the grammar has its own environment variable. These should be simple enough to set up for Nix or a VM image.

ailrst (approved):

PR #432, opened 2025-05-26
Revert "runtime evaluation of Scala code (#404)"

katrinafyi:
When trying to add a command-line flag for the EvalScala feature, I found that it does not function at all outside of the unit tests (the same MissingCoreLibraryException as seen on Windows).

I tried to fix this using some Mill tweaks but I was unsuccessful. Ultimately, it's not worth spending more time fixing what is a fundamentally terrible piece of code.

The goal of EvalScala was to allow for parsing Basil IR as input. I'm working on another approach to implement this feature.

This reverts commit 171754af2904ebf375b9cfc04da52d283fe3008d.

Conflicts:
build.mill
src/test/scala/EvalScalaTest.scala


ailrst (approved):
pending CI

PR #443, opened 2025-06-03
Update readme2

ailrst:
Basically just deletes redundant repetition in the docs so that everything is in `usage.md` and we mainly recommend the nix path.


readme.md
```diff
@@ -76,27 +78,31 @@ $ ./mill run --load-directory-gtirb src/test/incorrect/basicassign/gcc --verify
...
 ```
-./mill run --load-directory-gtirb example.gts [--output output.bpl] [--simplify] [--interpret]
+./mill run --load-directory-gtirb ./example.gts [--output output.bpl] [--simplify] [--interpret]
 ```
 
 Windows: 
 
 ```
-./mill.bat run --load-directory-gtirb example.gts [--output output.bpl] [--simplify] [--interpret]
+./mill.bat run --load-directory-gtirb ./example.gts [--output output.bpl] [--simplify] [--interpret]
```
katrinafyi, line 103: not in this pr, but we should fix the requirement for ./

readme.md
```diff
@@ -76,27 +78,31 @@ $ ./mill run --load-directory-gtirb src/test/incorrect/basicassign/gcc --verify
 
 ### Usage
 
-BASIL requires a `.gts` file produced by [gtirb-semantics](https://github.com/UQ-PAC/gtirb-semantics), 
+The `./mill` script should be sufficient to bootstrap Scala, provided you have a JVM (>17) installed.
```
katrinafyi, line 81: should be >= 17

docs/usage.md
```diff
@@ -5,6 +5,24 @@ from the perspective of a user of the tool.
 It may also be useful for BASIL developers as a reference for how to lift
 new binaries for manual testing.
 
+## Running Basil
+
+Basil needs a JVM(>17), recommended JDK-21, which can be installed using system packages.
```
katrinafyi, line 10: >= 17

docs/usage.md
```diff
@@ -5,6 +5,24 @@ from the perspective of a user of the tool.
...
+
+- e.g. `sudo apt-get install openjdk-21-jdk`
+
+On first run `./mill` (on linux/windows) or `./mill.bat` (on Windows) should bootstrap Scala automatically. Otherwise it can be set up through [coursier](https://get-coursier.io/docs/cli-installation).
+
+### Test Dependencies
+
+Both unit tests and system tests will require a working installation of the Boogie verifier backend and Z3 SMT solver.
+
+These can be installed through [nix](https://github.com/katrinafyi/pac-nix?tab=readme-ov-file#first-time) (recommended):
```
katrinafyi, line 20: I always thought z3 and Boogie were fine to install by system packages or dotnet. In fact, i think the nix Boogie package is outdated and considered insecure.

docs/usage.md
```diff
@@ -5,6 +5,24 @@ from the perspective of a user of the tool.
...
+
+- `nix profile install nixpkgs#z3 nixpkgs#boogie`
+
+Alternatively, Z3 is usually available in system packages. For manual Boogie install instructions see subheading [Verifying the Boogie file](#verifying-the-boogie-file).
+
 ## Lifting a single binary
```
katrinafyi, line 26: it could be worth adding a breadcrumb link to the docker littering docs, because it's useful for basil Devs to know the difference

docs/usage.md
ailrst: I just want to avoid dealing with and supporting homebrew/choco/dotnet/whatever, environment variable garbage and recommending a specific version. but that would require nix to actually work across platforms anyway.

ailrst: I would rather we could just say 'use this devenv' and if you're on windows...don't. But I don't know if thats feasible.

readme.md
ailrst: This is intentional because if we ever break 17 I'm going to fix it by just increasing this version requirement.

katrinafyi: I'd like to support 17 as long as it's not prohibitively difficult. in any case, it should be consistent and other files say >=17

docs/usage.md
katrinafyi: idk I never like recommending installing packages from nixpkgs# because it's updated so frequently. so you never know what version someone has and you end up using more disk space because every nixpkgs# install might add its own glibc.

also with boogie, since it's insecure it'd need NIXPKGS_ALLOW_INSECURE=1 or something which is :/

realistically, it shouldn't matter where z3 and boogie come from (unlike a C compiler which needs special config). surely students have installed a package before and they should know how to fix PATH if needed, and if not it's not too hard to learn.

PR #447, opened 2025-06-05
ci: use scalatest runner directly (again)

katrinafyi:
this brings back the old scalatest.sh which directly invokes scalatest's runner. now that there's a case fallback if MILL_WORKSPACE_ROOT is unset, this seems to work fine and leads to the best behaviour:

tests are run all in one thread, allowing for the best runtime performance and correctly-sorted output.

for reference:
- commit before mill 12.10 and scalatest.sh removal https://github.com/UQ-PAC/BASIL/commit/55587f08f4f90ecd35e86c14c8729669f67ca0b3
- my unfortunate suggestion to use `./mill test` directly https://github.com/UQ-PAC/BASIL/pull/421#issuecomment-2861915997
- hacky band-aid by re-sorting output. this worked but it was slow. https://github.com/UQ-PAC/BASIL/pull/429

i'm sorry for the trouble lol

### DIFF
--- a/.github/workflows/run-examples.yml
+++ b/.github/workflows/run-examples.yml
@@ -173,7 +173,7 @@ jobs:
           AnalysisSystemTest3,
           AnalysisSystemTest4,
         ]
-    fail-fast: false
+      fail-fast: false
 
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/run-examples.yml
+++ b/.github/workflows/run-examples.yml
@@ -173,6 +173,7 @@ jobs:
           AnalysisSystemTest3,
           AnalysisSystemTest4,
         ]
+    fail-fast: false
 
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/run-examples.yml
+++ b/.github/workflows/run-examples.yml
@@ -163,10 +163,16 @@ jobs:
   AnalysisSystemTests:
     runs-on: ubuntu-latest
 
-
+    # run 4 AnalysisSystemTests in 4 parallel github jobs.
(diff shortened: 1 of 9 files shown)


katrinafyi:
ug this is silly. of course using separate github actions jobs makes it faster because we get more cpu. maybe that's just the way to go

l-kent:
This seems like an improvement. Even if you think it makes sense to use separate GitHub actions as well (which I'm fine with if you want to add that back), we still want to incorporate this.

I did try this approach when we were working on updating Mill but ran into problems with `mill-server/ exitCode file not found` errors. Is `--ticker false` what fixes that?

katrinafyi:
I've seen that issue too, randomly. But I think it's just random or a caching problem. I don't think that argument would've fixed it.

l-kent:
Hmm, strange. It happened in the GitHub action when I tried it, so I don't think it's a caching issue. I'd prefer not to introduce another way for tests to fail arbitrarily but it doesn't seem to be happening right now at least?

katrinafyi:
It seems that the exitCode issue was caused by using runMainLocal instead of runMain. I can reproduce it locally if using runMainLocal. It seems to work fine with runMain. Was there a reason you tried with runMainLocal?

Mill's help has this to say about them:
```
runMainLocal(RunModule.scala:126)
    Same as `runLocal`, but lets you specify a main class to run

runLocal(RunModule.scala:101)
    Runs this module's code in-process within an isolated classloader. This is
    faster than `run`, but in exchange you have less isolation between runs
    since the code can dirty the parent Mill process and potentially leave it
    in a bad state.
```
So it sounds like this is a risk of using the local method. Maybe the scalatest runner does not play nice.

PR #448, opened 2025-06-05
Make --relf optional

ailrst:
Make --relf optional for tasks not requiring symbols. The symbol table is only really required for the spec input and DSA heuristics and the relf parser tends to be by far the most fragile part of the frontend.

When relf is not provided look for `--main-procedure-name` (default 'main') in the procedure list to set as Program.mainProcedure. This is previously all we do to detect the entrypoint, except with the relf symbol table to find a procedure address to search for.

I think eventually we should be able to retrieve all the same information from the gtirb `Module.symbols` but that's a bigger task.

### DIFF
--- a/src/main/scala/Main.scala
+++ b/src/main/scala/Main.scala
@@ -96,11 +96,11 @@ object Main {
     bapInputDirName: Option[String],
     @arg(name = "load-directory-gtirb", doc = "Load relf and gts from directory (and spec from parent directory)")
     gtirbInputDirName: Option[String],
-    @arg(name = "input", short = 'i', doc = "BAP .adt file or GTIRB/ASLi .gts file")
+    @arg(name = "input", short = 'i', doc = "BAP .adt file or GTIRB/ASLi .gts file (requires --relf)")
     inputFileName: Option[String],
     @arg(name = "relf", short = 'r', doc = "Name of the file containing the output of 'readelf -s -r -W'.")
     relfFileName: Option[String],
-    @arg(name = "spec", short = 's', doc = "BASIL specification file.")
+    @arg(name = "spec", short = 's', doc = "BASIL specification file (requires --relf).")
     specFileName: Option[String],
     @arg(name = "output", short = 'o', doc = "Boogie output destination file.")
     outFileName: String = "basil-out.bpl",
--- a/src/main/scala/Main.scala
+++ b/src/main/scala/Main.scala
@@ -96,9 +96,13 @@ object Main {
     bapInputDirName: Option[String],
     @arg(name = "load-directory-gtirb", doc = "Load relf and gts from directory (and spec from parent directory)")
     gtirbInputDirName: Option[String],
-    @arg(name = "input", short = 'i', doc = "BAP .adt file or GTIRB/ASLi .gts file (requires --relf)")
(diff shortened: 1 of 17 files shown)


src/main/scala/Main.scala
```diff
@@ -92,11 +96,11 @@ object Main {
     bapInputDirName: Option[String],
     @arg(name = "load-directory-gtirb", doc = "Load relf and gts from directory (and spec from parent directory)")
     gtirbInputDirName: Option[String],
-    @arg(name = "input", short = 'i', doc = "BAP .adt file or GTIRB/ASLi .gts file")
+    @arg(name = "input", short = 'i', doc = "BAP .adt file or GTIRB/ASLi .gts file (requires --relf)")
```
l-kent, line 99: If `--relf` is now optional, why have you changed the comment to say it's required when used alongside `--input`

src/main/scala/util/RunUtils.scala
```diff
@@ -119,24 +119,32 @@ object IRLoading {
...
+          loadGTIRB(q.inputFile, Some(mainAddress), Some(q.mainProcedureName))
+        } else {
+          throw Exception(s"input file name ${q.inputFile} must be an .adt or .gts file")
+        }
+      specification = IRLoading.loadSpecification(q.specFile, program, globals)
+    } yield (IRContext(symbols, externalFunctions, globals, funcEntries, globalOffsets, specification, program))
+
+    val ctx = withRelf.getOrElse {
+      val prog = loadGTIRB(q.inputFile, None, Some(q.mainProcedureName))
+      IRLoading.load(prog)
```
l-kent, line 141: This seems like a particularly convoluted approach that obscures the logic. If `--relf` isn't provided, it also assumes that a `.gts` file was provided as input, which could cause confusion. If you don't want to make `BAPToIR` support returning a main procedure based on the provided name then you should at least throw an exception.

l-kent:
The symbol data is used more widely than you say it is. For instance, it's used to identify which sections of memory to include in the `.bpl` output, which is important enough that we should not make the `.relf` files optional at the moment. 

If there are issues with the parser being fragile then we should address that.

`--dsa` and `--analyse` (and potentially some of the other flags, I'm not sure) would also need to state that they require `--relf` if it were to be made optional.

ailrst:
> If there are issues with the parser being fragile then we should address that.

I don't think maintaining the relf parser is a good use of our (very limited) time and resources. Its an unstable, unspecified and human readable format  with very little structure, there is going to be an extremely long tail of edge cases (e.g. stupid stuff like heading pluralisation) to support.

ailrst:
I'll cleanup the logic and help text surrounding ony supporting gtirb without relf

l-kent:
> I don't think maintaining the relf parser is a good use of our (very limited) time and resources. Its an unstable, unspecified and human readable format with very little structure, there is going to be an extremely long tail of edge cases (e.g. stupid stuff like heading pluralisation) to support.

This is a reason to work towards replacing it with solely relying on the extra information from GTIRB, not to say that it's optional while it is still being actively relied upon.

ailrst:
I'm not _removing_ the flag? There are paths that don't need it so its not _required_ for every input to basil.

l-kent:
> The symbol data is used more widely than you say it is. For instance, it's used to identify which sections of memory to include in the .bpl output, which is important enough that we should not make the .relf files optional at the moment.

PR #450, opened 2025-06-11
use hashmap to improve dsl loading perf by 2000%

ailrst:
Caches the `label -> block` and `name -> procedure` mapping to avoid linear search.

The tests just reconstruct this cache every time as they usually modify the procedure and incrementally resolve.

### DIFF
--- a/src/main/scala/ir/dsl/DSL.scala
+++ b/src/main/scala/ir/dsl/DSL.scala
@@ -92,23 +92,25 @@ def R(i: Int): Register = Register(s"R$i", 64)
 
 def bv_t(i: Int) = BitVecType(i)
 
-case class DelayNameResolve(ident: String) {
-  def resolveProc(prog: Program): Option[Procedure] = prog.collectFirst {
-    case b: Procedure if b.name == ident => b
-  }
+case class Resolver(program: Program) {
+  lazy val procs = program.procedures.map(p => p.name -> p).toMap
+  lazy val blocks = program.procedures.map(p => p.name -> (p.blocks.map(b => b.label -> b)).toMap).toMap
+}
 
-  def resolveBlock(prog: Program): Option[Block] = prog.collectFirst {
-    case b: Block if b.label == ident => b
-  }
+case class DelayNameResolve(ident: String) {
+  def resolveProc(prog: Resolver): Option[Procedure] = prog.procs.get(ident)
+
+  def resolveBlock(resolver: Resolver, parent: String): Option[Block] =
+    resolver.blocks.get(parent).flatMap(_.get(ident))
 }
 
 sealed trait EventuallyStatement extends DeepEquality {
-  def resolve(p: Program): Statement
+  def resolve(p: Resolver): Statement
   def cloneable = this
 }
 
 case class CloneableStatement(s: NonCallStatement) extends EventuallyStatement {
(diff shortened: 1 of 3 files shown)


src/main/scala/ir/dsl/DSL.scala
```diff
@@ -92,31 +92,33 @@ def R(i: Int): Register = Register(s"R$i", 64)
 
 def bv_t(i: Int) = BitVecType(i)
 
+case class Resolver(program: Program) {
+  lazy val procs = program.procedures.map(p => p.name -> p).toMap
+  lazy val blocks = program.procedures.map(p => p.name -> (p.blocks.map(b => b.label -> b)).toMap).toMap
```
katrinafyi, line 97: how does one deal with invalidating the lazy values? is that something we have to worry about?

ailrst: Construct a new Resolver() to pass to resolve. I don't think we need to emphasize the use case of incrementally modifying the EventuallyX and resolving repeatedly. In that case you can call `.resolve(p: Program ...)` instead and it just computes the hashmap every time, approximately as slow as linear search everytime. There's not much reason for these fields to be lazy here.

katrinafyi: Yeah I also think it doesn't need to be lazy. Optionally, I think it would be nice to make the vals private and only expose `def procs(name: String): Option[Procedure]`, but optional.

src/main/scala/ir/dsl/DSL.scala
```diff
@@ -164,21 +167,21 @@ case class EventuallyCall(
 case class EventuallyGoto(targets: Iterable[DelayNameResolve], label: Option[String] = None)
     extends EventuallyJump
     with DefaultDeepEquality {
-  override def resolve(p: Program, proc: Procedure): GoTo = {
-    val tgs = targets.flatMap(tn => tn.resolveBlock(p))
+  override def resolve(p: Resolver, proc: String): GoTo = {
+    val tgs = targets.flatMap(tn => tn.resolveBlock(p, proc))
     GoTo(tgs, label)
```
katrinafyi, line 172: i know this isn't new in this PR, but a flatMap will silently discard targets which don't resolve successfully. is this a good idea?

src/main/scala/ir/dsl/DSL.scala
```diff
@@ -92,31 +92,33 @@ def R(i: Int): Register = Register(s"R$i", 64)
 
 def bv_t(i: Int) = BitVecType(i)
 
+case class Resolver(program: Program) {
```
katrinafyi, line 95: i wonder if it would be good to make this private. otherwise, add a comment warning the user how to use it correctly wrt making a new Resolver when blocks/procs are changed.

_also,_ i'm unsure of the Resolver name, since we already use resolver for a different thing in this file (the lambda continuation in the `resolve` methods). but the best i can come up with is CachedProcedureAndBlocks which is.. not great

_also_, i think this bookkeeping should eventually be moved into Program/Procedure. i'm sure it's useful elsewhere as well.

katrinafyi (approved):

src/main/scala/ir/dsl/DSL.scala
ailrst: Yeah I should fix this

PR #452, opened 2025-06-12
docs: fix cross-linking to bnfc javadocs

katrinafyi:
previously, in pages like
https://uq-pac.github.io/BASIL/api/basil/ir/parsing/BasilEarlyBNFCVisitor.html,
the links to bnfc-generated classes like Visitor would not work. this fixes those links.

old:
image

new:
image


ailrst:
Very cool

PR #455, opened 2025-06-13
fix --load-directory when used with filenames with multiple dots

katrinafyi:
assuming you had files atomic.-O0.gts  atomic.-O0.relf  atomic.-O0.spec in a directory, previously, this would not work:

    ./mill run --load-directory-gtirb atomic/atomic/atomic.-O0.gts

the logic would incorrectly replace /two/ levels of file extension and look for files named atomic.gts and atomic.relf.

this fixes the logic to only strip off one file extension.

a test case is this zip from Aaron:
https://github.com/user-attachments/files/20717350/atomic.zip

### DIFF
--- a/src/main/scala/Main.scala
+++ b/src/main/scala/Main.scala
@@ -23,6 +23,9 @@ object Main {
     case Bap
   }
 
+  /** Replaces the extension of the given path with the given new extension, if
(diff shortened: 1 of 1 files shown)


ailrst:
LGTM

Issue #454, opened 2025-06-12
Exception: encountered TempIf not at end of statement list


aaronbembenek:
Hi there,

I tried to run the attached program with Basil (commit d9f60a08) and ran into the following exception:

```
[229] Exception in thread "main" java.lang.Exception: encountered TempIf not at end of statement list: List(AtomicStart 4196252_0: , 4196252_1$0: LocalVar(load3, 0, bv32) := MemoryLoad(Memory(mem, 64, 8), Register(R3, bv64), LittleEndian, 32), 4196252_1$1: LocalVar(Exp16__6_2_10, 0, bv32) := LocalVar(load3, 0, bv32), TempIf((LocalVar(Exp16__6_2_10, 0, bv32) == Register(R0, bv64)[32:0]), List(4196252_2: Memory(mem, 64, 8)[Register(R3, bv64)] := MemoryStore(Register(R4, bv64)[32:0], LittleEndian, 32)), List()), AtomicEnd 4196252_3: , 4196252_4: Register(R0, bv64) := ZeroExtend(32, LocalVar(Exp16__6_2_10, 0, bv32)))
[229]   at translating.GTIRBToIR.$anonfun$8(GTIRBToIR.scala:259)
[229]   at scala.collection.immutable.List.exists(List.scala:396)
[229]   at translating.GTIRBToIR.insertPCIncrement(GTIRBToIR.scala:257)
[229]   at translating.GTIRBToIR.createIR$$anonfun$4$$anonfun$1$$anonfun$2(GTIRBToIR.scala:195)
[229]   at scala.collection.IterableOnceOps.foreach(IterableOnce.scala:619)
[229]   at scala.collection.IterableOnceOps.foreach$(IterableOnce.scala:617)
[229]   at scala.collection.AbstractIterable.foreach(Iterable.scala:935)
[229]   at scala.collection.IterableOps$WithFilter.foreach(Iterable.scala:905)
[229]   at translating.GTIRBToIR.createIR$$anonfun$4$$anonfun$1(GTIRBToIR.scala:194)
[229]   at scala.collection.immutable.HashSet.foreach(HashSet.scala:951)
[229]   at translating.GTIRBToIR.createIR$$anonfun$4(GTIRBToIR.scala:189)
[229]   at scala.runtime.function.JProcedure1.apply(JProcedure1.java:15)
[229]   at scala.runtime.function.JProcedure1.apply(JProcedure1.java:10)
[...]

PR #457, opened 2025-06-13
allow TempIf in the middle of blocks, given unique branching behaviour

katrinafyi:
allows TempIf to occur in the middle of blocks, but only if both sides of the tempif branch or both sides do not branch.

this allows handling of, for example, `casal w0, w4, [x3]` 0x88e0fc64 which looks like this:
```
AtomicStart.0 {{  }} (  ) ;
constant bits ( 32 ) Exp16__6 = Mem.read.0 {{ 4 }} ( __array _R [ 3 ],4,9 ) ;
if eq_bits.0 {{ 32 }} ( Exp16__6,__array _R [ 0 ] [ 0 +: 32 ] ) then {
  Mem.set.0 {{ 4 }} ( __array _R [ 3 ],4,9,__array _R [ 4 ] [ 0 +: 32 ] ) ;
}
AtomicEnd.0 {{  }} (  ) ;
__array _R [ 0 ] = ZeroExtend.0 {{ 32,64 }} ( Exp16__6,64 ) ;
```

i wish i could write a unit test for this, but alas...

Fixes: https://github.com/UQ-PAC/BASIL/issues/454

### DIFF
--- a/src/main/scala/translating/GTIRBToIR.scala
+++ b/src/main/scala/translating/GTIRBToIR.scala
@@ -248,28 +248,39 @@ class GTIRBToIR(
     Program(procedures, intialProc, initialMemory)
   }
 
+  /** Determines if all paths through the given isnStmts have the same branching behaviour.
+   *  Returns true if all paths branch, or false if all paths do not branch.
+   *  Throws an error if some paths branch but some do not.
+   */
+  private def findUniqueBranchBehaviour(isnStmts: immutable.Seq[Statement]): Boolean =
+    isnStmts.exists {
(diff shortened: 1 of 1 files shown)


katrinafyi:
Yeah I agree. If I just add it to src/test for now, is it a lot of trouble for you to add this to basil-tests?

ailrst:
I want to avoid increasing basil repo size in the meantime let me add it directly to a new repo

ailrst:
Otherwise this PR LGTM, semantics looks right for this example

x

PR #459, opened 2025-06-16
Ctags setup to enable go-to-definition and listing symbols

katrinafyi:
[Ctags](https://en.wikipedia.org/wiki/Ctags) is an ancient piece of software that uses regular expressions to implement go-to-definition for virtually any programming language. This can be somewhat unreliable, but it is universally-available and very robust. This is very useful for me, personally, because I don't use the scala LSP.

This PR adds `./mill ctags` to generate the ctags file in the root of the repository. Once generated, this can be read by any editor which supports ctags.

I've written more in [editor-setup](https://github.com/UQ-PAC/BASIL/blob/5e8d1bb97fb4ff0211a2ed061a5c6e31bb3f3c61/docs/development/editor-setup.md#ctags), and there is also an example setup. Using it looks like this:
image


ailrst (approved):
:100: thank you for giving back my ram and swap

PR #469, opened 2025-06-19
Offline lifter integration

ailrst:
This does the minimum integration of the offline aslp lifter for gtirb input.

- Add `--lifter` flag for gtirb input which uses the offline, compiled scala lifter in place of the json serialised semantics 
- Add support for loading directly from `.gtirb` file with magic bytes, without semantics attached (requiring `--lifter`)

The `--lifter` flag imposes practically no performance overhead, it differs in load time for cntlm-noduk by 1 second

The IR construction interface to the offline lifter is mostly copied from my earlier prototypes and I haven't tested it too closely. I would expect it to still have bugs, e.g. possibly emitting ill-typed code. Not sure where we want to get it to for this initial PR, maybe at least getting SystemTests to pass with the `--lifter` flag. 

Architecture:

- The lifter itself is a pre-compiled Scala `.jar` file pulled from the aslp releases page by mill
- `InstructionBuilder.scala` implements the interface the lifter calls to construct the IR into basil IR + TempIf
- The rest of the gtirb backend is left mostly alone.

Motivation

- I ran into a difficult to address heap exhaustion problem with the json strategy for a ~10M binary. This is far larger than we can realistically handle in the end, but the offline lifter neatly avoids those kinds of problems by efficiently lifting on-demand.
- The built-in lifter lets us avoid running a separate tool (gtirb-semantics) from the basil side, so we no longer directly require an ocaml toolchain to be installed.
- We statically know how what IR it is possible for the lifter to emit (e.g. the set of intrinsics) through what we define in the interface


ailrst:
This now passes SystemTests but the types of the floating point intrinsics are definitely going to be wrong with new files. It would be helpful to have a basil-side type checker for this so we aren't relying on boogie.

ailrst:
There is an issue with the interaction between `--lifter` and `--pc assert`, e.g. 

```
$  ./mill run -i examples/cntlm-noduk/cntlm-noduk.gts --lifter --pc assert --main-procedure-name acl_add --simplify --trim-early --dump-il pc  --verify
[257] [INFO]   Boogie result: Unknown(basil-out.bpl(5324,23): Error: undeclared identifier: #R30_in
[257]          basil-out.bpl(21225,23): Error: undeclared identifier: #R30_in
[257]          basil-out.bpl(29177,23): Error: undeclared identifier: #R30_in
[257]          basil-out.bpl(60998,23): Error: undeclared identifier: #R30_in
[257]          basil-out.bpl(74491,23): Error: undeclared identifier: #R30_in
[257]          basil-out.bpl(95699,23): Error: undeclared identifier: #R30_in
[257]          6 name resolution errors detected in basil-out.bpl
[257]          )
```

ailrst:
> maybe GTIRBToIR can take a (blockUUID: ByteString, blockCountIn: Int, blockAddress: Option[BigInt]) => Seq[Seq[Statement]] as an argument

This seems difficult as it closes over all the gtirb data maps in gtirbloader but maybe they can be also be factored out

katrinafyi:
Yeah, but imo that function should be made entirely gtirb independent because it just converts aslp to basil. Control flow is handled later by the gtirb CFG.

ailrst:
You have to get the addr + opcodes out of the block... the instruction lifter level stuff is gtirb-independent.

katrinafyi:
You would need to change the functions type signature, but you can definitely pull out the opcode data before padsing it to the function.

ailrst:
There's no such thing as a pure function here

katrinafyi (approved):

PR #474, opened 2025-06-22
Mill command to run profiler

ailrst:
This enables downloading and invoking basil with async-profiler through mill.

It should support Linux and mac, though I don't have a mac to test on.

```
./mill runProfile --profileDest profile.html -i examples/cntlm-noduk/cntlm-noduk.gts
```

### DIFF
--- a/basilmill/util.mill
+++ b/basilmill/util.mill
@@ -18,6 +18,12 @@ trait DownloadModule extends Module {
   def name: String
   def url: String
   def local: Target[Either[String, String]]
+  def unzip: Option[Unzip] = None
+
+  sealed trait Unzip
+  case class TarGZ(extractedName: String) extends Unzip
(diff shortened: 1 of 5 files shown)


basilmill/util.mill
```diff
@@ -18,6 +18,32 @@ trait DownloadModule extends Module {
   def name: String
   def url: String
   def local: Target[Either[String, String]]
+  def unzip: Option[Unzip] = None
+
+  sealed trait Unzip
+  case class TarGZ(extractedName: String) extends Unzip
+  case class Zip(extractedName: String) extends Unzip
```
katrinafyi, line 25: fileToExtract instead of extractedName? extracted name makes it sound like we're unpacking the zip into the given name.

katrinafyi (approved):
optional

PR #478, opened 2025-06-24
Organise imports

ailrst:
Add scalafix to remove redundant imports and organise imports automatically. Checked in CI like scalafmt. 

I would like to enable `-Werror` but it prevents `./mill fix` from actually fixing the warnings.


katrinafyi:
can we change the scalafmt so they're not fighting?

ailrst:
They're not fighting scalafmt wins

katrinafyi (approved):

PR #481, opened 2025-06-27
Loop invariant generation

b-paul:
Implements a loop invariant generation template. Abstract domains that are able to encode their results as predicates (so boogie assertions) can be used to get loop invariants by looking at the value of an analysis result at loop headers. Adding a new analysis to the list is simple (see `analysis/InvariantGenerator.scala`).

```scala
  def addInvariantsToProc(procedure: Procedure) = {
    reversePostOrder(procedure)

    val intervals = DoubleIntervalDomain(Some(procedure))
    ForwardLoopInvariantGenerator(intervals, NarrowingWorklistSolver(intervals)).genAndAddInvariants(procedure)

    val gammas = MayGammaDomain(
      LatticeMap.TopMap(procedure.formalInParam.unsorted.map(v => (v, LatticeSet.FiniteSet(Set(v)))).toMap)
    )
    ForwardLoopInvariantGenerator(gammas, worklistSolver(gammas)).genAndAddInvariants(procedure)

    // Add more domains here
  }
```

A test case for generating constant loop bounds needed correct widening and narrowing to give an optimal invariant, so the worklist solver was extended to support this. I am unsure if I did this in the best way so I'd appreciate it if someone could give their thoughts.

### DIFF
--- a/docs/src/SUMMARY.md
+++ b/docs/src/SUMMARY.md
@@ -14,6 +14,7 @@
 - [Basil IR](development/basil-ir.md) Explanation of BASIL's intermediate representation
 - [Interpreter](development/interpreter.md) Explanation of IR interpreter
 - [Transform Passes](development/transforms.md) Explanation of ir transforms and simplifications
+- [Program annotations](development/summaries-invariants.md) Explanation of loop invariant and function summary generators
 
 # Development
 
--- a/docs/src/development/summaries-invariants.md
+++ b/docs/src/development/summaries-invariants.md
@@ -1,0 +1,97 @@
+# Abstract interpretation with predicates
+
+## Predicates
+
+- see [src/main/scala/analysis/procedure_summaries/Predicate.scala](https://github.com/UQ-PAC/BASIL/blob/main/src/main/scala/analysis/procedure_summaries/Predicate.scala)
+
+Predicates encode semantic information about program states.
+They are distinct to IR boolean expressions as they can explicitly encode information about the gammas of variables.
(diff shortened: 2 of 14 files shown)


b-paul:
That makes a lot more sense than what I'm currently doing, I'll implement it now

ailrst (approved):
This seems useful, it infers a lower and upper bound on a trivial for loop as below so seems to be working as expected. 

```
memory shared $mem : (bv64 -> bv8);

prog entry @main_4195968;

proc @main_4195968 (R0_in:bv64, R1_in:bv64, R30_in:bv64, _PC_in:bv64, i_2_in:bv64, var2_4196000_bv64_1_in_in:bv64)
    -> ()
  { .name = "main"; .address = 0x400680 }
[
  block %main_entry {.address = 0x400680} [
    store le $mem var2_4196000_bv64_1_in_in:bv64 0x0:bv64 64;
    ( var i_2_3: bv64 := 0x0:bv64 );
    goto(%main_loop1_8);
  ];
  block %main_loop1_8 [
    // inferred:
    // assert booland(bvule(0x0:bv64, i_2_3:bv64), bvule(i_2_3:bv64, 0xb:bv64));
    goto(%phi_11, %phi_12);
  ];
  block %phi_11 [
    guard bvule(i_2_3:bv64, 0xa:bv64);
    ( var i_2_3: bv64 := bvadd(i_2_3:bv64, 0x1:bv64) );
    goto(%main_loop1_8);
  ];
  block %phi_12 [
    guard bvugt(i_2_3:bv64, 0xa:bv64);
    goto(%main_return);
  ];
  block %main_return [
    return ();
  ]
];

```

PR #493, opened 2025-07-07
download aslpOffline from Github packages

katrinafyi:
Github packages provides a maven-compatible interface, so we can use mill's ivy fetching infrastructure. this will also fix the mill build.

unfortunately, github requires a token in order to read packages - even if they're "public". i've hardcoded a read:packages token. this is probablyyyyyyy fine

### DIFF
--- a/build.mill
+++ b/build.mill
@@ -3,6 +3,8 @@ import $packages._
 
 import mill._
 import scalalib._
+import coursier.core.Authentication
+import coursier.maven.MavenRepository
 
 import $file.basilmill.mdbook.mdbookBinary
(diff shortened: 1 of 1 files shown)


ailrst (commented):
This seems like the kind of thing mill will decide to suddenly break on a point release but its probably less janky for now if it works

ailrst:
Auth + custom maven repo

katrinafyi:
That's true lol

ailrst (approved):
This is a bit silly but I don't know what else we can do, it possibly deserves a comment though on why you are obscuring the token

PR #508, opened 2025-07-15
avoid recursion in text wrap for graph labels

ailrst:


ailrst:
Kait has made the point that this will be quadratic due to string concatenation, my counterpoint is that the jvm is supposed to be magic. Anyway it should probably be replaced with a stringbuilder before merging

PR #512, opened 2025-07-16
 gtsrelf: include `LOCAL` globals, apply `atEnd`, and add flag to throw on mismatch

katrinafyi:
works towards https://github.com/UQ-PAC/BASIL/issues/509

### DIFF
--- a/src/main/scala/Main.scala
+++ b/src/main/scala/Main.scala
@@ -1,6 +1,7 @@
 // package scala
 
 import mainargs.{Flag, ParserForClass, arg, main}
+import gtirb.GTIRBReadELF
 import util.boogie_interaction.BoogieResultKind
 import util.{
   AnalysisResultDotLogger,
@@ -399,12 +400,14 @@ object Main {
         val realRelfFile = loadingInputs.relfFile
 
         Logger.setLevel(LogLevel.DEBUG)
-        val (relf, gtirb) = (realRelfFile, gtirbRelfFile) match {
-          case (Some(relfFile), _) =>
-            val (a, b) = IRLoading.loadReadELFWithGTIRB(relfFile, loadingInputs)
-            (Some(a), b)
-          case (None, Some(_)) => (None, Some(IRLoading.loadGTIRBReadELF(loadingInputs)))
-          case _ => throw IllegalArgumentException("--dump-relf requires either --relf or a GTIRB input")
+        val (relf, gtirb) = GTIRBReadELF.enableRelfCompatibilityAssertion.withValue(false) {
+          (realRelfFile, gtirbRelfFile) match {
+            case (Some(relfFile), _) =>
+              val (a, b) = IRLoading.loadReadELFWithGTIRB(relfFile, loadingInputs)
+              (Some(a), b)
+            case (None, Some(_)) => (None, Some(IRLoading.loadGTIRBReadELF(loadingInputs)))
+            case _ => throw IllegalArgumentException("--dump-relf requires either --relf or a GTIRB input")
+          }
         }
 
         // skip writing files if the given path is an empty string. this checks compatibility and exits.
@@ -474,7 +477,9 @@ object Main {
 
     Logger.info(programNameVersionHeader)
 
-    val result = RunUtils.run(q)
+    val result = gtirb.GTIRBReadELF.enableRelfCompatibilityAssertion.withValue(false) {
+      RunUtils.run(q)
+    }
     if (conf.verify.value) {
       assert(result.boogie.nonEmpty)
       var failed = false
--- a/src/main/scala/Main.scala
+++ b/src/main/scala/Main.scala
@@ -1,7 +1,7 @@
 // package scala
 
-import mainargs.{Flag, ParserForClass, arg, main}
 import gtirb.GTIRBReadELF
+import mainargs.{Flag, ParserForClass, arg, main}
 import util.boogie_interaction.BoogieResultKind
 import util.{
   AnalysisResultDotLogger,
--- a/src/main/scala/Main.scala
+++ b/src/main/scala/Main.scala
@@ -400,7 +400,7 @@ object Main {
         val realRelfFile = loadingInputs.relfFile
 
         Logger.setLevel(LogLevel.DEBUG)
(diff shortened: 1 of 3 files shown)


ailrst:
> Do we want to precisely mimic the (probably wrong) oldrelf? Or, should we strike out and I'll modify the compat check?

We probably want them somewhere, but should implement it correctly rather than just matching the previous implementation.

If they are just mapped as variables into the this proceses' address space then it might make sense to treat them like regular globals.

katrinafyi:
Yeah, so the oldrelf line
```
    ExternalFunction("optind", BigInt("4391784")),
```
also has a global variables line
```
    SpecGlobal("optind", 32, None, BigInt("4391784")),
```
and the SpecGlobal is also present in the gtsrelf. So maybe I'll remove them from ExternalFunctions.

The only problem might be if two libraries are linked together, I think we lose the information that the underlying variables are the same. But this shouldn't be a problem as long as we look at whole ELFs.

l-kent:
> But @ l-kent says re-lifting didn't help. Maybe worth looking into it.

This is likely compiler/OS etc. version differences, or potentially a ddisasm version difference given that it seems to be related to a warning about relocating the BSS section that I get from ddisasm

> Do we want to precisely mimic the (probably wrong) oldrelf?

Including those external variables as ExternalFunctions is an instance where the existing ReadELFLoader is imprecise - external functions can appear in the .rela.dyn section but aren't readily distinguishable from external variables, so they're currently all included as ExternalFunctions. I think the way to improve this is by matching the names to the .dynsym table entries but having those extraneous ExternalFunctions hadn't previously caused any problems so no one had looked any further at it.

l-kent:
The optind external variable in cntlm-noduk seems to just be a rarer case that we haven't given any particular thought to before - it's type R_AARCH64_COPY in .rela.dyn and appears in .dynsym and .symtab too, all with the same address, so the .symtab entry causes it to be included as a SpecGlobal and the .rela.dyn entry causes it to be (incorrectly) included as an ExternalFunction.

katrinafyi:
To clarify, when you say "none of it is solved by re-lifting" in the other thread, are you using main branch or this https://github.com/UQ-PAC/BASIL/pull/512 branch to do the gtsrelf/oldrelf comparison? In https://github.com/UQ-PAC/BASIL/issues/509#issuecomment-3082217972, I was using the https://github.com/UQ-PAC/BASIL/pull/512 branch. Can you confirm which you used? Sorry for the confusion.

l-kent:
That was with the main branch. On this branch, the only discrepancies for the malloc_memcpy_strlen_memset_free tests involve __bss_start, which re-lifting doesn't change.

```
   + [WARN]   PLEASE REPORT THIS ISSUE! (https://github.com/UQ-PAC/BASIL/issues/509) include the gts and relf files.
          gtirb relf discrepancy, symbol tables differ:
          gtirb - relf = HashSet(ELFSymbol(70,69664,0,NOTYPE,GLOBAL,DEFAULT,Section(23),__bss_start__), ELFSymbol(92,69664,0,NOTYPE,GLOBAL,DEFAULT,Section(23),__bss_start))
          relf - gtirb = HashSet(ELFSymbol(92,69660,0,NOTYPE,GLOBAL,DEFAULT,Section(23),__bss_start), ELFSymbol(70,69660,0,NOTYPE,GLOBAL,DEFAULT,Section(23),__bss_start__)) [checkReadELFCompatibility@GTIRBReadELF.scala:290]
   + [WARN]   PLEASE REPORT THIS ISSUE! (https://github.com/UQ-PAC/BASIL/issues/509) include the gts and relf files.
          gtirb relf discrepancy, symbol tables differ:
          gtirb - relf = HashSet(ELFSymbol(70,69664,0,NOTYPE,GLOBAL,DEFAULT,Section(23),__bss_start__), ELFSymbol(92,69664,0,NOTYPE,GLOBAL,DEFAULT,Section(23),__bss_start))                                                                                                        
          relf - gtirb = HashSet(ELFSymbol(92,69660,0,NOTYPE,GLOBAL,DEFAULT,Section(23),__bss_start), ELFSymbol(70,69660,0,NOTYPE,GLOBAL,DEFAULT,Section(23),__bss_start__)) [checkReadELFCompatibility@GTIRBReadELF.scala:290
```]

katrinafyi:
I also see the ddisasm warning with your binary. But honestly, idk where to go from this. I'm using the Docker-based infrastructure in #288 and that produces a binary that lifts with no warnings and seems to work. I'm happy enough with that, and I think it would be really hard to dig into why your system's compiler has this problem.

Anyway, do you want to review this PR?

katrinafyi:
I don't really want to special case certain symbol entries by name. I've lessened the error message so it doesn't say "PLEASE REPORT THIS ISSUE" for these particular test cases. Eventually, we should move to a repeatable lifting environment which doesn't have the mismatch

l-kent:
I think it makes a lot of sense to have a special case where we ignore a known inconsistency in behaviour between two different tools (readelf and DDisasm) that doesn't seem to actually matter. The point of these checks is to just to make sure that the behaviour of BASIL is consistent in reading the symbol table etc. data from readelf vs. DDisasm + GTIRB, isn't it, to ensure that the new DDisasm + GTIRB symbol table pipeline is working correctly? In this case, we've determined that the inconsistency is due to DDisasm, not due to an error in BASIL.

It's possible to be more precise about what DDisasm is doing - it is changing the address of the `__bss_start` and `__bss_start__` symbols to match the start of section 24, the `.bss` section. If this is the only difference between the symbols we get from readelf and DDisasm/GTIRB, then it seems fine to ignore it. If there's other inconsistencies with the addresses of those symbols, then that would be a new problem we'd have to look into.

l-kent:
> In the alternative, we would not notice if more binaries started having the bss mismatch.

Why do we need to notice this? This isn't an issue with anything other than DDisasm and readelf handling certain binaries (where the .bss section doesn't exactly match the __bss_start symbol) differently.

ailrst (approved):
@l-kent assuming you aren't interested in reviewing

This change set looks fine. The only difference in cntlm is now the `@GLIBC` tag dropped and the extern globals only appearing in the globals section which is what we want.

PR #522, opened 2025-07-25
Interval-tnum reduced product

alegs-a:
A domain computing the reduced product of the interval and tnum analyses. For a whole lot of words, go read [winter_research_reduced_product.pdf](https://github.com/user-attachments/files/21423271/winter_research_reduced_product.pdf)

### DIFF
--- a/src/main/scala/analysis/Domains.scala
+++ b/src/main/scala/analysis/Domains.scala
@@ -29,98 +29,121 @@ trait ProductDomain[L1, L2] extends AbstractDomain[(L1, L2)] {
   def bot: (L1, L2) = (d1.bot, d2.bot)
 }
 
-/** A domain that is the reduced product of two other domains.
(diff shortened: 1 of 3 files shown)


alegs-a:
This can't be merged yet because the known bits analysis doesn't play nicely. Specifically, getting variables out of known bits' map (`unreduced._2(v)`) on line 88 of Domains.scala will cause a crash if that variable doesn't exist. This is the reason for the test which currently fails. If the known bits analysis used a LatticeMap then we could .getOrElse() to go to bottom in this case, or there are probably other ways of gracefully handling this that I'm not aware of.

agle:
My plan with this is to refactor this on top of a single-value domain abstraction to avoid the complexity of computing the reduced product of two different map types. This means waiting for a slightly substantial refactor of the known bits analysis.

PR #532, opened 2025-08-04
tests: fix exhaustivity warning

katrinafyi:
previously, the compiler flagged this exhaustivity warning
```
[warn] 108 |    val failureMsg = boogieResult.kind match {
[warn]     |                     ^^^^^^^^^^^^^^^^^
[warn]     |match may not be exhaustive.
[warn]     |
[warn]     |It would fail on pattern case: util.boogie_interaction.BoogieResultKind.Verified(_, _), AssertionFailed
```
because the match statement used predicate conditions. the scala compiler could not reason that the predicates covered all cases by way of law of excluded middle.

this change moves the conditionals into the rhs of the match, so exhaustivity is satisfied.


katrinafyi:
i'm gonna go ahead and say that this Closes https://github.com/UQ-PAC/BASIL/issues/339 because the deprecation warnings have been addressed and this is the last remaining code warning.

Issue #547, opened 2025-08-22
Missing precondition when using `--noif` flag


aaronbembenek:
Hi there,

I've run the following program through Basil (4c3b0cc) with the `--parameter-form --noif` flags:

```c
void __attribute__((noinline)) my_assert(int x) {
  while (!x)
    ;
}

int main() {
	my_assert(0);
}
```

with this spec:

```
Subroutine: my_assert
Requires: R0 != 0bv64
```

and gotten this Boogie output (fragment):

```boogie
procedure p$my_assert_4196164(#R0_in: bv64, #R31_in: bv64, #_PC_in: bv64) returns (#R0_out: bv64, #R31_out: bv64, #_PC_out: bv64);
  modifies $stack;
  free requires (memory_load32_le($mem, 4196252bv64) == 131073bv32);
```

Note that the precondition `R0 != 0bv64` for `my_assert` from the spec is missing from the Boogie procedure. It's included when I omit the `--noif` flag.

Thanks!

[missing-spec.zip](https://github.com/user-attachments/files/21943821/missing-spec.zip)

PR #550, opened 2025-08-25
add spec to --noif output

agle:
Fixes https://github.com/UQ-PAC/BASIL/issues/547 but makes no attempt to do something meaningful if the spec contains undefined gammas.


katrinafyi (approved):
maybe unnecessaru default arguments and some things could be factored, but nbd

Issue #548, opened 2025-08-22
Variable name mismatch in generated Boogie code


aaronbembenek:
Hi there,

I ran this program through Basil (4c3b0cc) using the `parameter-form` flag:

```c
void __attribute__((noinline)) my_assert(int x) {
  while (!x)
    ;
}

int main() {
	my_assert(0);
}
```

with this spec:

```
Subroutine: my_assert
Requires: R0_in != 0bv64
```

and got this Boogie output (fragment):

```boogie
procedure p$my_assert_4196164(#R0_in: bv64, #Gamma_R0_in: bool, #R31_in: bv64, #Gamma_R31_in: bool, #_PC_in: bv64, #Gamma__PC_in: bool) returns (#R0_out: bv64, #Gamma_R0_out: bool, #R31_out: bv64, #Gamma_R31_out: bool, #_PC_out: bv64, #Gamma__PC_out: bool);
  modifies $Gamma_stack, $stack;
  requires ($R0_in != 0bv64);
```

Note that there are two variables representing the initial `R0` register, `#R0_in` (a parameter) and `$R0_in` (in the precondition). This doesn't happen if the spec refers to `R0` instead of `R0_in`.

Thanks!

[var-mismatch.zip](https://github.com/user-attachments/files/21943923/var-mismatch.zip)

PR #551, opened 2025-08-25
fix spec referring to explicit proc params

agle:
Fixes https://github.com/UQ-PAC/BASIL/issues/548

Note that only L613-L618 contributes to the fix and the expr spec case isn't tested.


src/main/scala/ir/transforms/ProcedureParameters.scala
```diff
@@ -580,8 +580,10 @@ object SpecFixer {
       case _ => DoChildren()
     }
 
-    def changeVar(v: Variable) = {
+    val defined = (varInPre.values ++ varInPost.values).map(v => v.name -> v).toMap
+    def changeVar(v: Variable): VisitAction[Variable] = {
       val repl = v match {
+        case l if defined.contains(l.name) => l
```
katrinafyi, line 586: `defined` can be a set, right?

src/main/scala/ir/transforms/ProcedureParameters.scala
```diff
@@ -608,9 +610,12 @@ object SpecFixer {
     makeLocal: Boolean = false
   )(b: BExpr): BExpr = {
     val varToOld = convVarToOld(varInPre, varInPost, isPost, makeLocal)
+    val defined = (varInPre.values ++ varInPost.values).toSet
     val preVars = varInPre.values.toSet
     val postVars = varInPost.values.toSet
```
katrinafyi, line 615: could use `val defined = preVars ++ postVars`

katrinafyi (approved):

Issue #464, opened 2025-06-17
Structure for value analysis


ailrst:
Ideally our IR would be structured as a free algebra with recursion schemes, lacking this...

We would like an interface defined for symbolic evaluation of expressions. This is implemented once to a reasonably complete extent in the [Known Bits Analysis](https://github.com/UQ-PAC/BASIL/blob/main/src/main/scala/analysis/KnownBits.scala). It would be good for this structure to be reusable for future planned value analyses to reduce some of the burden of deciding how to do this each time an analysis is written. Possible analyses we will implement in future are live bits, interval, SASI etc.

### Acceptance critera

1. implement an interface for defining a single-value 1-dimensional abstract domains (domains whose concretisation is a set of values) that abstracts out the recursion over expressions. 
   - It should be possible to instantiate a state domain (e.g. maplattice) with an arbitrary a value domain (matching the interface defined in branch `value-analysis-iface`).
2. validate this by refactoring the known bits analysis to use this abstraction 
   - use the implement the value abstraction interface rather than defining its own recursive breakdown of the expressions
   - remove the custom state domain map from known bits and use the generic map lattice state domain

ailrst:
Partial work in branch `value-analysis-iface`

PR #553, opened 2025-08-25
add ValueLattice abstraction and refactor TNum for known bits

katrinafyi:
there is a big problem here with the Top value. the new ValueAnalysis infrastructure expects a unique value `top: Value` but the `top` in TNum requires a bit width parameter. the TNum lattice is actually not really valid because values with differing widths are incompatible - meet/join will fail when given different widths.

maybe we need an abstraction of bit-width which would allow the use of _width-dependent_ lattices and coerce to Top when widths differ.


katrinafyi:
i've added a TypedLattice to abstract the bitwidth-dependence of the lattices
https://github.com/UQ-PAC/BASIL/blob/34bed3353d11ef229c5dfc90754f1626b15f7970/src/main/scala/analysis/ValueAnalysis.scala#L102-L120

however, there are still some problems:
- it would be nice to allow the "inner" lattice operations to return the universal top/bot. however, this is kind of tricky because it would be mutually recursive or it would make it harder to define the cases which are width-specific or both.
- F-bounded polymorphism is really annoying because it requires subtyping for customisation. however, subclassing the TypedLattice enum causes problems especially with the enum classes (which i believe wont pick up the subtyped parent class).

katrinafyi:
related to https://github.com/UQ-PAC/BASIL/issues/464

Issue #271, opened 2024-11-13
GTIRB frontend produces GoTos between procedures


ailrst:
In the attached example, when lifted with ddisasm and gtirb_semantics it produces a goto in close_file that targets a block in __stdio_seek. 

[gate_server.tar.gz](https://github.com/user-attachments/files/17724906/gate_server.tar.gz)

```
[ERROR]   4246544$2: GoTo($__stdio_seek_4245436$__0__$_ka7cqh7QzOFuSiPbkNnsQ) has target outside parent procedure close_file_4246448 [blocksUniqueToEachProcedure@BlocksUniqueToProcedure.scala:64]
```

This property is checked with the following code: https://gist.github.com/ailrst/847a9aecc909a47b05c8634c5aa8070a. This is checked in to the simplification-pass branch/pr.

l-kent:
Without having the input to examine it's hard to say what the issue is, but it could very well be a bug in ddisasm's function bounds identification.

Can you share the .json output just for the close_file subroutine? That should be enough to figure out what's going on.

ailrst:
Sorry, I forgot to mention it will require the fixes to gtirb_semantics that aren't merged yet.

This commit should work as is, the PR branch will require a small change to the basil frontend to handle the error json format for unsupported instructions.  https://github.com/UQ-PAC/gtirb-semantics/commit/9ff14e9efdb29e8650807c48b97b2ef5c83b047d

I'm on leave now so don't have access to my computer with the lifted files. It might be easier to wait until I'm back and can merge and update the nix package.

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

l-kent:
If you're on leave then you don't need to respond to this until you're back, it's fine.

Looking at the assembly, there's an indirect call in close_file that uses `br` not `blr` which would be why ddisasm treats it as a within-function jump. The issue is that it's actually a non-returning tail call which isn't something we currently account for, but it should be easy to recognise that case and handle it better. The control flow won't really be working properly until we have non-returning calls implemented properly though.

The indirect call appears to be to a function pointer contained within a struct that's passed by reference to close_file, so it's not easy to tell where it can point - if ddisasm has managed resolved it to a single location then that's somewhat surprising.

ailrst:
The json for that procedure is here: https://gist.github.com/ailrst/0178f78739fe18aed99e445f1b16b2a4#file-close_file-txt-L235

[...]

PR #565, opened 2025-09-15
Improved support for non-returning indirect branches

l-kent:
Handles #271

In cases where there are non-returning indirect branches to other procedures, the frontend will now treat these as calls to those procedures (previously they were not handled or treated as GoTos which did not work), followed by Unreachable as R30 is not set to point to the following block.

Further work still needs to be done to handle these calls more accurately. This approach should be sound, but it should be possible to improve things with an analysis that identifies what R30 points to when these procedures have a Return statement, and adjust the control flow appropriately. The simplest case, which is the only one we've encountered so far, is tail calls, where R30 is set to be the input R30 to the calling procedure, so when the callee returns it just jumps to the callsite of the calling procedure (skipping a level, basically). I think constructing a heuristic to deal specifically with this case shouldn't be too difficult.

I haven't tested this much due to not having any suitable examples on hand - it seems to work but I haven't dug into whether the output is exactly what it should be because none of the examples in #271 are really suitable for more extensive investigation unfortunately (in one case the relevant procedure isn't actually called anywhere and is optimised out, in the other the resulting .bpl is multiple GB)


l-kent:
I don't know why SMTTest is timing out, it works fine for me locally and I'm not sure how this change would have any effect on it

agle (approved):
This looks reasonable for the oggenc example;

```
  block %oggpack_writecopy_helper_loop1_62 {.address = 0x406e50; .originalLabel = "AuDgVY4rT7y7zwaSb90+Zw=="} [
    assert eq(_PC_32:bv64, 0x406e50:bv64) { .comment = "pc-tracking"; .label = "pc-tracking" };
    var Exp16__5_113_1: bv8 := load le $mem bvadd(R1_in:bv64, R19_5:bv64) 8 { .label = "4222544_0$0" };
    var R1_7: bv64 := zero_extend(56, Exp16__5_113_1:bv8) { .label = "4222544_1" };
    var R2_9: bv64 := 0x8:bv64 { .label = "4222552_0" };
    var R30_2: bv64 := 0x406e64:bv64 { .label = "4222560_0" };
    goto(%phi_14, %phi_15);
  ];
  block %phi_14 {.originalLabel = "AuDgVY4rT7y7zwaSb90+Zw=="} [
    assume eq(R3_in:bv64, 0x405124:bv64);
    var (R0_14:bv64, R10_16:bv64, R11_16:bv64, R12_16:bv64, R13_16:bv64, R14_16:bv64, R15_16:bv64, R16_16:bv64, R17_16:bv64, R18_16:bv64, R19_11:bv64, R1_11:bv64, R20_17:bv64, R21_17:bv64, R29_17:bv64, R2_13:bv64, R30_6:bv64, R31_17:bv64, R3_16:bv64, R4_16:bv64, R5_16:bv64, R6_16:bv64, R7_16:bv64, R8_16:bv64, R9_16:bv64, _PC_42:bv64)
        := call @oggpackB_write_4215076 (R21_12:bv64, R10_11:bv64, R11_11:bv64, R12_11:bv64, R13_11:bv64, R14_11:bv64, R15_11:bv64, R16_11:bv64, R17_11:bv64, R18_11:bv64, bvadd(R19_5:bv64, 0x1:bv64), R1_7:bv64, R20_12:bv64, R21_12:bv64, R22_10:bv64, R3_in:bv64, R24_2:bv64, R1_in:bv64, R29_12:bv64, R2_9:bv64, R30_2:bv64, R31_12:bv64, R3_11:bv64, R4_11:bv64, R5_11:bv64, R6_11:bv64, R7_11:bv64, R8_11:bv64, R9_11:bv64, R3_in:bv64, indirectCallTarget_in:bv64);
    goto(%phi_17); // jumps to %oggpack_writecopy_helper_56
  ];
  block %phi_15 {.originalLabel = "AuDgVY4rT7y7zwaSb90+Zw=="} [
    assume eq(R3_in:bv64, 0x404a20:bv64);
    var (R0_13:bv64, R10_14:bv64, R11_14:bv64, R12_14:bv64, R13_14:bv64, R14_14:bv64, R15_14:bv64, R16_14:bv64, R17_14:bv64, R18_14:bv64, R19_9:bv64, R1_10:bv64, R20_15:bv64, R21_15:bv64, R29_15:bv64, R2_12:bv64, R30_5:bv64, R31_15:bv64, R3_14:bv64, R4_14:bv64, R5_14:bv64, R6_14:bv64, R7_14:bv64, R8_14:bv64, R9_14:bv64, _PC_40:bv64)
        := call @oggpack_write_4213280 (R21_12:bv64, R10_11:bv64, R11_11:bv64, R12_11:bv64, R13_11:bv64, R14_11:bv64, R15_11:bv64, R16_11:bv64, R17_11:bv64, R18_11:bv64, bvadd(R19_5:bv64, 0x1:bv64), R1_7:bv64, R20_12:bv64, R21_12:bv64, R22_10:bv64, R3_in:bv64, R24_2:bv64, R1_in:bv64, R29_12:bv64, R2_9:bv64, R30_2:bv64, R31_12:bv64, R3_11:bv64, R4_11:bv64, R5_11:bv64, R6_11:bv64, R7_11:bv64, R8_11:bv64, R9_11:bv64, R3_in:bv64, indirectCallTarget_in:bv64);
    goto(%phi_16); // jumps to %oggpack_writecopy_helper_56
  ];
  block %oggpack_writecopy_helper_56 {.address = 0x406e64; .originalLabel = "lIyI+XNSQeeOJXRY9J88fQ=="} [
    assert eq(_PC_44:bv64, 0x406e64:bv64) { .comment = "pc-tracking"; .label = "pc-tracking" };
    goto(%phi_20, %phi_21);
  ];


```

@katrinafyi  as far as I can tell it won't require any changes for `--pc assert`, here the assertion is the precondition of the called procedures or jump targets.

agle:
I think that was because boogie alone can't prove the assertions at least some of the time

l-kent:
Seems like it's just one example that's causing most of the problems, I'll have a look at it.

The LiveVarsAnalysisTests need to account for it too but that should be easy to fix.

agle:
Probably in the `find()` method possibly due to a substitution cycle in the domain, its not checked for anything like that because the DSA shouldn't let it happen in theory I think, that assumption could easily be messed up by a weird traversal order regarding unreachable blocks. If we're really lucky it could just be a substitution chain longer than the stack :)

l-kent:
It's reproducible with `run -i examples\cntlm-noduk\cntlm-noduk.gts -r examples\cntlm-noduk\cntlm-noduk.relf --trim-early --simplify -o cntlm-noduk.bpl --dsa=local -m=direct_request` but getting anything useful out of it requires fiddling around with the debugger & more

agle:
So my guess is that when variables are read-uninitialised in a loop, dsa assumes there is an initial value for the variable and creates phi nodes on the back edges etc. which creates this cycle copyprop sees. Copyprop expects to see an initial definition either in a parameter or assignment so that find always terminates on some initial definition. I should add an invariant check for read-uninitialised at some point.

As an aside I shouldn't expect the `R30_begin` proc name mangling to be neccessary, inlining should be correct with regards to local variables.

agle (approved):

PR #571, opened 2025-09-17
Add read-uninitialised locals check

agle:


agle:
In test `IntervalDSATest/cntlm globals - local phase` 4 more procedures collapse now:

```
      "hlist_subcmp_all",
      "headers_recv",
      "scanner_hook",
      "config_open"
```

Not sure which change causes this, it doesnt look like the inlining fix  or the change to liveness alone cause it.

agle:
Seems fixed now, idk why I couldn't get the tail-recursive `find` op to work, probably a skill issue.

l-kent:
Is the read-uninitialised invariant check supposed to catch the sort of case we had with cntlm_noduk?

If I revert the fix for the inter-procedural live variables analysis and run the cntlm-no-duk  InvervalDSATest, the infinite loop in the copy prop is identified and an exception is thrown, but the read-uninitialised invariant doesn't seem to catch any of that?

agle:
It detects it when placing the invariant check in the simplify pipeline before the copyprop pass, immediately after DSA, I'll commit a check there too.

src/test/scala/IntervalDSATest.scala
```diff
@@ -626,6 +626,8 @@ class IntervalDSATest extends AnyFunSuite with test_util.CaptureOutput {
     val locals = res.dsa.get.local
     assert(locals.values.forall(_.glIntervals.size == 1))
 
+    println(locals.values.filterNot(IntervalDSA.checksStackMaintained).map(_.proc.procName).toSet)
```
l-kent, line 629: Was this just for debugging purposes?

PR #575, opened 2025-09-23
add top to symvalues domain

agle:
Attempts to fix crashes seen in SVA when exprs evaluate to top in the abstract domain and calling `toOffsets`, e.g. in cntlm-O3 `https://github.com/agle/basilbench/tree/main/cntlm-noduk`

`./mill run -i ../tvbins/cntlm-noduk/cntlm-noduk.gts --lifter --dump-il cntlm --simplify --dsa= `

Just joins a `Constant -> top` to the symvalset domain if adding a top constant, theoretically with split-globals it should probably also map all the global bases to top since they could alias? Regardless the assertion should be enough to justify keeping them separate.


katrinafyi (approved):
looks reasonable. does using withDefaultValue cause any problems? an alternative would be using LatticeMap.TopMap.

agle:
I refactored it to be more sensible, I think i should be able to replace line 266 with `???`, as it was previously. It doesn't really make sense to have a top value of the the map of symbolic bases. But yeah map.withDefaultValue would cause problems, its horrendous, but topMap would probably require a really annnoying refactor.

PR #589, opened 2025-11-13
Jesse interface thesis proj

Jesseable:


agle:
Tests are failing due to the IR stashing between transform passes crashing on IRToDSL.resolve, probably a real bug in IRToDSL. This stashing should be disabled properly when the server isn't being run regardless.

frontend/src/utils/graphLayout.ts
```diff
@@ -0,0 +1,298 @@
...
+      `DEBUG: Parsed ${uniqueNodeCount} unique raw nodes for ELK. Total unique nodes: ${rawNodesData.length}`
+    );
+    console.log(
+      `DEBUG: Final rawNodesData after parsing (unique IDs):`,
+      rawNodesData.map((n) => n.id)
+    );
+
+    // --- Edge Parsing ---
+    const edgeRegex =
+      /(?:\"([^"]+)\"|(\w+))\s*->\s*(?:\"([^"]+)\"|(\w+))(?:\[label="((?:[^"\\]|\\.)*?)"(?:, ([^\]]*))?\])?/g;
```
agle: This seems like a very complex and fragile approach, it would have been more sensible to use a dot parsing library to load it into a structured format, or have the backend return the elk json format instead of dot, I think we'll have to refactor to do the latter because this code isn't maintainable.

src/main/scala/ir/transforms/SimplifyPipeline.scala
```diff
@@ -2,26 +2,43 @@ package ir.transforms
...
+): Unit = {
   // writeToFile(dotBlockGraph(program, program.filter(_.isInstanceOf[Block]).map(b => b -> b.toString).toMap), s"blockgraph-before-simp.dot")
   Logger.info("[!] Running Simplify")
   val timer = PerformanceTimer("Simplify")
   val program = ctx.program
 
+  val beforeIrreducibleLoops = IRToDSL.convertProgram(ctx.program).resolve
   IrreducibleLoops.transform_all_and_update(program)
+  val afterIrreducibleLoops = IRToDSL.convertProgram(ctx.program).resolve
+  addEpochSnapshot("irreducible_loops", beforeIrreducibleLoops, afterIrreducibleLoops, collectedEpochs)
```
agle: addEpochSnapshot is executed conditionally on whether the server is running, the IR cloning should also be conditional conditional on that so its not wasting work when the server isn't running.

The way the frontend is set up now, it would also make more sense to create a single snapshot between each pass, rather than two snapshots either side each pass; its creating 2x too many snapshots.

src/main/scala/server/IrServiceRoutes.scala
```diff
@@ -0,0 +1,479 @@
...
+   */
+  private val getBeforeCfgRoute: HttpRoutes[IO] = HttpRoutes.of[IO] { case GET -> Root / "cfg" / epochName / "before" =>
+    ensureReady {
+      logger.info(s"Received GET /cfg/$epochName/before request.") *>
+        epochStore
+          .getEpoch(epochName)
+          .flatMap {
+            case Some(epoch) =>
+              logger.info(s"Attempting to generate 'before' CFG dot graph for epoch: $epochName.") *>
+                IO.delay(generateDotGraphs(epoch.beforeTransform)).flatMap { dotGraph =>
```
agle: 90% of the code in getBforeCfgRoute and getAfterCfgRoute is identical and should be shared.

src/main/scala/server/IrServiceRoutes.scala
```diff
@@ -0,0 +1,479 @@
...
+      val cleanedDotOutput = removeFontAttributes(originalDotOutput)
+      proc.procName -> cleanedDotOutput
+    }.toMap
+  }
+
+  private def removeFontAttributes(dotString: String): String = {
+    val fontAttributePattern = Pattern.compile(
+      """\s*(?:fontname="[^"]+"|fontsize="\d+(?:\.\d+)?"|fontname=[a-zA-Z0-9_]+|fontsize=\d+(?:\.\d+)?)\s*""",
+      Pattern.CASE_INSENSITIVE
+    )
```
agle: Not sure what this is for? Assuming its removing stuff the frontend cant handle, in which case it should be the frontend's responsibility. It is also probably fragile to any changes in the node attribute list intended to change how dot renders the graph? Will it break if we change the fontcolour?

agle: It seems to be purely to make this line in the frontend's dot parser sufficient to prevent the graph attributes declaration from being interpreted as a node `  let cleanedDotString = dotString.replace(/graph\s*\[\];/g, '');`. It would be more robust to simply make that regex more general, `^graph.*;$` would be fine.

frontend/README.md
```diff
@@ -0,0 +1,69 @@
+# React + TypeScript + Vite
+
+This template provides a minimal setup to get React working in Vite with HMR and some ESLint rules.
```
agle: This file seems generic, does it provide information specific to the frontend you've implemented?

docs/src/development/api-information.md
```diff
@@ -0,0 +1,129 @@
...
+1. Install Frontend Dependencies: The web component uses a tool Vite that requires Node modules. 
+Install this before running the development server:
+```shell
+npm install
+```
+2. Run the _API_ Server (Backend). The following command uses the Mill build tool to automatically 
+download and compile all necessary Scala libraries, then starts the _API_ server.
+```shell
+./mill runMain API/ApiServer
+```
```
agle: Need to update this because i changed the package name, its now server/ApiServer

frontend/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "dev": "vite",
+    "build": "tsc -b && vite build",
+    "lint": "eslint .",
+    "preview": "vite preview",
+    "format": "prettier --write \"src/**/*.{js,jsx,ts,tsx,css,md}\"",
+    "check-format": "prettier --check \"src/**/*.{js,jsx,ts,tsx,css,md}\""
+  },
+  "dependencies": {
+    "@hpcc-js/wasm": "^2.24.3",
+    "@hpcc-js/wasm-graphviz": "^1.9.2",
```
agle: As far as I can tell wasm and wasm-graphviz are never actually used and these dependencies, and useGraphvizWASM can be deleted.

agle:
Given your thesis project is ended there is not an expectation you address any of these comments unless you really want to, thanks for your hard work this semester @Jesseable this is a really good result.

Jesseable:
Thanks for the comments @agle!
I was planning on doing some further work with this project and potentially open sourcing the frontend on my github. I'll try to look through the comments and respond to them over the weekend – although I'm slightly under the weather atm (classic post uni cold) so it might take a bit more time. I was also going to implement LOD functionality with these changes. I'm happy to push my changes to this MR if you'd like and ensure the tests that are passing atm continue passing. 
Thanks!

frontend/package.json
Jesseable: Good catch, removed

src/main/scala/server/ApiServer.scala
```diff
@@ -0,0 +1,130 @@
...
+        isReady,
+        semaphoreInstance,
+        generateIRAsync(epochStore, semaphoreInstance, isReady)
+      ).routes
+      httpApp = Router("/" -> irServiceRoutes).orNotFound
+
+      exitCode <- EmberServerBuilder
+        .default[IO]
+        .withHost(ipv4"0.0.0.0")
+        .withPort(port"8080")
```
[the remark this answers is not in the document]
agle: Make it a CLI flag

frontend/src/utils/graphLayout.ts
Jesseable: Definetly a better idea. I have changed the backend to send across a JSON result and modified the frontend acordingly.

src/main/scala/server/IrServiceRoutes.scala
```diff
@@ -0,0 +1,479 @@
...
+            _ <- logger.info(s"Using default hardcoded path: $inputFile")
+
+            fiber <- generateIRAsync(inputFile, relfFile)
+              .handleErrorWith(e => logger.error(e)("IR analysis failed"))
+              .start
+
+            _ <- fiber.join
+
+            resp <- Ok(s"Analysis successfully triggered for default path: $inputFile / ${relfFile.getOrElse("none")}")
+          } yield resp
```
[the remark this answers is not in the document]
Jesseable: I kept the default path as a 'starter example' so new users can see the UI working immediately without hunting for a file. It can be removed, but I feel it could be helpful for summer students or new BASIL workers.

src/main/scala/server/IrServiceRoutes.scala
Jesseable: As this is a backend change I will only tackle it as a bonus if I finish the frontend stuff early

Jesseable:
The last change was fairly substantial. I have removed the complete IR view and replaced it with a procedure-by-procedure view. I was not able to implement LOD functionality*.

Additionally, I’ve moved the frontend to a separate repository: [Decompiler-User-Interface-Project](https://github.com/Jesseable/Decompiler-User-Interface-Project) Any future changes will be made in that repo.

Thanks so much for all your help this year, it’s been a really cool project!

* Because the frontend controls the diff, there’s no way to determine how much information the backend must provide to display a certain number of visible lines, making LOD impractical :(

agle:
@katrinafyi how do I get the docs to build

PR #591, opened 2026-01-21
BasilIR.cf update from obasil: phi, var load/store, named params

katrinafyi:
compared to obasil, changes the phi block syntax to use square brackets to avoid ambiguity with simulassign statement.

see src/test/scala/ir/ParserTest.scala to learn the effect of phi blocks and named call params. these get lowered into sbasil simul assigns and unnamed params.


src/main/scala/ir/dsl/DSL.scala
```diff
@@ -270,8 +271,26 @@ case class EventuallyBlock(
...
+
+          val simulAssign = predBlock.statements.lastElem match {
+            case Some(assign: SimulAssign) if assign.label == Some("phi") => assign
+            case _ =>
+              val assign = SimulAssign(Vector(), Some("phi"))
+              predBlock.statements.append(assign)
+              assign
+          }
+
+          simulAssign.assignments :+= (phiVar -> predVar)
```
agle: LGTM

katrinafyi:
you want this? any suggestions for the indentation?
```
  block %inputs2
  (
    var R29:bv64 := phi(%inputs -> R29_in:bv64, %inputs2 -> R30_in:bv64),
  )
  [
    goto (%inputs);
  ];
  ```

agle:
The first option, indentation can be the same as it already is.

PR #622, opened 2026-04-10
add vacuous check to default boogie invocation

agle:


l-kent:
Does this need to update the version of Boogie for the Windows tests? Does this mean that the Windows and Linux Boogie versions are out of sync?

PR #627, opened 2026-05-12
Update printers to match BinCaml input

JTrenerry:
The output of BASIL does not match the input of BinCaml in a couple places, this updates that to output what it expects.

Uninterp function should be a more function programming looking function i.e.

`name : (a1 -> a2) -> o1`

## TODO
Variables can no longer have `.` in the name, the current idea is just to map them to be _ instead, but I don't like that idea and I would love something better.

### DIFF
--- a/src/main/scala/translating/PrettyPrinter.scala
+++ b/src/main/scala/translating/PrettyPrinter.scala
@@ -259,7 +259,7 @@ class BasilIRPrettyPrinter(
 
     val uninterp = getUninterp(p)
     val ufdecls = uninterp.map { case (n, pt, rt) =>
-      s"declare-fun ${Sigil.BASIR.globalVar}$n : (${pt.map(vtype).mkString(", ")}) -> ${vtype(rt)}"
+      s"val ${Sigil.BASIR.globalVar}$n : (${pt.map(vtype).mkString("-> ")}) -> ${vtype(rt)}"
     }
 
     Prog(
--- a/src/main/scala/translating/PrettyPrinter.scala
+++ b/src/main/scala/translating/PrettyPrinter.scala
@@ -259,7 +259,7 @@ class BasilIRPrettyPrinter(
 
     val uninterp = getUninterp(p)
     val ufdecls = uninterp.map { case (n, pt, rt) =>
-      s"val ${Sigil.BASIR.globalVar}$n : (${pt.map(vtype).mkString("-> ")}) -> ${vtype(rt)}"
+      s"val ${Sigil.BASIR.globalVar}$n : ${pt.map(vtype).mkString("-> ")} -> ${vtype(rt)}"
     }
 
     Prog(
--- a/src/main/bnfc/BasilIR.cf
+++ b/src/main/bnfc/BasilIR.cf
@@ -46,9 +46,9 @@ Decl_SharedMem . Decl ::= "memory" "shared" GlobalIdent ":" Type ;
 Decl_UnsharedMem . Decl ::= "memory" GlobalIdent ":" Type ;
 Decl_Var . Decl ::= "var" GlobalIdent ":" Type  ;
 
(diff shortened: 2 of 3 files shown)


katrinafyi:
is 
```
name : (a1 -> a2) -> o1
```
meant to be a function taking two arguments and returning o1? is this what basil outputs?

it seems like an incorrect application of currying..

JTrenerry:
`val $FPAdd_64 : (bv64 -> bv64 -> bv32) -> bv64;`

val name : type

Honestly, looking over the grammar, I am unsure why this parses, but it does and gives

map (map (bv64, map(bv64, bv64)) , bv64)

that doesnt make sense though, but removing the brackets does I think? (which I think is what you were saying)

`name : a1 -> a2 -> o1`

which gets parsed like

(a1 -> (a2 -> o1))

src/main/scala/translating/PrettyPrinter.scala
```diff
@@ -259,7 +259,7 @@ class BasilIRPrettyPrinter(
 
     val uninterp = getUninterp(p)
     val ufdecls = uninterp.map { case (n, pt, rt) =>
-      s"declare-fun ${Sigil.BASIR.globalVar}$n : (${pt.map(vtype).mkString(", ")}) -> ${vtype(rt)}"
+      s"val ${Sigil.BASIR.globalVar}$n : (${pt.map(vtype).mkString("-> ")}) -> ${vtype(rt)}"
```
katrinafyi: yeah remove the brackets
```suggestion
      s"val ${Sigil.BASIR.globalVar}$n : ${pt.map(vtype).mkString("-> ")} -> ${vtype(rt)}"
```

katrinafyi (changes requested):
can you also add a minimal test case?

JTrenerry:
> can you also add a minimal test case?

edited the failing test case to match the new syntax, it looks like the is just loading it, did you want a test that tests printing it too?

katrinafyi (approved):
@JTrenerry i've pushed some commits to fix the parser test. let me know if you approve.

the sbasil grammar is quite outdated atm and this is really just a stop-gap until we bring the grammar more up to date with obasil, e.g. with https://github.com/UQ-PAC/BASIL/pull/591.

also, sbasil doesn't even store the uninterpreted functions into the IR....

JTrenerry:
works, thank you

src/main/scala/translating/PrettyPrinter.scala
JTrenerry, line 585:
```suggestion
override def vrvar(e: Variable): PPProg[Variable] = 
val namefixed = e.name.replace(".", "7")
e match {
case l: LocalVar =>
val sigil = if localSigils then Sigil.BASIR.localVar else ""
BST(s"${sigil}${e.name}:${vtype(e.getType)}")
case l: Global => BST(s"${Sigil.BASIR.globalVar}${e.name}:${vtype(e.getType)}")
}
override def vlvar(e: Variable): PPProg[Variable] = {
val namefixed = e.name.replace(".", "7")
e match {
case l: LocalVar =>
val sigil = if localSigils then Sigil.BASIR.localVar else ""
BST(s"var $sigil${e.name}: ${vtype(e.getType)}")
case l: Global => BST(s"${Sigil.BASIR.globalVar}${e.name}: ${vtype(e.getType)}")
}
}
```

@katrinafyi I also want to replace the "." in all variable names with anything else (record access needs . in bincaml) is there a better way to do this (probably shouldn't use "7" but just as an example lol)

katrinafyi: whoops yeah maybe I jumped the gun. I'll take a look

JTrenerry: Nah, I am not sure if there is a fix for this that is worthy of being on main haha

PR #632, opened 2026-06-12
CalleePreservedParam bug fix

l-kent:
Fixes bug in CalleePreservedParam found by @utting. 

The `preserved` set was never accessed as the match cases that lead to it were redundant, so assertions were never generated for R19 to R29. This is now included as intended.


agle:
@katrinafyi lychee false positive

